PR #1, opened 2021-04-14
Replace header.component.html href attributes with AngularRouter routerLink

mfrey43:


dtslvr (approved):
Thanks for your contribution! 😎

PR #8, opened 2021-04-17
Add license

dtslvr:


vzickner:
Maybe the license should be added to the [package.json](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license) as well?

dtslvr:
> Maybe the license should be added to the [package.json](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license) as well?

Yes, very good point. Thanks 😃

PR #71, opened 2021-05-05
Add account filter with chips on transactions

mantovanig:
I added the possibility to filter the transactions by account using chips component.

This is a first basic solution but in the future it can be improved by combining it with other fields such as type to build advanced search queries.

The chips component is displayed only when there are more than 2 different accounts in the transactions.

The search with text input is applied to the list filtered by account.

Preview:
Screenshot 2021-05-05 at 17 59 54

### DIFF
--- a/apps/client/src/app/components/transactions-table/transactions-table.component.html
+++ b/apps/client/src/app/components/transactions-table/transactions-table.component.html
@@ -1,3 +1,24 @@
+<div *ngIf="chipsAccount.length > 1">
+  <mat-chip-list aria-label="Account selection" class="d-block mb-4">
+    <mat-chip
+      [selected]="true"
+      #c="matChip"
+      (click)="!c.selected && c.toggleSelected()"
+      (selectionChange)="applyAccountFilter($event)"
+    >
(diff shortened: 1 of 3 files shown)


dtslvr:
Hmm, I'm not completely convinced 🙂 If the filter is not combined with the search field, it uses a lot of space (especially on mobile). I think it would already help if you could enter a part of the account name and it filters correctly. I have created an issue for that #72. What do you think @mantovanig?

mantovanig:
I agree with you for the problem of space. But if we just fix the filter on text input you can't do advanced or composite search queries.
For example if you want to search all transactions of a specific symbol in just one account.

Probably another solution can be a small select before the search text input on the same row.
Example: https://material.angular.io/components/form-field/overview#form-field-theming

But if you're still not convinced for now let's just do the issue you opened.

vzickner:
Hi @mantovanig, I created a PR with a filter box that supports filtering for accounts with a plain text filter: https://github.com/ghostfolio/ghostfolio/pull/76 What do you think about this approach?

PR #73, opened 2021-05-06
Add filterPredicate on transactions table to filter by account name

### LINKED ISSUE #72: Filtering by account name is not working in transactions table
### Steps to reproduce

1. Go to https://www.ghostfol.io
2. Click _Live Demo_
3. Navigate to _Transactions_
4. Enter "coin" into the search field

### Expected behavior
The `BTCUSD` transaction should show up

mantovanig:
I added the filterPredicate to filter transactions by Account name.

The default filter predicate of material table doesn't work with nested properties.

image

Closes #72

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed the filtering by account name in the transactions table
+
 ## 1.0.0 - 05.05.2021
 
 ### Added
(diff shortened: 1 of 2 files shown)


dtslvr:
Great work, @mantovanig.

I just had to change the type from `OrderModel` to `OrderWithAccount`. Please see https://github.com/ghostfolio/ghostfolio/pull/73/commits/c081e5e5a3903f75131ee62006608073b9c60f4d.

mantovanig:
Good! @dtslvr Only one question about your minor fix.

Are you sure that it's a good practice import types from another app? Do you think it's better to declare this types inside client app?

dtslvr:
> Good! @dtslvr Only one question about your minor fix.
> 
> Are you sure that it's a good practice import types from another app? Do you think it's better to declare this types inside client app?

@mantovanig I've seen that _eslint_ is not happy with it: `Imports of apps are forbiddeneslint@nrwl/nx/enforce-module-boundaries`. Do you have an idea how to solve this? What about a library `@ghostfolio/libs/common` to share types?

PR #74, opened 2021-05-07
Add link to retrieve manually the current market price

mantovanig:
Affected component: `create-or-update-transaction-dialog.component`

I added a link (suggestion) under the unit price input to retrieve and use the current market price retrieved from the data source.

In this way the user has more control and everything is more transparent.
The link is displayed only when a valid symbol is present and a current market price could be retrieved.

In addition I added the `fetchSymbolItem` on init, so if the user come from "Clone" action can immediately see and click the link to use the current price.

image

### DIFF
--- a/apps/client/src/app/pages/transactions/create-or-update-transaction-dialog/create-or-update-transaction-dialog.component.ts
+++ b/apps/client/src/app/pages/transactions/create-or-update-transaction-dialog/create-or-update-transaction-dialog.component.ts
@@ -30,9 +30,9 @@ import { CreateOrUpdateTransactionDialogParams } from './interfaces/interfaces';
 })
 export class CreateOrUpdateTransactionDialog {
   public currencies: Currency[] = [];
+  public currentMarketPrice = null;
   public filteredLookupItems: Observable<LookupItem[]>;
   public isLoading = false;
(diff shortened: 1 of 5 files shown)


apps/client/src/app/pages/transactions/create-or-update-transaction-dialog/create-or-update-transaction-dialog.component.ts
```diff
@@ -65,6 +66,21 @@ export class CreateOrUpdateTransactionDialog {
...
+        .subscribe(({ marketPrice }) => {
+          this.currentMarketPrice = marketPrice;
+          this.cd.markForCheck();
+        });
+    }
+  }
+
+  public applyCurrentMarketPrice() {
+    this.data.transaction.unitPrice = this.currentMarketPrice;
+    this.cd.markForCheck();
```
dtslvr: `markForCheck()` is only needed inside of `subscribe()`

dtslvr:
I like the idea to have a transparent concept, @mantovanig. Thanks for your contribution.

I have made some changes and don't show the data source as it is not so relevant for the user. What do you think?

dtslvr (approved):

PR #76, opened 2021-05-07
Feature/improve transaction filtering

vzickner:
2021-05-07 18 35 43 localhost 5d46e71dfa0d
2021-05-07 18 35 16 localhost dc00aea8e6ca

### DIFF
--- a/apps/client/src/app/components/transactions-table/transactions-table.component.ts
+++ b/apps/client/src/app/components/transactions-table/transactions-table.component.ts
@@ -1,6 +1,7 @@
 import {
   ChangeDetectionStrategy,
   Component,
+  ElementRef,
   EventEmitter,
   Input,
   OnChanges,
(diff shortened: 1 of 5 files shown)


mantovanig:
Hi @vzickner 👋 

I think that search parameters displayed as general input suggestions are not clear for the user. It's not easy to understand what user can search for and which fields the values refer to.

The scope of auto suggestions mechanism is not only to speed up the search process but above all to guiding the user helping them in constructing an advanced search query. 

Probably a better solution consist in move all the search parameters in ad-hoc drop down menu or dialog, like Jira.
telegram-cloud-document-4-5814148411720468934

What do you think?

In addition the style of the input component that you have refactored is a bit different from other applications inputs.

apps/client/src/app/components/transactions-table/transactions-table.component.ts
```diff
@@ -169,4 +227,40 @@ export class TransactionsTableComponent
...
+
+  private static getFilterableValues(
+    transaction,
+    fieldValues: Set<string> = new Set<string>()
+  ): string[] {
+    fieldValues.add(transaction.currency);
+    fieldValues.add(transaction.symbol);
+    fieldValues.add(transaction.type);
+    fieldValues.add(transaction.Account?.name);
+    fieldValues.add(transaction.Account?.Platform?.name);
```
mantovanig, line 263: If some of these fields can be undefined or null you have to check and you don't need to add to fieldValues, otherwise it can cause errors in the rest of the code you have written. 

Ex. https://github.com/ghostfolio/ghostfolio/pull/76/files#diff-23494c0c884566b94e8cf55674adf63fe2007424d0dfc8c765347d6691b395a0R239

vzickner: Good catch, thank you. I fixed it.

vzickner:
I really like the filtering in Jira. However, in my opinion for a mobile device one filter is better suited compared to multiple filters. When you have multiple filters the navigation is more complicated, and you lose a lot of space. But those are just my thoughts.

dtslvr:
Thanks a lot for your contribution, @vzickner, as well as your comment @mantovanig. I like this concept: it's quite powerful and close to natural search.

I have made a small tweak to the UI:
transaction-filter

dtslvr (approved):

PR #104, opened 2021-05-17
Feature/disable base currency selector for demo user

dtslvr:


vzickner:
How does it know that a new user has this new permission?

dtslvr:
> How does it know that a new user has this new permission?

The permission itself is not new:
https://github.com/ghostfolio/ghostfolio/blob/abd0e08566e639f6fa6bdc5ebde4187c7ea8a5d2/libs/common/src/lib/permissions.ts#L21

vzickner (approved):

PR #105, opened 2021-05-17
Feature/refactor search functionality

dtslvr:


apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -136,6 +138,49 @@ export class YahooFinanceService implements DataProviderInterface {
...
+        'json',
+        200
+      );
+
+      const result = await get();
+      items = result?.quotes
+        ?.filter((quote) => {
+          return quote.isYahooFinance;
+        })
+        .filter(({ quoteType }) => {
```
vzickner, line 157: is there no `?` needed here? The `.filter` before has one

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -136,6 +138,49 @@ export class YahooFinanceService implements DataProviderInterface {
     }
   }
 
+  public async search(aSymbol: string): Promise<{ items: LookupItem[] }> {
+    let items = [];
+
+    try {
+      const get = bent(
+        `https://query1.finance.yahoo.com/v1/finance/search?q=${aSymbol}&lang=en-US&region=US&quotesCount=8&newsCount=0&enableFuzzyQuery=false&quotesQueryId=tss_match_phrase_query&multiQuoteQueryId=multi_quote_single_token_query&newsQueryId=news_cie_vespa&enableCb=true&enableNavLinks=false&enableEnhancedTrivialQuery=true`,
```
vzickner, line 146: You might want to extract the hostname to a field to make testing with a mock HTTP server possible.

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -136,6 +138,49 @@ export class YahooFinanceService implements DataProviderInterface {
...
+        'GET',
+        'json',
+        200
+      );
+
+      const result = await get();
+      items = result?.quotes
+        ?.filter((quote) => {
```
vzickner, line 154: Any reason to not write `.filter((quote) => quote.isYahooFinance)`?

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
vzickner, line 146: are you sure that a user is not able to pass in a custom `aSymbol` which is overwriting other parameters?

apps/api/src/services/data-provider.service.ts
```diff
@@ -184,4 +185,22 @@ export class DataProviderService implements DataProviderInterface {
 
     return dataOfYahoo;
   }
+
+  public async search(aSymbol: string) {
+    let results: { items: LookupItem[] } = { items: [] };
+
+    switch (this.configurationService.get('DATA_SOURCES')[0]) {
```
vzickner, line 192: I would propose to extract the switch case which is determining the `dataProviderInterface` to use to a separate method and then call the `.search(...)` in here, outside the switch case. This will increase the re-usability once we have new methods.

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
dtslvr: For me, the verbose version is easier to read 😮

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
dtslvr: It's in a try / catch block, so none of them is really needed...

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
dtslvr: ✅

dtslvr: No. Do you have an idea how to escape the query parameter?

apps/api/src/services/data-provider.service.ts
dtslvr: I don't understand 😳 Can you give an example?

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
vzickner: I think a normal URL escape should be fine `encodeURIComponent`

apps/api/src/services/data-provider.service.ts
vzickner: Something like this:
```typescript
    public async search(aSymbol: string) {
        return this.getDataProvider().search(aSymbol);
    }

    private getDataProvider(): DataProviderInterface {
        switch (this.configurationService.get('DATA_SOURCES')[0]) {
            case DataSource.ALPHA_VANTAGE:
                return this.alphaVantageService;
                break;
            case DataSource.YAHOO:
                return this.yahooFinanceService;
                break;
            default:
                throw new Error('No data provider has been found.');
        }
    }
```

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
dtslvr: Okay, I will encode the query in the controller 👍🏻

apps/api/src/services/data-provider.service.ts
dtslvr: Okay, that looks nice. Thanks @vzickner! 😎

PR #119, opened 2021-05-24
Bugfix/fix performance chart

dtslvr:


dtslvr:
I have found the issue with your suspicious performance chart @gizmodus! 🤓

PR #123, opened 2021-05-24
optimize fetching from multiple data sources

vzickner:


apps/api/src/services/interfaces/data-provider.interface.ts
```diff
@@ -7,6 +7,8 @@ import {
 } from './interfaces';
 
 export interface DataProviderInterface {
+  hasHistoricalData(symbol: string): boolean;
```
dtslvr, line 10: I would rename it to something like `isActive(symbol: string): boolean;`, because it's not restricted to historical data, but if the data provider is used in general

apps/api/src/services/data-provider.service.ts
```diff
@@ -132,73 +130,45 @@ export class DataProviderService implements DataProviderInterface {
...
-          );
-          result[symbol] = data?.[symbol];
-          break;
-        }
+      const dataProvider = this.getDataProvider(dataSource);
+      if (dataProvider.hasHistoricalData(symbol)) {
+        promises.push(
+          dataProvider
+            .getHistorical([symbol], undefined, from, to)
+            .then((data) => ({ data: data?.[symbol], symbol }))
```
dtslvr, line 143: Can you refactor to `async` / `await`?

vzickner: No, `async` / `await` doesn't support multiple promises at the same time - if you want the unoptimized version, you just need to cherry-pick the first commit

apps/api/src/services/interfaces/data-provider.interface.ts
vzickner: Maybe `isActiveForSymbol`? Since otherwise the parameter is confusing

dtslvr: `canHandle` is fine! 👍🏻

dtslvr (approved):

PR #127, opened 2021-05-27
add yarn build:all to .travis.yml

vzickner:


dtslvr (changes requested):
Thanks for your contribution @vzickner 😃 

Are you aware of `yarn affected:build`? "Build applications and publishable libraries affected by changes" - https://nx.dev/latest/angular/cli/affected-build

It would also be possible to do `yarn affected:test` instead of `yarn test`. What do you think?

vzickner:
Would be an option, but it would then build/test nothing in the `main` branch since there are no differences to `main`. Not sure if that is good.

dtslvr (approved):

PR #161, opened 2021-06-14
Restrict webauthn to fingerprint only and improve UX

mfrey43:
Restrict webauthn to fingerprint only
Move webauthn login to separate page /webauthn where if it fails the user can try again or disable it

### DIFF
--- a/apps/client/src/app/components/header/header.component.html
+++ b/apps/client/src/app/components/header/header.component.html
@@ -1,9 +1,4 @@
 <mat-toolbar class="p-0">
-  <ng-container *ngIf="!user">
-    <a [routerLink]="['/']" class="no-min-width px-2" mat-button>
-      <gf-logo></gf-logo>
-    </a>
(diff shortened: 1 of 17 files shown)


apps/client/src/app/components/header/header.component.html
```diff
@@ -1,4 +1,9 @@
 <mat-toolbar class="p-0">
+  <ng-container *ngIf="!user">
+    <a [routerLink]="['/']" class="no-min-width px-2" mat-button>
+      <gf-logo></gf-logo>
+    </a>
+  </ng-container>
```
mfrey43, line 6: Suggestion: Shows logo when user is not signed in yet.

dtslvr: It's not so easy, because the logo should not be displayed on the landing page (big logo below)

dtslvr (approved):

PR #243, opened 2021-08-01
fix investment values for portfolio.service.ts

vzickner:


dtslvr (approved):
Thanks for the fix @vzickner! 😎

PR #264, opened 2021-08-07
Bugfix/add fallback if exchange service is not initialized

dtslvr:


apps/api/src/services/exchange-rate-data.service.ts
```diff
@@ -83,7 +84,12 @@ export class ExchangeRateDataService {
       factor = this.currencies[`${aFromCurrency}${aToCurrency}`];
     }
 
-    return factor * aValue;
+    if (isNumber(factor)) {
+      return factor * aValue;
+    }
+
+    // Fallback if currencies are not available
```
vzickner, line 91: Maybe add a warning to the log, since otherwise people might be confused?

dtslvr: Okay, I'm logging an error message now.

vzickner (approved):

PR #265, opened 2021-08-07
Feature/improve support for draft transactions

dtslvr:
draft-transaction

timeline-with-draft-transaction

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -38,7 +38,12 @@ import {
 } from '@ghostfolio/common/types';
 import { Inject, Injectable } from '@nestjs/common';
 import { REQUEST } from '@nestjs/core';
-import { Currency, DataSource, Type as TypeOfOrder } from '@prisma/client';
+import {
+  Currency,
+  DataSource,
+  Prisma,
+  Type as TypeOfOrder
+} from '@prisma/client';
 import Big from 'big.js';
 import {
   endOfToday,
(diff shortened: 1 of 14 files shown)


apps/client/src/app/components/investment-chart/investment-chart.component.ts
```diff
@@ -154,8 +154,8 @@ export class InvestmentChartComponent implements OnChanges, OnDestroy, OnInit {
     }
   }
 
-  private isInFuture(aContext: any, aValue: any) {
-    return isAfter(new Date(aContext?.p0?.parsed?.x), new Date())
+  private isInFuture<T>(aContext: any, aValue: T) {
```
vzickner, line 157: ❤️

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -750,16 +764,28 @@ export class PortfolioService {
...
+    userId: string;
+  }) {
+    const where: Prisma.OrderWhereInput = { userId };
+
+    if (includeDrafts === false) {
+      where.isDraft = false;
+    }
+
     return this.orderService.orders({
+      where,
```
vzickner, line 781: In my opinion the entire query shouldn't be in this service

vzickner (commented):
Looks good, but I think the order querying part should be inside the order service.

apps/api/src/app/portfolio/portfolio.service.ts
dtslvr: Okay, I have moved `getOrders()` to the order service.

PR #276, opened 2021-08-08
Feature/improve data gathering

dtslvr:


apps/api/src/app/cache/cache.module.ts
```diff
@@ -8,6 +15,16 @@ import { CacheService } from './cache.service';
 @Module({
   imports: [RedisCacheModule],
   controllers: [CacheController],
-  providers: [CacheService, PrismaService]
+  providers: [
+    AlphaVantageService,
+    CacheService,
+    ConfigurationService,
+    DataGatheringService,
```
dtslvr, line 22: This is annoying. I only need the `DataGatheringService`, but I have to provide all its dependencies as well 🤔 

Any idea @vzickner?

vzickner: Can you not just import another module which exports the services?

dtslvr: > Can you not just import another module which exports the services?

#278

PR #284, opened 2021-08-11
Fix position chart for missing historical data

dtslvr:
### Before

position-detail-chart-before

### After

position-detail-chart-after

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -362,8 +362,17 @@ export class PortfolioService {
       );
 
       const historicalDataArray: HistoricalDataItem[] = [];
-      let maxPrice = marketPrice;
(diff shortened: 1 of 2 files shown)


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -362,8 +362,17 @@ export class PortfolioService {
       );
 
       const historicalDataArray: HistoricalDataItem[] = [];
-      let maxPrice = marketPrice;
-      let minPrice = marketPrice;
+      let maxPrice = orders[0].unitPrice;
```
vzickner, line 365: Is this the correct currency?

dtslvr: Yes, that should be fine here 😃

vzickner (approved):

PR #285, opened 2021-08-12
Feature/improve portfolio summary

dtslvr:
### Before

portfolio-summary-before

### After

portfolio-summary-after


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -680,6 +655,42 @@ export class PortfolioService {
...
+      currency
+    );
+    const orders = await this.orderService.getOrders({ userId });
+    const fees = this.getFees(orders);
+    const firstOrderDate = orders[0]?.date;
+
+    const totalBuy = this.getTotalByType(orders, currency, TypeOfOrder.BUY);
+    const totalSell = this.getTotalByType(orders, currency, TypeOfOrder.SELL);
+
+    const committedFunds = new Big(totalBuy).sub(new Big(totalSell));
```
vzickner, line 675: You don't need the second `new Big(...)`

vzickner (approved):

apps/api/src/app/portfolio/portfolio.service.ts
dtslvr: Thank you for the hint! 🙏🏻

PR #326, opened 2021-09-01
Add a new symbol allocation chart

I-Valchev:
Adds a new chart for symbol allocation.

Also, IMHO, it may be a bit nicer to have a 3-column layout, instead of a 2-column one for the charts on this page. Can make a PR for it, if the maintainers agree.

image

### DIFF
--- a/apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
@@ -45,6 +45,7 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
   public symbols: {
     [name: string]: { name: string; value: number };
   };
+
   public user: User;
 
   private unsubscribeSubject = new Subject<void>();
@@ -110,13 +111,13 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
     };
     this.positions = {};
     this.positionsArray = [];
-    this.symbols = {
+    this.sectors = {
       [UNKNOWN_KEY]: {
         name: UNKNOWN_KEY,
         value: 0
       }
     };
-    this.sectors = {
+    this.symbols = {
       [UNKNOWN_KEY]: {
         name: UNKNOWN_KEY,
         value: 0
--- a/apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
@@ -49,6 +49,7 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
   public user: User;
 
   private unsubscribeSubject = new Subject<void>();
+
   /**
    * @constructor
    */
@@ -148,12 +149,6 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
       };
       this.positionsArray.push(position);
 
-      if (this.symbols[symbol] === undefined) {
-        this.symbols[symbol] = { name: symbol, value: position.value };
-      } else {
-        this.symbols[symbol].value += position.value;
-      }
-
       if (position.assetClass !== AssetClass.CASH) {
         // Prepare analysis data by continents, countries and sectors except for cash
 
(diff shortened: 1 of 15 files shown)


apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
```diff
@@ -42,10 +42,12 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
   public sectors: {
     [name: string]: { name: string; value: number };
   };
+  public symbols: {
+    [name: string]: { name: string; value: number };
+  };
   public user: User;
 
   private unsubscribeSubject = new Subject<void>();
-
```
dtslvr, line 48: Why do you remove this line break? 😄

apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
```diff
@@ -108,6 +110,12 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
     };
     this.positions = {};
     this.positionsArray = [];
+    this.symbols = {
```
dtslvr, line 113: Can you please move this below the `this.sectors` block?

apps/client/src/app/pages/portfolio/allocations/allocations-page.html
```diff
@@ -5,6 +5,30 @@ <h3 class="d-flex justify-content-center mb-3" i18n>Allocations</h3>
...
+            [defaultValue]="period"
+            [isLoading]="false"
+            [options]="periodOptions"
+            (change)="onChangePeriod($event.value)"
+          ></gf-toggle>
+        </mat-card-header>
+        <mat-card-content>
+          <gf-portfolio-proportion-chart
+            [baseCurrency]="user?.settings?.baseCurrency"
+            [isInPercent]="false"
```
dtslvr, line 22: I guess this should be `[isInPercent]="hasImpersonationId"` because you won't have absolute values in impersonation mode.

dtslvr (changes requested):
This is a great idea Ivo! Thanks for your contribution 👍🏻 

I have added some comments. Before merging, I will check it with my own portfolio.

Regarding the layout change, I think we have to play around with different screen sizes, especially for large screens 2 columns is not so nice.

apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
I-Valchev: Just playing around! I also ran the linters so the tests could pass. I'm putting it back :-)

apps/client/src/app/pages/portfolio/allocations/allocations-page.component.ts
```diff
@@ -42,10 +42,12 @@ export class AllocationsPageComponent implements OnDestroy, OnInit {
   public sectors: {
     [name: string]: { name: string; value: number };
   };
+  public symbols: {
+    [name: string]: { name: string; value: number };
+  };
   public user: User;
 
   private unsubscribeSubject = new Subject<void>();
```
I-Valchev, line 50:
```suggestion
  private unsubscribeSubject = new Subject<void>();

```

I-Valchev:
Thanks @dtslvr , I've adjusted. And I'm on the same page with you regarding the column layout. Maybe something like `col-md-6 col-lg-4` would do. You're better at UI stuff than I am, so I'll leave this with you.

dtslvr:
Hi @I-Valchev,

Have you seen my PR into your fork here https://github.com/I-Valchev/ghostfolio/pull/1? 😃

I-Valchev:
@dtslvr sorry this one slipped through the cracks...

Looks good. I don't have a strong opinion about whether to include cash in the piechart or not, but if you prefer it that way 👍

dtslvr:
> Thanks @dtslvr , I've adjusted. And I'm on the same page with you regarding the column layout. Maybe something like `col-md-6 col-lg-4` would do. You're better at UI stuff than I am, so I'll leave this with you.

https://github.com/ghostfolio/ghostfolio/pull/333

PR #333, opened 2021-09-03
Feature/restructure allocations page including allocations by symbol

dtslvr:
allocations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a link below the holdings to manage the transactions
 - Added the allocation chart by symbol
 
+### Changed
+
+- Restructured the allocations page
+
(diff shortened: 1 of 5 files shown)


dtslvr:
Do you think it's better now @I-Valchev?

I like that by _Account_, _Asset Class_ and _Currency_ are grouped and all the symbol related allocations are also together.

I-Valchev (approved):
👍 Yes, that looks great! 

On a related, though maybe separate note: can/should this pie chart component be adapted to show the % or total value for the big items? At the moment you need to hover over it. For countries it could use the 2 letter country codes/locales.

dtslvr:
> 👍 Yes, that looks great!

Thank you for the feedback.

> On a related, though maybe separate note: can/should this pie chart component be adapted to show the % or total value for the big items? At the moment you need to hover over it. For countries it could use the 2 letter country codes/locales.

I have created this ticket with an idea: #335

PR #341, opened 2021-09-04
Even out whitespace on the sides of the line chart

I-Valchev:
Makes the white space on the left and right "look" the same. Alternatively, it can be `mx-3` or no margin at all (but, then it may look incomplete).

### DIFF
--- a/apps/client/src/app/pages/zen/zen-page.html
+++ b/apps/client/src/app/pages/zen/zen-page.html
@@ -26,7 +26,7 @@
           class="chart-container d-flex flex-column col justify-content-center"
         >
           <gf-line-chart
-            class="mr-3"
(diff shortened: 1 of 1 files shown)


dtslvr:
Is it only an issue on the zen page or also on the home page?

A little screenshot would help here to illustrate your proposal 🙂

I-Valchev:
Sure, there it is, that gap on the right:
image

dtslvr:
Thanks for the screenshot @I-Valchev. Now I understand what you mean.

On the zen page, it is actually as intended "by design" 😛  But the chart on the home page is broken:

home

dtslvr:
> But the chart on the home page is broken

The layout has been improved recently. I'm closing this issue now.

PR #356, opened 2021-09-09
Fix tab buttons overlap on iOS `Add to Home screen`

I-Valchev:
On iOS, using the Add to Homescreen feature, the browser bar at the bottom doesn't show up, so the site takes the full screen. The bottom bar buttons then overlap with the white stripe for minimising the app. This adds a bit of margin to fix the issue :-)


dtslvr:
Hello @I-Valchev

Could you please attach a screenshot of the overlap? Thank you.

I-Valchev:
Hi @dtslvr , yes. Here it is:

image

PR #364, opened 2021-09-11
Feature/add annualized performance

dtslvr:


apps/api/src/app/portfolio/portfolio.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      null,
+      null,
+      null
+    );
+  });
+
+  /**
+   * Source: https://www.investopedia.com/terms/a/annualized-total-return.asp#annualized-return-formula-and-calculation
+   */
+  fit('Get annualized performance', async () => {
```
vzickner, line 23: The f shouldn't be here

apps/api/src/app/portfolio/portfolio.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    ).toEqual(0);
+
+    expect(
+      portfolioService.getAnnualizedPerformancePercent({
+        daysInMarket: 0,
+        netPerformancePercent: 0
+      })
+    ).toEqual(0);
+
+    expect(
```
vzickner, line 38: Maybe another test with 365 days?

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -80,6 +81,21 @@ export class PortfolioService {
...
+  public getAnnualizedPerformancePercent({
+    daysInMarket,
+    netPerformancePercent
+  }: {
+    daysInMarket: number;
+    netPerformancePercent: number;
+  }) {
+    if (isNumber(daysInMarket) && daysInMarket > 0) {
+      const exponent = new Big(365).div(daysInMarket).toNumber();
+      return Math.pow(1 + netPerformancePercent, exponent) - 1;
```
vzickner, line 93: Why not with big?

dtslvr: `pow` only works with `int` in `Big.js` → https://mikemcl.github.io/big.js/#pow

apps/api/src/app/portfolio/portfolio.service.spec.ts
dtslvr: Good idea! 👍🏻

apps/api/src/app/portfolio/portfolio.service.spec.ts
dtslvr: Eagle eye! Thank you @vzickner 🙂

apps/api/src/app/portfolio/portfolio.service.ts
vzickner: Sad

PR #371, opened 2021-09-15
Smooth update transitions in allocation charts

mfrey43:
By updating the individual datasets of the allocation charts we can smooth out the transition between 'initial' and 'current' view modes. 

For this to work we need to somehow map the new data to the old one. The current implementation only requires that the label and dataset arrays remain the same size.

smoot_allocation


vzickner:
Nice feature, I am looking forward to it :)

dtslvr:
Stale PR, closed.

PR #387, opened 2021-09-21
Feature/support unlimited currencies

dtslvr:


dtslvr:
**Current limitation**
* Adding a transaction in a new currency does not reload the exchange rates correctly

apps/api/src/services/exchange-rate-data.service.ts
```diff
@@ -1,27 +1,45 @@
...
-  public constructor(private dataProviderService: DataProviderService) {
+  public constructor(
+    private readonly dataProviderService: DataProviderService,
+    private readonly prismaService: PrismaService
+  ) {
     this.initialize();
   }
 
+  public getCurrencies() {
+    return this.currencies?.length > 0 ? this.currencies : [baseCurrency];
```
vzickner, line 26: base currency might not be part of `this.currencies`, or? I just try to understand why the `? :` is required

dtslvr: If the instance has no transactions and no users, it could be problematic if there is no currency at all

PR #431, opened 2021-10-20
add Dockerfile with all in one docker image

vzickner:
The size of the docker image is around ~500 MB. It is based on the alpine image, which is the reason why the seed needs to be a JS instead of TS file. However, there are no significant changes since no typings are used in there anyway.


dtslvr (approved):

docker/docker-compose-build-local.yml
```diff
@@ -0,0 +1,24 @@
...
+  redis:
+    image: 'redis:alpine'
+
+  ghostfolio:
+    build: ../
+    env_file:
+      - ../.env
+    environment:
+      REDIS_HOST: 'redis'
+      DATABASE_URL: postgresql://user:password@postgres:5432/ghostfolio-db?sslmode=prefer
```
dtslvr, line 19: This is already part of the `.env` file (https://github.com/ghostfolio/ghostfolio/blob/main/.env#L14), but I don't know how to eliminate 😞

PR #480, opened 2021-11-19
add support for euro cryptocurrencies, ALGO and remove unknown crypto…

vzickner:
…currencies from list

### DIFF
--- a/apps/api/src/services/cryptocurrency/custom-cryptocurrencies.json
+++ b/apps/api/src/services/cryptocurrency/custom-cryptocurrencies.json
@@ -1,5 +1,6 @@
 {
   "1INCH": "1inch",
+  "ALGO": "Algorand",
   "AVAX": "Avalanche",
   "MATIC": "Polygon",
   "SHIB": "Shiba Inu"
--- a/apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
+++ b/apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
@@ -197,16 +197,17 @@ export class YahooFinanceService implements DataProviderInterface {
           // filter out undefined symbols
           return quote.symbol;
         })
-        .filter(({ quoteType }) => {
+        .filter(({ quoteType, symbol }) => {
           return (
-            quoteType === 'CRYPTOCURRENCY' ||
+            (quoteType === 'CRYPTOCURRENCY' && this.isCryptoCurrency(symbol)) ||
             quoteType === 'EQUITY' ||
             quoteType === 'ETF'
           );
         })
         .filter(({ quoteType, symbol }) => {
           if (quoteType === 'CRYPTOCURRENCY') {
-            // Only allow cryptocurrencies in USD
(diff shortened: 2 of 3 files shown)


apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -197,17 +197,17 @@ export class YahooFinanceService implements DataProviderInterface {
...
+            (quoteType === 'CRYPTOCURRENCY' && this.isCryptoCurrency(symbol)) ||
             quoteType === 'EQUITY' ||
             quoteType === 'ETF'
           );
         })
         .filter(({ quoteType, symbol }) => {
           if (quoteType === 'CRYPTOCURRENCY') {
-            // Only allow cryptocurrencies in USD
-            return symbol.includes('USD');
+            // Only allow cryptocurrencies in USD and EUR
```
dtslvr, line 209: I only want to support _Bitcoin_ once (in `USD`). Next users would add `CAD`, then `CHF` and so on... A lot of redundant data.

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -253,21 +253,26 @@ export class YahooFinanceService implements DataProviderInterface {
...
         // SOL1USD -> SOL1-USD
-        return aSymbol.replace('USD', '-USD');
+        return aSymbol.replace(new RegExp('-?(USD|EUR)$'), '-$1');
       }
     }
 
     return aSymbol;
   }
 
+  private isCryptoCurrency(aSymbol: string) {
```
dtslvr, line 268: What do you want to improve exactly?

The current implementation is hacky, but now we would have to deal with `isCrypto()` and `isCryptoCurrency()` and two regular expressions.

apps/api/src/services/cryptocurrency/custom-cryptocurrencies.json
```diff
@@ -1,5 +1,6 @@
 {
   "1INCH": "1inch",
+  "ALGO": "Algorand",
```
dtslvr, line 3: 👍🏻

dtslvr (changes requested):

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
vzickner: There are two usages of this method. See line 202.
Thie idea is to not have cryptocurrencies in the search result which you don't consider as a cryptocurrency.
If you don't like it, it might be an option to do the check with the `1` replacement in the `cryptocurrencyService`

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
vzickner: Ok, makes sense.

dtslvr (approved):

PR #536, opened 2021-12-11
Feature/remove transactions import limit

dtslvr:


apps/api/src/services/configuration.service.ts
```diff
@@ -27,6 +27,7 @@ export class ConfigurationService {
       GOOGLE_SECRET: str({ default: 'dummySecret' }),
       JWT_SECRET_KEY: str({}),
       MAX_ITEM_IN_CACHE: num({ default: 9999 }),
+      MAX_ORDERS_TO_IMPORT: num({ default: Number.MAX_SAFE_INTEGER }),
```
vesely-david, line 30: Do I assume correctly that `Number.MAX_SAFE_INTEGER` is "infinity"?

vesely-david (approved):
LGTM

apps/api/src/services/configuration.service.ts
dtslvr: Exactly. According to [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)`Number.MAX_SAFE_INTEGER` _represents the maximum safe integer in JavaScript_.

PR #541, opened 2021-12-13
fix(custom-cryptocurrencies): add uniswap (uni3)

### LINKED ISSUE #539: Cannot get symbol for Uniswap
Hey guys!

I was trying to insert some transactions on Uniswap but the API seems to not find the symbol.
Here's a screenshot for the console:
Screenshot 2021-12-13 at 10 06 52

And here are the requests:
Screenshot 2021-12-13 at 10 07 20

When trying to find directly on Yahoo Finance (https://finance.yahoo.com/quote/UNI3-USD/) there's a hyphen on the URL - maybe that's the problem.

OliRafa:
Fix #539


dtslvr (approved):
Thank you 👍🏻

PR #545, opened 2021-12-15
build(prisma): fix populating database when attributes have null value

OliRafa:
This PR fixes database population.

When trying to populate the database for the first time, prisma complained about some attributes having `null` value.
In this PR we changed these attributes' values from `null` to `undefined` in order to fix this issue.


dtslvr (approved):
👍🏻

PR #632, opened 2022-01-17
Portfolio calculator rework

gizmodus:


apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -134,6 +148,14 @@ export class PortfolioCalculator {
   }
 
   public async getCurrentPositions(start: Date): Promise<CurrentPositions> {
+    if (true) {
```
dtslvr, line 151: In the calculator you do not need conditionals (with the factory described above)

apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -344,6 +366,406 @@ export class PortfolioCalculator {
...
+      marketSymbolMap[format(start, DATE_FORMAT)]?.[symbol];
+
+    const unitPriceAtEndDate =
+      marketSymbolMap[format(endDate, DATE_FORMAT)]?.[symbol];
+
+    if (!unitPriceAtStartDate || !unitPriceAtEndDate) {
+      hasErrors = true;
+    }
+
+    // Add a placeholder order at the start and the end date
```
dtslvr, line 570: // Add a **synthetic** placeholder order... ❓

apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -344,6 +366,406 @@ export class PortfolioCalculator {
...
+      itemType: 'end',
+      name: '',
+      quantity: new Big(0),
+      type: TypeOfOrder.BUY,
+      unitPrice: unitPriceAtEndDate ?? new Big(0)
+    });
+
+    // Sort orders so that the start and end placeholder order are at the right
+    // position
+    orders = sortBy(orders, (order) => {
```
dtslvr, line 599: Why don't you sort first (do you really have to?) and prepend / append the synthetic orders?

dtslvr (changes requested):

apps/api/src/app/portfolio/interfaces/portfolio-calculator.interface.ts
```diff
@@ -0,0 +1,5 @@
+import { PortfolioOrder } from './portfolio-order.interface';
+
+export interface PortfolioOrderItem extends PortfolioOrder {
+  itemType?: '' | 'start' | 'end';
```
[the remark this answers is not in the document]
gizmodus: I chose it like that because it should not be possible to make it a start and end date at the same time. The string type acts as an enum.

apps/api/src/app/portfolio/portfolio-calculator.ts
gizmodus: 👍

apps/api/src/app/portfolio/portfolio-calculator.ts
gizmodus: Because the synthetic start order is not the first order if there are transaction before the start date.

dtslvr (approved):

PR #654, opened 2022-01-28
import csv with account name or id

ronaldkonjer:
import CSV using an account column to assign the transactions/orders to a specific account


apps/client/src/app/services/import-transactions.service.ts
```diff
@@ -4,14 +4,16 @@ import { CreateOrderDto } from '@ghostfolio/api/app/order/create-order.dto';
 import { DataSource, Type } from '@prisma/client';
 import { parse } from 'date-fns';
 import { isNumber } from 'lodash';
-import { parse as csvToJson } from 'papaparse';
+import { Papa } from 'ngx-papaparse';
```
dtslvr, line 7: What is the benefit of using `ngx-papaparse` over `papaparse`?

I would prefer to stick with `papaparse`, because it can be upgraded independently from _Angular_.

apps/client/src/app/services/import-transactions.service.ts
```diff
@@ -20,28 +22,32 @@ export class ImportTransactionsService {
...
     fileContent,
     primaryDataSource
   }: {
-    defaultAccountId: string;
+    user: User;
     fileContent: string;
     primaryDataSource: DataSource;
   }) {
-    const content = csvToJson(fileContent, {
+    let content;
```
dtslvr, line 36: `let content: any[] = [];` ❓

apps/client/src/app/services/import-transactions.service.ts
```diff
@@ -4,14 +4,16 @@ import { CreateOrderDto } from '@ghostfolio/api/app/order/create-order.dto';
...
+import { Papa } from 'ngx-papaparse';
 import { EMPTY } from 'rxjs';
 import { catchError } from 'rxjs/operators';
+import { User } from '@ghostfolio/common/interfaces';
 
 @Injectable({
   providedIn: 'root'
 })
 export class ImportTransactionsService {
+  private static ACCOUNT_ID = ['account', 'accountid'];
```
dtslvr, line 16: 👍🏻

apps/client/src/app/pages/portfolio/transactions/transactions-page.component.ts
```diff
@@ -204,8 +203,8 @@ export class TransactionsPageComponent implements OnDestroy, OnInit {
           } else if (file.name.endsWith('.csv')) {
             try {
               await this.importTransactionsService.importCsv({
+                user: this.user,
```
dtslvr, line 206: Can you move `user` down? For consistency, we are ordering attributes alphabetically 🙂

apps/client/src/app/services/import-transactions.service.ts
```diff
@@ -20,28 +22,32 @@ export class ImportTransactionsService {
   private static TYPE_KEYS = ['action', 'type'];
   private static UNIT_PRICE_KEYS = ['price', 'unitprice', 'value'];
 
-  public constructor(private http: HttpClient) {}
+  public constructor(private http: HttpClient, private papa: Papa) {}
 
   public async importCsv({
-    defaultAccountId,
+    user,
```
dtslvr, line 28: Also here...

apps/client/src/app/services/import-transactions.service.ts
```diff
@@ -20,28 +22,32 @@ export class ImportTransactionsService {
...
     fileContent,
     primaryDataSource
   }: {
-    defaultAccountId: string;
+    user: User;
```
dtslvr, line 32: ...and here, move `user` to the end

dtslvr (changes requested):

apps/client/src/app/pages/portfolio/transactions/transactions-page.component.ts
ronaldkonjer: now i'm aware

apps/client/src/app/services/import-transactions.service.ts
ronaldkonjer: removed

apps/client/src/app/services/import-transactions.service.ts
ronaldkonjer: changed

PR #750, opened 2022-03-12
Feature/fix issue with recent transactions

gizmodus:


dtslvr:
The test `PortfolioCalculatorNew › get current positions › with BALN.SW buy` fails and I think it's indeed not correct:

```
2021-11-30 BUY for 136.6
2021-12-18 Market Price: 148.9
```

`148.9 / 136.6 * 100 = 109.004392387` -> `grossPerformancePercentage = 0.09004392386530014641` (before)
With the adapted calculation in this branch, the `grossPerformancePercentage` is only `0.04502196193265007321`.

PR #751, opened 2022-03-12
Add type

dtslvr:


dtslvr:
@gadicc Is this the proper way to reuse types? The path looks a bit odd 😯

gadicc:
Oh wow, nice!! :raised_hands: 

Yeah, that's exactly right.  In the future when we drop CommonJS support, the paths will get a lot simpler.  And then I guess we can provide a codemod to help everyone update their paths.  Also, you can `import type` vs just `import` to be explicit that that's what you want, it can help your toolchain understand your needs better.

Amazing that you decided to jump straight in, let me know how it goes.  Definitely don't hesitate to "gradually" adopt and only move bits and pieces of your project from `.js` to `.ts`.  My initial experience was "agghhhh why is TypeScript such a pain" but the more you use it, the more you appreciate it, especially when you argue with it only to realise it totally helped you avoid a mistake in your code.  My experience at least :)  Good luck!

dtslvr:
> Oh wow, nice!! 🙌
> 
> Yeah, that's exactly right. In the future when we drop CommonJS support, the paths will get a lot simpler. And then I guess we can provide a codemod to help everyone update their paths. Also, you can `import type` vs just `import` to be explicit that that's what you want, it can help your toolchain understand your needs better.
> 
> Amazing that you decided to jump straight in, let me know how it goes. Definitely don't hesitate to "gradually" adopt and only move bits and pieces of your project from `.js` to `.ts`. My initial experience was "agghhhh why is TypeScript such a pain" but the more you use it, the more you appreciate it, especially when you argue with it only to realise it totally helped you avoid a mistake in your code. My experience at least :) Good luck!

Thanks for the hint @gadicc! I have refactored `import` to `import type` 🤓

PR #758, opened 2022-03-16
Bugfix/fix get quotes for multiple ghostfolio symbols

dtslvr:


apps/api/src/services/data-provider/ghostfolio-scraper-api/ghostfolio-scraper-api.service.ts
```diff
@@ -84,33 +84,42 @@ export class GhostfolioScraperApiService implements DataProviderInterface {
...
   ): Promise<{ [symbol: string]: IDataProviderResponse }> {
+    const response: { [symbol: string]: IDataProviderResponse } = {};
+
     if (aSymbols.length <= 0) {
-      return {};
+      return response;
     }
 
     try {
-      const [symbol] = aSymbols;
```
dtslvr, line 92: @FedericoComoglio The `getQuotes()` function for symbols with data source `GHOSTFOLIO` was just supporting a single symbol 🙈
As soon as you have added one more activity, the calculation was broken.

PR #946, opened 2022-05-23
Add REDIS_PASSWORD support into docker files

willyp713:
Added REDIS_PASSWORD variable but left blank because not everyone will have one.

### DIFF
--- a/.env
+++ b/.env
@@ -3,6 +3,7 @@ COMPOSE_PROJECT_NAME=ghostfolio-development
 # CACHE
 REDIS_HOST=localhost
 REDIS_PORT=6379
+REDIS_PASSWORD=
 
 # POSTGRES
 POSTGRES_DB=ghostfolio-db
--- a/docker/docker-compose.build.yml
+++ b/docker/docker-compose.build.yml
@@ -7,6 +7,7 @@ services:
     environment:
       DATABASE_URL: postgresql://user:password@postgres:5432/ghostfolio-db?sslmode=prefer
       REDIS_HOST: 'redis'
+      REDIS_PASSWORD: 'password'
     ports:
       - 3333:3333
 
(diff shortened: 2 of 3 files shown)


LeonStoldt:
Hi,
@dtslvr FYI this redis password feature was requested in the unraid support thread and I asked to create an issue / feature request for this one.

@willyp713 I believe that it is not that easy to just provide an env variable (Thomas pls correct me if I am wrong).
I think you need to pass the password to the redis container (in unraid it will be done by modifying the unraid container but to ensure it works with docker compose and in development it should be added and tested with the docker/docker-compose* files). If redis knows the password, ghostfolio need to know aswell - so pass it to the ghostfolio container via env variable (also compose files), if necessary add env in ghostfolio Dockerfile for running via docker cli command and you need to ensure ghostfolio can programatically connect to redis with a password. You need to find the inplementation and (as I think it is set to NoAuth currently) this part needs to check and authenticate if a password is provided. Maybe write some test to ensure working code and add this information to the Readme.

willyp713:
> Hi, @dtslvr FYI this redis password feature was requested in the unraid support thread and I asked to create an issue / feature request for this one.
> 
> @willyp713 I believe that it is not that easy to just provide an env variable (Thomas pls correct me if I am wrong). I think you need to pass the password to the redis container (in unraid it will be done by modifying the unraid container but to ensure it works with docker compose and in development it should be added and tested with the docker/docker-compose* files).
> 
> Might be helpful: https://nickjanetakis.com/blog/docker-tip-27-setting-a-password-on-redis-without-a-custom-config
> 
> If redis knows the password, ghostfolio need to know aswell - so pass it to the ghostfolio container via env variable (also compose files), if necessary add env in ghostfolio Dockerfile for running via docker cli command and you need to ensure ghostfolio can programatically connect to redis with a password. You need to find the inplementation and (as I think it is set to NoAuth currently) this part needs to check and authenticate if a password is provided. Maybe write some test to ensure working code and add this information to the Readme.

I actually figured it was going to be more complicated than this but was trusting in the community to point out what else was needed. Thanks @LeonStoldt.

willyp713:
Moved to #947

PR #947, opened 2022-05-24
Feature/add-redis-password

willyp713:
Starting work on adding redis password support. I started updating .env and the docker compose files, but not 'dev' under the assumption that it won't be password protected. But there is probably still more work and testing to be done, per #946 and this comment from @LeonStoldt:

> Hi, @dtslvr FYI this redis password feature was requested in the unraid support thread and I asked to create an issue / feature request for this one.
> 
> @willyp713 I believe that it is not that easy to just provide an env variable (Thomas pls correct me if I am wrong). I think you need to pass the password to the redis container (in unraid it will be done by modifying the unraid container but to ensure it works with docker compose and in development it should be added and tested with the docker/docker-compose* files).
> 
> Might be helpful: https://nickjanetakis.com/blog/docker-tip-27-setting-a-password-on-redis-without-a-custom-config
> 
> If redis knows the password, ghostfolio need to know aswell - so pass it to the ghostfolio container via env variable (also compose files), if necessary add env in ghostfolio Dockerfile for running via docker cli command and you need to ensure ghostfolio can programatically connect to redis with a password. You need to find the inplementation and (as I think it is set to NoAuth currently) this part needs to check and authenticate if a password is provided. Maybe write some test to ensure working code and add this information to the Readme.


willyp713:
I added REDIS_PASSWORD as a variable in the compose files such that they will reference the .env file. Not sure if you want to set it up that way, given that the rest of the compose files don't use variables like that.

dtslvr:
Thanks for creating a pull request @willyp713.

Variable substitutions are fine for me. I don't think it's not enough to just add it in `.env` and the docker compose files.

Please have a look at these files:

* https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/configuration.service.ts#L38
* https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/interfaces/environment.interface.ts#L29
* https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/app.module.ts#L43
* https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/redis-cache/redis-cache.module.ts#L17

willyp713:
> 

Agreed and this starts to get outside my comfort zone. My first guess is that we can add `REDIS_PASSWORD` as a `string` to those files but I'm not sure if it's that simple.

willyp713:
I took a stab at editing the files suggested by @dtslvr and added `REDIS_PASSWORD` as a string to each of them, with default being blank. Trying to build this locally to test but would appreciate another set of eyes if someone has a chance.

dtslvr:
> I took a stab at editing the files suggested by @dtslvr and added `REDIS_PASSWORD` as a string. Trying to build this locally to test but would appreciate another set of eyes if someone has a chance.

It looks fine to me. Thank you @willyp713!

PR #994, opened 2022-06-09
Bugfix/force reload accounts of user after change

dtslvr:


apps/client/src/app/services/user/user.service.ts
```diff
@@ -16,13 +16,13 @@ export class UserService extends ObservableStore<UserStoreState> {
   public constructor(private http: HttpClient) {
     super({ trackStateHistory: true });
 
-    this.setState({ user: undefined }, 'INIT_STATE');
+    this.setState({ user: undefined }, UserStoreActions.Initialize);
   }
 
-  public get() {
+  public get(force = false) {
```
dtslvr, line 22: What do you think of this flag to force reload the user @cnschwarz?

cnschwarz: Yes I think that is a good approach for now. An alternative would be to monitor a BehaviourSubject and in case of an account update trigger an update on the BehaviourSubject on which all subscribers would get updated. But that would require a couple more changes in the existing structure.

dtslvr: Thank you for your feedback @cnschwarz 🙂

Issue #1004, opened 2022-06-12
Acessing Ghostfolio behind a reverse-proxy in Docker


GitBOUF:
Hello,

I've currently setup Ghostfolio in Docker behind a Traefik 2 container. All other apps in my docker-compose works behind Traefik, except Ghostfolio. I've noticed that the docker image pulled from DockerHub for Ghostfolio is currently setup to listen by default on localhost. Is there a way to change it to listen on 0.0.0.0 instead?

Since we can change the port with the PORT environment variable, I've tried to setup the HOST environment variable, but no success.
PORT=3333
HOST=0.0.0.0

And the logs keep showing on every restart:
`Listening at http://localhost:3333`

Just like we can set the PORT environment variable, having a HOST environment variable would be useful for changing the interface to listen on.

Here's how Ghostfolio is setup in my docker-compose:
``` YAML
  ghostfolio:
    image: ghostfolio/ghostfolio:latest
    container_name: ghostfolio
    networks:
      - traefik_proxy
      # Use an internal network for the database and redis
      - ghostfolio_network
     #No need to open the port here, Traefik handles it.
     #ports:
     # - 3333:3333
    depends_on:
      - ghostfolio-db
      - ghostfolio-redis
    environment:
      PORT: 3333
      HOST: 0.0.0.0
      REDIS_HOST: ghostfolio-redis
      REDIS_PORT: 6379
      REDIS_PASSWORD: ${GHOSTFOLIO_REDIS_PASSWORD}
      POSTGRES_DB: ghostfolio
      POSTGRES_USER: ghostfolio
      POSTGRES_PASSWORD: ${GHOSTFOLIO_DB_PASSWORD}
      POSTGRES_PORT: 5432
      DATABASE_URL: postgresql://ghostfolio:${GHOSTFOLIO_DB_PASSWORD}@ghostfolio-db:5432/ghostfolio?sslmode=prefer
      ACCESS_TOKEN_SALT: ${GHOSTFOLIO_ACCESS_TOKEN_SALT}
      JWT_SECRET_KEY: ${GHOSTFOLIO_JWT_SECRET_KEY}
    labels:
      - "traefik.enable=true"
      ## HTTP Routers Auth
      - "traefik.http.routers.ghostfolio-rtr.entrypoints=websecure"
      - "traefik.http.routers.ghostfolio-rtr.rule=Host(`ghostfolio.$DOMAINNAME`)"
      ## Middlewares
[...]

PR #1013, opened 2022-06-13
Feature/change default host

dtslvr:


kikoenai:
I had the same problem as #1004 , thank you for the fix! :-) It was not so obvious for me to realise that HOST was exposed. It might make sense to add which env variables are required/optional to the self host section or use the HOST keyword in the example .env file. I can perhaps do this and add to the README if this is helpful to you?

dtslvr:
@kikoenai a table like [this](https://github.com/ghostfolio/ghostfolio#body) with all the environment variables and a description would be helpful. Adding all entries from [here](https://github.com/ghostfolio/ghostfolio/blob/main/.env) and additionally `BASE_CURRENCY`, `CACHE_TTL`, `HOST`, `MAX_ITEM_IN_CACHE` and `PORT`.

What do you think?

kikoenai:
@dtslvr Yes I agree with the way you would put it! I would suggest the table format to be something like:

| Variable | Description | Format |
|---|---|---|

Where Description also mentions the default value.

dtslvr:
Format is not needed I think.
What about Variable, Default Value, Description?

kikoenai:
Yes I think that can work too! I was wondering about the format since maybe it is not so clear for a user to choose  eg. for base_currency to write EUR or Euro etc. It might be obvious with the default value though.

Jun 14, 2022 22:11:04 Thomas Kaul ***@***.***>:

dtslvr:
@kikoenai I have just added the [supported environment variables](https://github.com/ghostfolio/ghostfolio#supported-environment-variables) to the documentation.

PR #1080, opened 2022-07-16
Feature/setup internet identity

dtslvr:
ghostfolio-registration


dtslvr:
Hi @samlinux

My friend @peterpeterparker has sent me your repository [angularMeetsII](https://github.com/samlinux/angularMeetsII) as an inspiration to setup _Internet Identity_ 👍🏻

apps/api/src/app/auth/auth.service.ts
```diff
@@ -40,6 +41,33 @@ export class AuthService {
     });
   }
 
+  public async validateInternetIdentityLogin(principalId: string) {
+    try {
+      const provider: Provider = 'INTERNET_IDENTITY';
+
+      let [user] = await this.userService.users({
```
dtslvr, line 48: @samlinux With this implementation, it would be possible to randomly find a user by its principal id. Is it safe anyway (because e.g. the principle id is long enough)?

apps/client/src/app/services/internet-identity.service.ts
```diff
@@ -0,0 +1,42 @@
...
+import { catchError } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class InternetIdentityService {
+  public constructor(private http: HttpClient) {}
+
+  public async login(): Promise<OAuthResponse> {
+    const authClient = await AuthClient.create();
```
peterpeterparker, line 15: FYI per default `AuthClient` observes misc events and automatically sign-out after a couple of minutes (like 10) if user is not active.

If you wish to deactivate this behavior (as I do i Papyrs), you can create it as following:

```
const createAuthClient = (): Promise<AuthClient> =>
  AuthClient.create({
    idleOptions: {
      disableIdle: true,
      disableDefaultIdleCallback: true
    }
  });
```

apps/client/src/app/services/internet-identity.service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    return await new Promise((resolve, reject) => {
+      authClient.login({
```
peterpeterparker, line 18: Here also, the default delegation least 10min (I think or something).

If you wish to defined a longer session you can do as following:

```
// How long the delegation identity should remain valid?
// e.g. BigInt(7 * 24 * 60 * 60 * 1000 * 1000 * 1000) = 7 days in nanoseconds
// For Papyrs: 4 hours
const delegationIdentityExpiration: bigint = BigInt(4 * 60 * 60 * 1000 * 1000 * 1000);

await authClient.login({
    maxTimeToLive: delegationIdentityExpiration,
    ...
```

apps/api/src/app/auth/auth.service.ts
peterpeterparker: Worth to note that the principal ID is generated anonymously and is different for each domain - i.e. you cannot track principal ID across websites

e.g. a user that uses Internet Identity with an anchor 1234 would get a principal ID 5678 for Ghostfolio but abcd for Papyrs

apps/client/src/app/services/internet-identity.service.ts
```diff
@@ -0,0 +1,42 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class InternetIdentityService {
+  public constructor(private http: HttpClient) {}
+
+  public async login(): Promise<OAuthResponse> {
+    const authClient = await AuthClient.create();
+
+    return await new Promise((resolve, reject) => {
```
peterpeterparker, line 17: not sure you need `await new`, maybe `new Promise...` is enough?

apps/client/src/app/services/internet-identity.service.ts
```diff
@@ -0,0 +1,42 @@
...
+          return reject();
+        },
+        onSuccess: () => {
+          const principalId = authClient.getIdentity().getPrincipal();
+
+          this.http
+            .get<OAuthResponse>(
+              `/api/v1/auth/internet-identity/${principalId.toText()}`
+            )
+            .pipe(
```
peterpeterparker, line 29: you don't unsubscribe here so there is a possible memory leak

`.pipe(first(), ...` or `takeUntil` etc.

peterpeterparker (commented):
Added few comments, hope it helps

Cool to see II used in a web2 project as well 👍

apps/client/src/app/services/internet-identity.service.ts
dtslvr: Not sure if I need it 🤔  I only use the `AuthClient` to get the `principalId`, then I generate a custom `jwt` with expiration time.

apps/client/src/app/services/internet-identity.service.ts
dtslvr: See above.

apps/client/src/app/services/internet-identity.service.ts
dtslvr: Great catch! `return new Promise()` will do it.

apps/client/src/app/services/internet-identity.service.ts
dtslvr: Oh, I did not know that it's also possible in services with `providedIn: 'root'` 🙂

apps/api/src/app/auth/auth.service.ts
dtslvr: > Added few comments, hope it helps
> 
> Cool to see II used in a web2 project as well 👍

Thank you very much for taking the time to give your feedback, @peterpeterparker! 😎

dtslvr:
> Added few comments, hope it helps
> 
> Cool to see II used in a web2 project as well 👍

Thank you very much for taking the time to give your feedback, @peterpeterparker! 😎

apps/client/src/app/services/internet-identity.service.ts
peterpeterparker: In that case I would disable the checker - i.e. you don't need to attach these listener if in any case you are not going to check if the delegation of internet identity expires or not.

peterpeterparker (approved):
Super

apps/client/src/app/services/internet-identity.service.ts
dtslvr: > In that case I would disable the checker - i.e. you don't need to attach these listener if in any case you are not going to check if the delegation of internet identity expires or not.

✅  Done

PR #1134, opened 2022-08-09
Build arm64 docker image and use github actions

Areedtz:
This pull request will change the project to use github actions instead of Travis CI for CI/CD.

Furthermore, it changes the docker image to also be built for the arm64 architecture instead of only amd64.

This is a draft pull request, as it currently has some kinks to iron out:
  - A non-cached build of the docker image can take [up to 90 minutes](https://github.com/Areedtz/ghostfolio/actions/runs/2824844958). If fully cached, it only takes [1 minute](https://github.com/Areedtz/ghostfolio/actions/runs/2825343083)
  - I had to change the dockerfile to use the ``slim`` variant instead of ``alpine``, which is about 60MB larger
  - It currently points to a grobut939/ghostfolio image I set up temporarily to test the workflow
  - Currently doesn't tag the docker image with a version

It will also be needed to set up secrets in the repo before the github actions will work.

In regards to using the ``slim`` variant instead of ``alpine``. We could work around this by having two separate dockerfiles. One for alpine and one for debian (slim). This will however mean that users will need to use the debian dockerfile for arm (which probably won't be the default ``latest`` tag, but instead a ``debian`` tag.


dtslvr:
Hey @Areedtz

That looks promising, thanks a lot for your contribution.

Is it possible to keep _Travis_ for the moment and just build the `arm64` image with _GitHub Actions_? _Travis_ runs smoothly and I could switch gradually to the new build process.

* Regarding caching: when is the build slow and when is it fast?
* I think we can use `slim` instead of `alpine`
* Tag could be something like `linux-arm64-1.770.0`
* [Here](https://hub.docker.com/r/portainer/portainer/tags) it looks like you can have `latest` tag for different platforms 🤔

Areedtz:
> Is it possible to keep Travis for the moment and just build the ``arm64`` image with GitHub Actions? Travis runs smoothly and I could switch gradually to the new build process.

As in have the ``amd64`` image built via Travis and the ``arm64`` image built via Github Actions? I'm not quite sure. I'm still new to these multi-arch docker images.
According to the [original docker blog](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) it seems like it is possible. I'm however not sure if it's a good idea to do it using 2 systems, as I could imagine a race condition could occur. I would probably recommend to only use either Travis or Github actions, or at the very least only build the docker image via one of them.

> Regarding caching: when is the build slow and when is it fast?

It seems like the ``arm64`` build specifically is the slow part as the ``yarn install`` and ``yarn build`` commands each run for around 30 minutes. The ``amd64`` image is fast to build, only running for about 10 minutes. 

It might actually be possible to speed this up, if we can build make the build step always run as ``amd64`` but build for both platforms like they're talking about in [this blog](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/). We might not even need to do actually run the build step on ``arm64``. I unfortunately don't have time to test that for now.

In general, the cache works just like when you build the docker images locally. So it will only rebuild a docker layer if changes has been made to it. So looking at the ``Dockerfile``, if things like the prisma schema,  ``package.json`` or ``yarn.lock`` changes, it will have to install all the dependencies, which is about 50% of the build time.

Also just noticed, that it probably is a good idea to move stuff like the ``CHANGELOG.md`` to a later point in the ``Dockerfile`` unless it's needed for the build step, as it changes with every release, meaning the docker layer caching won't be at all.

> Tag could be something like linux-arm64-1.770.0

> [Here](https://hub.docker.com/r/portainer/portainer/tags) it looks like you can have ``latest`` tag for different platforms :thinking: 

Yeah, I definitely think the way that portainer does it is a very verbose and user-friendly way, and probably the way to go.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
```
dtslvr, line 33:
```
# platforms: linux/amd64,linux/arm64
platforms: linux/arm64
```

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+          push: true
+          tags: grobut939/ghostfolio:latest
```
dtslvr, line 35: How can I use the current tag instead of `latest`?

dtslvr (changes requested):
I would try it with both _Travis_ and _GitHub Actions_ in a first step and therefore keep `.travis.yml` and `publish-docker-image.sh`.

.github/workflows/docker-image.yml
Areedtz: I'm still not sure it's possible to build one architecture using Travis and another using Github Actions. I'd imagine you'd end up with only one of the architectures being saved under the `latest` tag instead of both.

dtslvr: Yes, I also think it is not possible with _Travis_ and _GitHub Actions_ to have the `latest` tag for both systems. But I guess the `arm64` community would be still happy to have a build at all. And after a few weeks, when the workflow is proven, we can switch both builds to _GitHub Actions_ (including the `latest` tag).

.github/workflows/docker-image.yml
Areedtz: When you want to go away from the ``arm64`` tag, I'd recommend the action to be set up something like [this](https://github.com/ghostfolio/ghostfolio/blob/4f50761be2122b9249f61c09ffaaedcae4b0034a/.github/workflows/docker-image.yml#L23).

This creates docker tags based on the version from the git tag. Both a tag only consisting of major and minor, and a tag consisting of the full version.
The ``latest`` tag will be handled automatically as explained [here](https://github.com/docker/metadata-action#latest-tag).

.github/workflows/docker-image.yml
Areedtz: I've set it up to now only build an image for the ``arm64`` architecture, and it will also only tag it using the ``arm64`` tag.

Areedtz:
> It might actually be possible to speed this up, if we can build make the build step always run as amd64 but build for both platforms like they're talking about in [this blog](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/). We might not even need to do actually run the build step on arm64. I unfortunately don't have time to test that for now.

Alright, I've managed to improve the build times drastically by doing this. The builder step in the ``Dockerfile`` will now run natively always, which means that the most demanding step of the ``Dockerfile`` no longer will be run using emulation.

> I would try it with both Travis and GitHub Actions in a first step and therefore keep ``.travis.yml`` and ``publish-docker-image.sh``.

I've re-added Travis CI and made the ``Build code`` Github Workflow only run when you trigger it manually. When you want to use this instead of Travis, you'll just have to re-add the line shown [here](https://github.com/ghostfolio/ghostfolio/blob/4f50761be2122b9249f61c09ffaaedcae4b0034a/.github/workflows/build-code.yml#L5).

The workflow that builds the ``arm64`` image will run on pull requests to the ``main`` branch and whenever you create a new tag of the ``*.*.*`` format. It will however **only** push the image on a tag, and it only runs on pull requests to make sure that it's possible to build the ``arm64`` image using the proposed changes like seen on [this](https://github.com/Areedtz/ghostfolio/pull/1) pull request.

dtslvr (approved):

PR #1163, opened 2022-08-17
automate database setup and upgrade

fabricemrchl:
Run database:migrate and database:seed at each startup so database will be automatically setup and upgraded. No command needed, just launch the container.

database:migrate on an empty instance create the schema and apply migration. Baseline is not need because database:migrate automatically baseline.
On next start, if no migration script is found nothing is done, if migration script is found databse:migrate will apply it.

database:seed create the base data (live demo account, platform info,...), if database is empty it create data, if database is already existing seed is apply again but I did see any issue with user data.

Please could you confirm that it safe to apply database:seed on existing database with user data ? I no you can remove it from command line. Database migrate will create schema and upgrade it's the most important.

I hope it will help. My main concern is to not run manual command when deploying docker container, like I use to with other app (database setup and upgrade is automatic).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "build:dev": "nx run api:build && nx run client:build && yarn replace-placeholders-in-build",
     "build:storybook": "nx run ui:build-storybook",
     "clean": "rimraf dist",
-    "database:baseline": "sh ./prisma/baseline.sh",
     "database:format-schema": "prisma format",
     "database:generate-typings": "prisma generate",
     "database:gui": "prisma studio",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "replace-placeholders-in-build": "node ./replace.build.js",
     "start": "node dist/apps/api/main",
     "start:client": "ng serve client --configuration=development-en --hmr -o",
-    "start:prod": "node apps/api/main",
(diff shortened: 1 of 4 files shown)


dtslvr:
This looks promising 😃 Thank you for your contribution @fabricemrchl!

`database:seed` is safe and could be applied many times.

package.json
```diff
@@ -42,7 +42,7 @@
     "replace-placeholders-in-build": "node ./replace.build.js",
     "start": "node dist/apps/api/main",
     "start:client": "ng serve client --configuration=development-en --hmr -o",
-    "start:prod": "node apps/api/main",
+    "start:prod": "yarn database:migrate && yarn database:seed && node main",
```
dtslvr, line 45: Could you please update the README.md, some steps can be left out I guess.

dtslvr (changes requested):
Is `database:baseline` still needed?

package.json
fabricemrchl: I don't think `database:baseline` is still needed because `database:migrate` will do it. 

I will  complete this PR by removing it and update the README

dtslvr (approved):
Just tested it and it looks fine. Great improvement @fabricemrchl! 😎

PR #1197, opened 2022-08-27
Feature/finalize GitHub action to create arm64 docker image

dtslvr:


dtslvr:
Hey @Areedtz

I have merged your PR (#1134) and now created the first PR.

I don't understand, what's happening 😃  Why has `Docker image CD / build_and_push (pull_request)` started, there should not be a push (to docker hub). I would expect that only when I add a tag 🤔 

What happened withe the step `Login to DockerHub`? I have not entered the secrets yet.

Check format is working as expected, I have no idea why your change has failed. The duration is acceptable like this, thanks for the optimization!

Areedtz:
Hi @dtslvr

Great! :)

> I don't understand, what's happening 😃 Why has ``Docker image CD / build_and_push (pull_request)`` started, there should not be a push (to docker hub). I would expect that only when I add a tag 🤔

The workflow that builds the arm64 image will run on pull requests to the main branch and whenever you create a new tag of the ``*.*.*`` format. It will however **only** push the image on a tag, and it only runs on pull requests to make sure that it's possible to build the arm64 image using the proposed changes like seen on https://github.com/Areedtz/ghostfolio/pull/1 pull request.

You can see the push condition here:
https://github.com/ghostfolio/ghostfolio/blob/10bae24c5c1955c41cb77e16f30cf633c26549d1/.github/workflows/docker-image.yml#L45

> What happened withe the step ``Login to DockerHub``? I have not entered the secrets yet.

The step is only run on on tags. Since we don't want to push an image on pull requests, we don't need to login to Dockerhub. This check is seen here:
 https://github.com/ghostfolio/ghostfolio/blob/10bae24c5c1955c41cb77e16f30cf633c26549d1/.github/workflows/docker-image.yml#L34

PR #1295, opened 2022-09-26
Gather sector data for mutual funds from Yahoo finance

m11tch:
Having mutual funds in my portfolio, I was a little disappointed by the lack of data in the allocations view. 
I noticed that Yahoo Finance has sector data for the mutual funds, so I figured why not fetch that? 

sadly there is no country/region data, so perhaps a different data source would be preferable, but I couldn't easily find one, so this was the next best thing for now. 

I'm a little rusty, so my implementation is probably a bit dirty, feedback welcome :sweat_smile:


dtslvr:
Thanks for your contribution @m11tch. Please have a look at #1298. I had to add a transformation for the sectors, because it does not match the general style (`Basic Materials` vs. `basic_materials`).

Issue #1264, opened 2022-09-16
Handle ISO 8601 date format


TaakoMagnusen:
When trying to import a csv of transactions i was told the date format is invalid, my CSV had the date written `YYYY-MM-DD` as per [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601).

I see from https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok.csv that the date is expected in `DD/MM/YYYY`.

The importer should accept ISO 8601 as it's the standard for timestamps,  `DD/MM/YYYY` is arbitrary and adheres to no standard.

PR #1303, opened 2022-09-30
parse csv date in ISO format

m11tch:
add support for ISO date format in csv import 
fixes https://github.com/ghostfolio/ghostfolio/issues/1264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added support for translated labels of asset and asset sub class
+- Added support for dates in _ISO 8601_ date format (`YYYY-MM-DD`) in the activities import
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


m11tch:
@dtslvr need any input on this?

dtslvr:
Hey @m11tch

> @dtslvr need any input on this?

My comment is still open: I'm not sure if it's a good idea to try parsing the date three times in a row, but I don't know what is more elegant. Any idea?

m11tch:
> Hey @m11tch
> 
> > @dtslvr need any input on this?
> 
> My comment is still open: I'm not sure if it's a good idea to try parsing the date three times in a row, but I don't know what is more elegant. Any idea?

I don't really have a more elegant solution at this time, an alternative could be to only support dates in ISO format as this seems more universal anyway.

dtslvr:
What do you think of my [proposal](https://github.com/ghostfolio/ghostfolio/pull/1303/commits/1fa0c644483fdc47bdecc0fcb717d75e3a3da178) inspired by https://stackoverflow.com/a/67639094/744012 @m11tch?

m11tch:
> What do you think of my [proposal](https://github.com/ghostfolio/ghostfolio/pull/1303/commits/1fa0c644483fdc47bdecc0fcb717d75e3a3da178) inspired by https://stackoverflow.com/a/67639094/744012 @m11tch?

I guess that's cleaner, maybe we could also add if statements for mm-dd-yyyy  and mm/dd/yyyy? (don't have time to test that myself right now) That should cover most import wishes :D

dtslvr (approved):

dtslvr:
> I guess that's cleaner, maybe we could also add if statements for mm-dd-yyyy and mm/dd/yyyy?

Thanks for the feedback. I think we can add it when it's actually needed.

PR #1319, opened 2022-10-03
Update messages.es.xlf

casitu:
I think I've messed up while translating and I should keep "other" instead of "otras" to show the plural of transaction in spanish. Singular works fine, so I hope my reasoning is correct and translated word will show with this change.


dtslvr (approved):
Yes, you are right. `other` is a reserved keyword.

It reads like `if 1 then "my singular string" else "my plural string"` 🤓

Issue #1103, opened 2022-07-29
Manual dark mode


alfredonodo:
Hi,
I saw that dark mode is already available, but it is based on browser/OS preferences. If the browser uses the anti-fingerprint technique, you cannot use the dark mode. So a simple switch to force the mode would be appropriate.
Thank you

dtslvr:
Add _Appearance_ to _Account Settings_: `AUTO`, `DARK`, `LIGHT`
(very similar to _Base Currency_)

yksolanki9:
@dtslvr I'll like to take this up. Please assign it to me

PR #1342, opened 2022-10-09
Add appearance option in settings

yksolanki9:
Fixes https://github.com/ghostfolio/ghostfolio/issues/1103
- By default, the appearance will be set to `AUTO`, the user can change it in the settings section
- Have added migration file for adding a new column in the Settings table

UI Preview:

https://user-images.githubusercontent.com/39493102/194762749-0ea4a0bc-aa31-40fd-b0cc-5412cecece96.mov


dtslvr (changes requested):
Impressive work @yksolanki9! I have added some feedback as discussed.

dtslvr (approved):
Thanks for your changes! Looks very good 👍🏻

PR #1343, opened 2022-10-09
import localize in first story fixes issue

### LINKED ISSUE #1323: [BUG] Cannot find name '$localize' when running Storybook
**Describe the bug**
_Storybook_ is not starting anymore.

**To Reproduce**
1. Run `yarn start:storybook`
2. Get various errors in files of `libs/ui/src` like: `Cannot find name '$localize'.`

**Expected behavior**
_Storybook_ starts without errors and can be accessed via http://localhost:4400/

**Screenshots**

**Logs**
```
Error: libs/ui/src/lib/activities-table/activities-table.component.ts:134:7 - error TS2304: Cannot find name '$localize'.
$localize`Do you really want to delete this activity?`
at addError (/dev/ghostfolio/source/node_modules/@angular-devkit/build-angular/node_modules/@ngtools/webpack/src/ivy/diagnostics.js:31:29)
```
**Environment**
 - Ghostfolio Version `1.201.0`

LoganArnett:
Closes #1323 

The `preview.js` is not working as expected here for Storybook and is not including the import globally. When attempting to utilize other Storybook versions that are compatible with this version of Angular I was still unsuccessful at getting the `preview.js` to work as Storybook intends it. 

However, importing within a single story makes the localize package available to all stories and allows Storybook to run without error.

### DIFF
--- a/libs/ui/.storybook/preview.js
+++ b/libs/ui/.storybook/preview.js
@@ -1,3 +1,1 @@
-import '@angular/localize/init';
-
 import '!style-loader!css-loader!sass-loader!../../../apps/client/src/styles.scss';
--- a/libs/ui/src/lib/fire-calculator/fire-calculator.component.stories.ts
+++ b/libs/ui/src/lib/fire-calculator/fire-calculator.component.stories.ts
@@ -1,3 +1,4 @@
+import '@angular/localize/init';
 import { CommonModule } from '@angular/common';
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 import { MatButtonModule } from '@angular/material/button';


dtslvr (changes requested):
Thanks for your contribution @LoganArnett. This looks like a reasonable solution 👍🏻 

Can you also add `import '@angular/localize/init';` to `libs/ui/src/lib/activities-table/activities-table.component.ts`? I think it's needed there as well.

LoganArnett:
Looking at that one @dtslvr it looks like it is not needed. This issue appears to be only related to `.stories.ts` files as Storybook isn't picking up the global import. When running the application itself the global for `$localize` is picked up as expected.

dtslvr (approved):
Please ignore my comment. As there is no story for `libs/ui/src/lib/activities-table/activities-table.component.ts`, there is no need to add the import there.

PR #1344, opened 2022-10-10
Feature/add total amount chart to investment timeline

dtslvr:
- [x] Improve total amount calculation
- [x] Handle impersonation
- [ ] Handle first value (should not be 0)


dtslvr:
investment-timeline

PR #1345, opened 2022-10-11
Feature/minor improvements for appearance selector

dtslvr:


dtslvr:
@yksolanki9 I have made some minor improvements (mainly moving code around and localization).

apps/client/src/app/app.component.ts
```diff
@@ -108,7 +108,9 @@ export class AppComponent implements OnDestroy, OnInit {
     this.materialCssVarsService.setDarkTheme(isDarkTheme);
 
     window.matchMedia('(prefers-color-scheme: dark)').addListener((event) => {
-      this.materialCssVarsService.setDarkTheme(event.matches);
+      if (!this.user?.settings.appearance) {
+        this.materialCssVarsService.setDarkTheme(event.matches);
```
dtslvr, line 112: I have found this edge case 😄 It should only override the theme if the user has no appearance setting set.

yksolanki9: Ahh, my bad! Nice catch btw

Issue #1363, opened 2022-10-16
[BUG] Fix placeholder color of disabled mat-select


dtslvr:
![mat-select-disabled-placeholder](https://user-images.githubusercontent.com/4159106/196026556-21f1a28b-79ad-46c6-9f4f-635f7581735d.png)

dtslvr:
A ticket for you @yksolanki9?

yksolanki9:
Oops, I'll fix it. Please assign it to me

PR #1365, opened 2022-10-16
Fix disabled placeholder color

yksolanki9:
Fixes https://github.com/ghostfolio/ghostfolio/issues/1363

UI Preview - Light Mode
Screenshot 2022-10-16 at 8 51 11 PM

UI Preview - Dark Mode
Screenshot 2022-10-16 at 8 51 04 PM

### DIFF
--- a/apps/client/src/styles.scss
+++ b/apps/client/src/styles.scss
@@ -103,11 +103,13 @@ body {
         color: rgba(var(--light-primary-text));
       }
 
-      &.mat-select-disabled .mat-select-placeholder {
-        color: rgba(
-          var(--palette-foreground-disabled-text-dark),
-          var(--palette-foreground-disabled-text-dark-alpha)
-        );
+      &.mat-select-disabled {
+        .mat-select-placeholder {
+          color: rgba(
+            var(--palette-foreground-disabled-text-dark),
+            var(--palette-foreground-disabled-text-dark-alpha)
+          );
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


apps/client/src/styles.scss
```diff
@@ -102,6 +102,13 @@ body {
       .mat-select-placeholder {
```
dtslvr, line 102: You could combine like this:

```
.mat-select-placeholder {
  color: rgba(var(--light-primary-text));
  
  &.mat-select-disabled {
    color: rgba(
      var(--palette-foreground-disabled-text-dark),
      var(--palette-foreground-disabled-text-dark-alpha)
    );
  }
}
```

apps/client/src/styles.scss
```diff
@@ -239,4 +246,11 @@ ngx-skeleton-loader {
   .mat-select-placeholder {
```
dtslvr, line 246: Here as well:

```
.mat-select-placeholder {
  color: rgba(var(--dark-primary-text));
  
  &.mat-select-disabled {
    color: rgba(
      var(--palette-foreground-disabled-text),
      var(--palette-foreground-disabled-text-alpha)
    );
  }
}
```

dtslvr (changes requested):
Looks as expected! 🤩

apps/client/src/styles.scss
dtslvr: Okay, that's fine! 👍🏻

dtslvr (approved):

PR #1403, opened 2022-10-29
New endpoint for Dividends aggregated per month

jpereiramp:
Here's the first draft of the new `/dividends` endpoint powering the _"Dividend Timeline"_ graph under the _"Analysis"_ page. 
I've tried to leverage as much as possible on the existing logic powering the _"Investment Timeline"_ graph of that same page, but please let me know if you'd prefer to have those logics and concerns separated.

I'm still owing (1) some tests and (2) reviewing the `.controller` code, which could be extracted into a private method since it's basically the same logic as the `/investment` endpoint.

### DIFF
--- a/apps/api/src/app/portfolio/portfolio-calculator.ts
+++ b/apps/api/src/app/portfolio/portfolio-calculator.ts
@@ -63,7 +63,7 @@ export class PortfolioCalculator {
     this.orders.sort((a, b) => a.date?.localeCompare(b.date));
   }
 
-  public computeTransactionPoints() {
+  public computeTransactionPoints(types: TypeOfOrder[] = ['BUY', 'SELL']) {
     this.transactionPoints = [];
     const symbols: { [symbol: string]: TransactionPointSymbol } = {};
 
@@ -85,16 +85,20 @@ export class PortfolioCalculator {
         let investment = new Big(0);
 
         if (newQuantity.gt(0)) {
-          if (order.type === 'BUY') {
+          if (order.type === 'BUY' && types.includes('BUY')) {
             investment = oldAccumulatedSymbol.investment.plus(
               order.quantity.mul(unitPrice)
             );
-          } else if (order.type === 'SELL') {
+          } else if (order.type === 'SELL' && types.includes('SELL')) {
             const averagePrice = oldAccumulatedSymbol.investment.div(
               oldAccumulatedSymbol.quantity
             );
             investment = oldAccumulatedSymbol.investment.minus(
               order.quantity.mul(averagePrice)
+            );
+          } else if (order.type === 'DIVIDEND' && types.includes('DIVIDEND')) {
+            investment = oldAccumulatedSymbol.investment.plus(
+              order.quantity.mul(unitPrice)
             );
           }
         }
@@ -492,9 +496,10 @@ export class PortfolioCalculator {
         }
 
         currentDate = parseDate(order.date);
+
         investmentByMonth = order.quantity
-          .mul(order.unitPrice)
-          .mul(this.getFactor(order.type));
+        .mul(order.unitPrice)
+        .mul(this.getFactor(order.type));
       }
 
       if (index === this.orders.length - 1) {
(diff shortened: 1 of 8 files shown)


apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -492,9 +496,10 @@ export class PortfolioCalculator {
         }
 
         currentDate = parseDate(order.date);
+
         investmentByMonth = order.quantity
-          .mul(order.unitPrice)
-          .mul(this.getFactor(order.type));
+        .mul(order.unitPrice)
```
dtslvr, line 501: Be careful, the format check will fail here (`yarn format:check`)

apps/api/src/app/portfolio/portfolio.controller.ts
```diff
@@ -230,6 +231,58 @@ export class PortfolioController {
     return { investments };
   }
 
+  @Get('dividends')
```
dtslvr, line 234: Can you move this method before `getInvestments()`? These are sorted alphabetically.

apps/client/src/app/pages/portfolio/analysis/analysis-page.component.ts
```diff
@@ -34,6 +34,7 @@ export class AnalysisPageComponent implements OnDestroy, OnInit {
   public hasImpersonationId: boolean;
   public investments: InvestmentItem[];
   public investmentsByMonth: InvestmentItem[];
+  public dividendsByMonth: InvestmentItem[];
```
dtslvr, line 37: Move up

apps/client/src/app/pages/portfolio/analysis/analysis-page.html
```diff
@@ -168,4 +168,40 @@ <h3 class="d-flex justify-content-center" i18n>Analysis</h3>
       </div>
     </div>
   </div>
+
```
dtslvr, line 171: Add `mb-5` to the row above (_Investment Timeline_)

apps/client/src/app/pages/portfolio/analysis/analysis-page.html
```diff
@@ -168,4 +168,40 @@ <h3 class="d-flex justify-content-center" i18n>Analysis</h3>
...
+        <gf-investment-chart
+          class="h-100"
+          groupBy="month"
+          [benchmarkDataItems]="dividendsByMonth"
+          [currency]="user?.settings?.baseCurrency"
+          [daysInMarket]="daysInMarket"
+          [isInPercent]="hasImpersonationId || user.settings.isRestrictedView"
+          [locale]="user?.settings?.locale"
+          [range]="user?.settings?.dateRange"
+          [savingsRate]="(hasImpersonationId || user.settings.isRestrictedView) ? undefined : user?.settings?.savingsRate"
```
dtslvr, line 202: You can remove this line, this is only needed for the _Investment Timeline_

apps/client/src/app/services/data.service.ts
```diff
@@ -182,6 +183,19 @@ export class DataService {
     );
   }
 
+  public fetchDividends({
```
dtslvr, line 186: Move up (alphabetical order)

apps/api/src/app/portfolio/portfolio.controller.ts
```diff
@@ -230,6 +231,58 @@ export class PortfolioController {
...
+      throw new HttpException(
+        getReasonPhrase(StatusCodes.FORBIDDEN),
+        StatusCodes.FORBIDDEN
+      );
+    }
+
+    let dividends: InvestmentItem[];
+
+    if (groupBy === 'month') {
+      dividends = await this.portfolioService.getInvestments({
```
dtslvr, line 254: I'm not sure if it's a good idea to mix with investments and all the changes in portfolio calculator.

Another option would be to create a dedicated function `this.portfolioService.getDividends()` which has a query to get all dividends:

```
this.orderService.getOrders({
  filters,
  userCurrency,
  userId,
  types: ['DIVIDEND']
});
```

What do you think?

dtslvr (changes requested):

dtslvr:
Implemented via #1498.

PR #1406, opened 2022-10-30
Add instructions for API Authorization

dtslvr:


dtslvr:
What do you think @DerAndereJohannes?

DerAndereJohannes (approved):
Looks better than what was written before! :-)

DerAndereJohannes (approved):

DerAndereJohannes:
For future: It might make sense to also create something like API.md (or as part of a wiki) which displays all endpoints and examples how to use them and their outputs.. I can assign myself to this if you know what you prefer more between the two

dtslvr:
> For future: It might make sense to also create something like API.md (or as part of a wiki) which displays all endpoints and examples how to use them and their outputs.. I can assign myself to this if you know what you prefer more between the two

Yes, that makes sense for multiple endpoints. Currently, there is only one public endpoint (import). I'd like to keep it in _README.md_ for simplicity at the moment.

DerAndereJohannes:
> Currently, there is only one public endpoint (import).

Simplicity is always best, I agree. Though, all the endpoints that are available to the webapp are also available to the user through http requests. E.g., using a GET on `http://localhost:3333/api/v2/portfolio/performance?range=1d` along with the authorization header gives you the json output that can be further processed.
```
{
  "chart": [
    {
      "date": "2022-10-31",
      "value": <my hidden result>
    },
    {
      "date": "2022-11-01",
      "value": <my hidden result>
    }
  ],
  "errors": [],
  "hasErrors": false,
  "performance": {
    "currentValue": <my hidden result>,
    "currentGrossPerformance": <my hidden result>,
    "currentGrossPerformancePercent": <my hidden result>,
    "currentNetPerformance": <my hidden result>,
    "currentNetPerformancePercent": <my hidden result>
  }
}
```
Are these stable endpoints?

dtslvr:
That is exactly the point. The (public) API for importing activities is considered to remain stable, all other endpoints are subject to change.

DerAndereJohannes:
Alright, thanks for the clarification! :-)

PR #1436, opened 2022-11-09
Feature/convert into pwa

albertsalles4:
Convert into a Progressive Web App. I set up `@angular/pwa` following [this comment](https://github.com/nrwl/nx/discussions/7183#discussioncomment-1409454):
> ```shell
> npm install @angular/pwa --save-dev
> nx g @angular/pwa:ng-add --project *project-name*
> ```

I only had to change a couple of things in `site.webmanifest`:
- The icons src url was prepended with `/en` and I removed it because they could not be found.
- The first icon was tagged as maskable so that it can be displayed safely.
- The `start_url` should be a relative path. In this case we use `/en/` as root because it is _the preferred URL that should be loaded when the user launches the web application_, as explained [here](https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url).

Also, I have added the theme-color meta tag in `index.html`. I copied the value from `site.webmanifest`.

I have tested that the PWA works using Lighthouse:
image

It should be noted that Lighthouse report indicates that it is not _PWA optimized_ because the value of `"background_color": "transparent"` in `site.webmanifest` is not accepted, we should set it to CSS color or HEX value. I can add it if you tell me which color is fine :)

Additionally, I also tested it on a selfhosted server and it works fine.


apps/client/src/assets/site.webmanifest
```diff
@@ -6,19 +6,20 @@
...
+      "src": "/assets/android-chrome-512x512.png",
       "type": "image/png"
     }
   ],
   "name": "Ghostfolio",
   "orientation": "portrait",
   "short_name": "Ghostfolio",
-  "start_url": "https://www.ghostfol.io",
+  "start_url": "/en/",
   "theme_color": "#FFFFFF",
```
dtslvr, line 23: `background_color: "#FFFFFF"` should also be fine (see above)

apps/client/project.json
```diff
@@ -57,6 +58,11 @@
             "glob": "**/*",
             "input": "apps/client/src/assets",
             "output": "./../assets/"
+          },
+          {
+            "glob": "site.webmanifest",
```
dtslvr, line 63: Can you move this below the `sitemap.xml` block?

package.json
```diff
@@ -138,6 +139,7 @@
     "@angular/compiler-cli": "14.2.0",
     "@angular/language-service": "14.2.0",
     "@angular/localize": "14.2.0",
+    "@angular/pwa": "^14.2.9",
```
dtslvr, line 142: Can you fix the version to `14.2.9` please and run `yarn install` again: `"@angular/pwa": "14.2.9"`

dtslvr (changes requested):
Thanks @albertsalles4! Looking forward to try it

albertsalles4:
Thanks for your response @dtslvr! Changes applied 😄. May I suggest adding PWA to the features list in the `README.md` file.

dtslvr (approved):
> Thanks for your response @dtslvr! Changes applied 😄. May I suggest adding PWA to the features list in the `README.md` file.

Thank you for your changes @albertsalles4! I still have reservations because of my bad experience 🙈 I just hope there is no problem with the auto update. But if it works, it can definitely be added to the features: `✅ Progressive Web App (PWA) with a mobile-first design`

PR #1445, opened 2022-11-11
Add PWA as feature

dtslvr:


dtslvr:
What do you think @albertsalles4?

albertsalles4:
That looks great! Thanks for considering my suggestion 😄

PR #1446, opened 2022-11-11
Fix missing comma in json body (API example)

thepill:


dtslvr (approved):
Thank you @thepill! 😃

PR #1472, opened 2022-11-21
Add Postgres connect_timeout for M1 "Can't reach database" error

delenamalan:
I got a "Can't reach database server at `postgres`:`5432`" on my M1 Mac when running Ghostfolio with docker-compose:

```
ghostfolio-development-ghostfolio-1  | [Nest] 330  - 11/21/2022, 7:12:25 PM   ERROR [PrismaService] Error: Can't reach database server at `postgres`:`5432`
ghostfolio-development-ghostfolio-1  |
ghostfolio-development-ghostfolio-1  | Please make sure your database server is running at `postgres`:`5432`.
```

The error is fixed when I add `?connect_timeout=300` to the Postgres connection string like in [this StackOverflow answer](https://stackoverflow.com/a/68495296/3486675).


dtslvr (commented):
Hi @delenamalan

Thanks for your contribution. I've checked your explanation and change.

I would expect the `connect_timeout` parameter in [.env](https://github.com/ghostfolio/ghostfolio/blob/main/.env) because the error will occur during development, right? The `docker-compose` files you have changed are used for production where you should not use your _M1 Mac_. What do you think?

DamianB-BitFlipper:
I ran into the same issue. I run ghostfolio on an Apple M1 laptop. Setting the `connect_timeout` resolved my issue.

dtslvr (approved):

PR #1583, opened 2023-01-08
Feature/French translation

MaVdbussche:
The goal of this pull request is to track progress on the French translation.
About a third of the strings have been translated so far, I will edit this pull request with new commit as I make progress.

### DIFF
--- a/apps/client/src/locales/messages.fr.xlf
+++ b/apps/client/src/locales/messages.fr.xlf
@@ -835,7 +835,7 @@
       </trans-unit>
       <trans-unit id="e698b03c34b459b1b006d7f0473a49b9fcf5dfc1" datatype="html">
         <source>Coupons</source>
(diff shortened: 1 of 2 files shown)


apps/client/src/locales/messages.fr.xlf
```diff
@@ -235,7 +235,7 @@
       </trans-unit>
       <trans-unit id="28f86ffd419b869711aa13f5e5ff54be6d70731c" datatype="html">
         <source>Edit</source>
-        <target state="new">Edit</target>
+        <target state="translated">Modifer</target>
```
dtslvr, line 238: Modif**i**er ❓

dtslvr (changes requested):
That looks very good! I have just added one comment, where I'm not sure: modifer vs. modifier

PR #1595, opened 2023-01-12
Update Node to version 18.x

LeonStoldt:
Hi,
I just tried to update ghostfolio to node v18.x locally and it seems to work fine. I build an image with node 18 and switched my production GUI of ghostfolio with my custom image and had no problem doing so. `yarn test` succeeded aswell.

```shell
docker pull leonstoldt/ghostfolio:development
```

While doing so, I though about switching to node:18-alpine for a smaller image. Or maybe create a second image with `ghostfolio/ghostfolio:x.x.x-alpine` or `ghostfolio/ghostfolio:alpine`. What are your thoughts about this? I could provide a PR for this idea.
**EDIT**: The difference between alpine and slim is not as big as I thought. 153.88 MB for ghostfolio based on slim image and 141.92 MB for ghostfolio based on alpine image. If you like the idea of having an alpine image though, I'll submit a PR. `docker pull leonstoldt/ghostfolio:development-alpine` is already build and was quickly tested in my prod environment.

Hope it helps ;)


dtslvr:
Hi @LeonStoldt

Thanks for your PR.

I have in mind that with Docker Buildx (used to create the `linux/amd64` and `linux/arm64` images), `node alpine` cannot be used (see [here](https://github.com/ghostfolio/ghostfolio/pull/1134#issue-1333291723)).

Otherwise, I don't see anything that speaks against the upgrade to _Node 18_.

LeonStoldt:
Ok, fair point. I tested it on amd64 platform. You suggested something like a `linux-arm64-1.770.0` tag for arm (using slim image). However, if it's just 12MB difference, it's not worth the hassle, I think. I'm fine with keeping it the slim base image.

dtslvr (approved):
Thanks @LeonStoldt!

PR #1596, opened 2023-01-12
Update postgres to version 15 in docker-compose files

LeonStoldt:
While doing my node update, I noticed that the postgres versions in docker-compose are behind. I tested ghostfolio with postgres 15 on my server and thought I provide a PR for this aswell. If it breaks your development environment or you think it's too risky for potential incompability - just ignore it ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Harmonized the logger output: <symbol> (<dataSource>)
 - Improved the language localization for Dutch (`nl`)
+- Ugraded _Postgres_ from version `12` to `15` in the `docker-compose` files
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Harmonized the logger output: <symbol> (<dataSource>)
 - Improved the language localization for Dutch (`nl`)
-- Ugraded _Postgres_ from version `12` to `15` in the `docker-compose` files
+- Upgraded _Postgres_ from version `12` to `15` in the `docker-compose` files
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


dtslvr:
Have you successfully tested the upgrade with an existing volume of postgres 12? Is it fully compatible?

LeonStoldt:
I had not Postgres v12 running - so I could not test compatibility in depth. I am running von v14 at the moment and have testet my current setup with v15.

I quickly created a fresh local ghostfolio instance with postgres v12 and added some random data (buy, sell, item, dividend... some history). Then I swapped the image with v15 and it's not compatible out of the box.

```shell
ghostfolio-development-postgres-1    | 2023-01-21 15:22:01.572 UTC [1] FATAL:  database files are incompatible with server
ghostfolio-development-postgres-1    | 2023-01-21 15:22:01.572 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 15.1 (Debian 15.1-1.pgdg110+1).
```
I guess you would have to dump all db content and restore from dump in the new db like so: https://www.cloudytuts.com/tutorials/docker/how-to-upgrade-postgresql-in-docker-and-kubernetes/.

Probably, this would break some existing installations if they are not very familiar with postgres etc.

dtslvr (approved):
I'm now also running _Postgres_ 15 in the cloud. As the `docker-compose` files are mainly for new self-hosters, I don't see any problem. What do you think?

LeonStoldt:
Hi @dtslvr,
I think it's fine. If someone decides to self-host, some [basic knowledge of Docker](https://github.com/ghostfolio/ghostfolio#run-with-docker-compose) (Compose) should be present. The only issues arise when updating from an older Postgres version to v15, but the log clearly indicates that it's due to an older database already existing.

By the way, I've created an Ansible [Ghostfolio installer script](https://github.com/LeonStoldt/ghostfolio-installer) that already uses Postgres 15 in the Docker Compose file. I made it for some folks in the Unraid Community Forum who were having trouble setting it up. It's a Docker container containing Ansible and sets up the Docker Compose files via variables. If you're interested, I can add a reference to it in the README.

dtslvr:
Thank you for the feedback @LeonStoldt!

Regarding your installer script, you can add the [ghostfolio](https://github.com/topics/ghostfolio) tag and people will start discovering it as a community project.

PR #1598, opened 2023-01-13
Fix group by month/year not working for YTD

### LINKED ISSUE #1574: [BUG] Grouping by month / year is broken for YTD
**Bug Description**

Grouping investments / dividends by month / year has missing data if the date range is `YTD`.

**To Reproduce**

1. Add any `BUY` activity in 2023
2. Go to _Portfolio_ > _Analysis_
3. Select date range `YTD`
4. No data is presented in _Dividend_ / _Investment Timeline_ (the grouping by month / year does not matter)

**Expected behavior**

A bar is displayed in the chart.

**Environment**

- Ghostfolio Version 1.224.0

yksolanki9:
Fixes #1574 
The issue here is that we were comparing the investment date with current date while including with the timestamp.
- For group by month, if investment date is start of month ( let's say `2023-01-01`) and current timestamp is `2023-01-13T16:11:49.108Z`. 
- So, for getting YTD cutoff date, we're setting the date to first day of year which gives us `2023-01-01T16:11:49.108Z`
- And we filter the investments which are not before the YTD cutoff date including the time which will always return false and so we didn't see any investments in YTD.


dtslvr (approved):
I have tested it and it looks much better now 🙂 

Also thanks for the explanation, Yash!

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1429,16 +1429,28 @@ export class PortfolioService {
   private getStartDate(aDateRange: DateRange, portfolioStart: Date) {
     switch (aDateRange) {
       case '1d':
-        portfolioStart = max([portfolioStart, subDays(new Date(), 1)]);
+        portfolioStart = max([
+          portfolioStart,
+          subDays(new Date().setHours(0, 0, 0, 0), 1)
```
dtslvr, line 1434: This function is exactly what you need: https://github.com/ghostfolio/ghostfolio/blob/main/libs/common/src/lib/helper.ts#L201 😄

dtslvr: Never mind, it's not the same:
`2022-12-31T23:00:00.000Z` (`setHours()`) vs. `2023-01-01T00:00:00.000Z` (`resetHours(new Date())`)

dtslvr (approved):

PR #1640, opened 2023-01-29
Feature/change performance calculation to average investment

dtslvr:
Current limitation: only applying to performance of position and not the overall performance.


dtslvr:
Superseded by #2778.

Issue #1680, opened 2023-02-09
problem index.html


fablin:
Hi

When I run :
 yarn start:server

I got this error:
[Nest] 91050  - 02/10/2023, 5:47:43 AM   ERROR [ExceptionsHandler] ENOENT: no such file or directory, stat '/Users/fabrice/workspace/ghostfolio-main/dist/apps/client/index.html'
Error: ENOENT: no such file or directory, stat '/Users/fabrice/workspace/ghostfolio-main/dist/apps/client/index.html'

<img width="1266" alt="image" src="https://user-images.githubusercontent.com/29496388/217947912-bee8f9fe-b0dd-4fe9-bd7e-455a097e4a91.png">

dtslvr:
Hi @fablin

You probably need to run `yarn build:dev` first as mentioned [here](https://github.com/ghostfolio/ghostfolio#setup).

fablin:
> Hi @fablin
> 
> You probably need to run `yarn build:dev` first as mentioned [here](https://github.com/ghostfolio/ghostfolio#setup).

I ran `yarn build:dev` first

fablin:
<img width="1298" alt="Screenshot 2023-02-10 at 22 59 22" src="https://user-images.githubusercontent.com/29496388/218123801-9aef8ac1-236a-4461-8c75-18fea6a8312c.png">

dtslvr:
That looks fine. Are you opening http://localhost:4200/en in your browser?

fablin:
> That looks fine. Are you opening http://localhost:4200/en in your browser?
That's working. 
Thanks for your help.

dtslvr:
> That's working.
> Thanks for your help.

Great. Thank you for your feedback. I have added this instruction to the developer guide: https://github.com/ghostfolio/ghostfolio/pull/1683

PR #1682, opened 2023-02-11
fixed blank index page when self-hosting or developing (Issue #1680)

DerAndereJohannes:
I started checking out the source code more today and walked into the same problem as Issue #1680. It turns out this has been around since the introduction of the heat-map! The landing page assumes that the `countriesOfSubscribers` is always not undefined, so it error-ed out when loading the index page. Looking at the info.service.ts file, we can see that the default value is defined for `countriesOfSubscribers`, but only when the feature is enabled.

This pull request simply wraps the heat-map data entry with an if statement to see if the `countriesOfSubscribers` is defined before adding the data.


dtslvr:
Thank you very much for providing a fix @DerAndereJohannes!

I have slightly adapted your suggested solution by setting a default value for `countriesOfSubscribers`.

dtslvr (approved):

Issue #1576, opened 2023-01-06
Import CVS = request entity too large


dakkusingh:
I am hosting using Docker and I get an error when Import of 700 items fails with 
{"statusCode":413,"message":"request entity too large"}

dakkusingh:
```
2023-01-06 19:24:30 [Nest] 277  - 01/06/2023, 7:24:30 PM   ERROR [ExceptionsHandler] request entity too large
2023-01-06 19:24:30 PayloadTooLargeError: request entity too large
2023-01-06 19:24:30     at readStream (/ghostfolio/apps/api/node_modules/raw-body/index.js:156:17)
2023-01-06 19:24:30     at getRawBody (/ghostfolio/apps/api/node_modules/raw-body/index.js:109:12)
2023-01-06 19:24:30     at read (/ghostfolio/apps/api/node_modules/body-parser/lib/read.js:79:3)
2023-01-06 19:24:30     at jsonParser (/ghostfolio/apps/api/node_modules/body-parser/lib/types/json.js:135:5)
2023-01-06 19:24:30     at Layer.handle [as handle_request] (/ghostfolio/apps/api/node_modules/express/lib/router/layer.js:95:5)
2023-01-06 19:24:30     at trim_prefix (/ghostfolio/apps/api/node_modules/express/lib/router/index.js:328:13)
2023-01-06 19:24:30     at /ghostfolio/apps/api/node_modules/express/lib/router/index.js:286:9
2023-01-06 19:24:30     at Function.process_params (/ghostfolio/apps/api/node_modules/express/lib/router/index.js:346:12)
2023-01-06 19:24:30     at next (/ghostfolio/apps/api/node_modules/express/lib/router/index.js:280:10)
2023-01-06 19:24:30     at cors (/ghostfolio/apps/api/node_modules/cors/lib/index.js:188:7)
```

dtslvr:
Hi @dakkusingh

How big is the file you want to upload (in MB)?

dakkusingh:
I think it was around 200K in CSV format and around 1.2MB in JSON.

dtslvr:
Reference: https://stackoverflow.com/a/59978098/744012

PR #1726, opened 2023-02-18
 Increase file size limit for imports

jason-feng:
## Summary

This PR is to resolve the following Issue: https://github.com/ghostfolio/ghostfolio/issues/1576.
I added test/import/ok-large-activities.json. This has 500 lines, on master, the above issue reproduces. 
On this new branch the issue is fixed. There is a separate issue where the activities generated take a very long time to import after the file is validated.


dtslvr:
Thank you very much for your PR @jason-feng. Great that you have added a test file 😃 

I have removed `urlencoded` for now and added the `body-parser` explicitly to the `package.json`.

Validation and import indeed takes a long time, but that's a different topic.

dtslvr (approved):

PR #1748, opened 2023-02-25
Feature/retirement date

robbert36:
An interesting metric for the fire dashboard would be the retirement date. This date can be obtained by determining a target net worth based on the 4% rule which will allow you to retire.
- Renamed investmentHorizon to retirementDate and used a date picker so the specific month can be chosen
- Added projectedTotalAmount as input field so a user can enter his required net worth for retirement.
- One of the input fields should be filled in, the other one will be calculated automatically
- The field that is modified the most recent is stored in the user settings, so it is persisted until next login/refresh. 

ghostfolio-retirementDate

### DIFF
--- a/apps/api/src/app/user/update-user-setting.dto.ts
+++ b/apps/api/src/app/user/update-user-setting.dto.ts
@@ -49,6 +49,10 @@ export class UpdateUserSettingDto {
   @IsOptional()
   locale?: string;
 
+  @IsNumber()
+  @IsOptional()
+  projectedTotalAmount?: number;
+
   @IsDate()
   @IsOptional()
   retirementDate?: Date;
@@ -56,10 +60,6 @@ export class UpdateUserSettingDto {
   @IsNumber()
   @IsOptional()
   savingsRate?: number;
-
-  @IsNumber()
-  @IsOptional()
-  projectedTotalAmount?: number;
 
   @IsIn(<ViewMode[]>['DEFAULT', 'ZEN'])
   @IsOptional()
--- a/apps/api/src/app/user/update-user-setting.dto.ts
+++ b/apps/api/src/app/user/update-user-setting.dto.ts
@@ -48,13 +48,21 @@ export class UpdateUserSettingDto {
   @IsOptional()
   locale?: string;
 
+  @IsString()
+  @IsOptional()
+  retirementDate?: string;
+
   @IsNumber()
   @IsOptional()
   savingsRate?: number;
 
   @IsNumber()
   @IsOptional()
-  targetNetWorth?: number;
+  projectedTotalAmount?: number;
+
+  @IsBoolean()
+  @IsOptional()
+  projectedTotalAmountSet?: boolean;
 
   @IsIn(<ViewMode[]>['DEFAULT', 'ZEN'])
   @IsOptional()
--- a/apps/api/src/app/user/update-user-setting.dto.ts
+++ b/apps/api/src/app/user/update-user-setting.dto.ts
@@ -3,10 +3,11 @@ import type {
   DateRange,
   ViewMode
 } from '@ghostfolio/common/types';
+import { Type } from 'class-transformer';
 import {
   IsBoolean,
-  IsDate,
   IsIn,
+  IsISO8601,
   IsNumber,
   IsOptional,
   IsString
@@ -53,9 +54,9 @@ export class UpdateUserSettingDto {
   @IsOptional()
   projectedTotalAmount?: number;
 
-  @IsDate()
+  @IsISO8601()
   @IsOptional()
-  retirementDate?: Date;
+  retirementDate?: string;
 
   @IsNumber()
   @IsOptional()
--- a/apps/api/src/app/user/update-user-setting.dto.ts
+++ b/apps/api/src/app/user/update-user-setting.dto.ts
@@ -5,6 +5,7 @@ import type {
 } from '@ghostfolio/common/types';
 import {
   IsBoolean,
+  IsDate,
   IsIn,
   IsNumber,
   IsOptional,
@@ -48,9 +49,9 @@ export class UpdateUserSettingDto {
   @IsOptional()
   locale?: string;
 
-  @IsString()
+  @IsDate()
   @IsOptional()
-  retirementDate?: string;
+  retirementDate?: Date;
 
   @IsNumber()
   @IsOptional()
@@ -59,10 +60,6 @@ export class UpdateUserSettingDto {
   @IsNumber()
   @IsOptional()
   projectedTotalAmount?: number;
-
-  @IsBoolean()
-  @IsOptional()
-  projectedTotalAmountSet?: boolean;
 
   @IsIn(<ViewMode[]>['DEFAULT', 'ZEN'])
   @IsOptional()
--- a/apps/api/src/app/user/update-user-setting.dto.ts
+++ b/apps/api/src/app/user/update-user-setting.dto.ts
@@ -52,6 +52,10 @@ export class UpdateUserSettingDto {
   @IsOptional()
   savingsRate?: number;
 
+  @IsNumber()
+  @IsOptional()
+  targetNetWorth?: number;
+
   @IsIn(<ViewMode[]>['DEFAULT', 'ZEN'])
   @IsOptional()
   viewMode?: ViewMode;
--- a/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
@@ -39,7 +39,6 @@ export class FirePageComponent implements OnDestroy, OnInit {
   public ngOnInit() {
     this.isLoading = true;
     this.deviceType = this.deviceService.getDeviceInfo().deviceType;
-
     this.dataService
       .fetchPortfolioDetails({})
       .pipe(takeUntil(this.unsubscribeSubject))
@@ -92,8 +91,29 @@ export class FirePageComponent implements OnDestroy, OnInit {
   }
 
   public onSavingsRateChange(savingsRate: number) {
+    console.log(this.user.settings);
     this.dataService
       .putUserSetting({ savingsRate })
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe(() => {
+        this.userService.remove();
+
+        this.userService
+          .get()
+          .pipe(takeUntil(this.unsubscribeSubject))
+          .subscribe((user) => {
+            this.user = user;
+
+            this.changeDetectorRef.markForCheck();
+          });
+      });
+  }
+
+  public onTargetNetWorthChanged(targetNetWorth: number) {
+    console.log('onTargetNetWorthChanged: ' + targetNetWorth);
+    console.log(this.user.settings);
+    this.dataService
+      .putUserSetting({ targetNetWorth })
       .pipe(takeUntil(this.unsubscribeSubject))
       .subscribe(() => {
         this.userService.remove();
--- a/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
@@ -91,9 +91,12 @@ export class FirePageComponent implements OnDestroy, OnInit {
       });
   }
 
-  public onSavingsRateChange(savingsRate: number) {
+  public onRetirementDateChange(retirementDate: Date) {
     this.dataService
-      .putUserSetting({ savingsRate })
+      .putUserSetting({
+        retirementDate: retirementDate.toISOString(),
+        projectedTotalAmount: null
+      })
       .pipe(takeUntil(this.unsubscribeSubject))
       .subscribe(() => {
         this.userService.remove();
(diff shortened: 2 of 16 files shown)


dtslvr:
Hi @robbert36

First of all, I think this is a great idea! 👍🏻

From a usability perspective, I still find it a bit confusing. Shouldn't the user either choose _Investment Horizon_ or _Target Net Worth_? Both at the same time makes no sense to me. My idea would be to place these two fields at least together by moving _Investment Horizon_ just before _Target Net Worth_.

If _Target Net Worth_ has been defined, the chart should consider this and render all years until 2046. _Projected Total Amount_ can be hidden in that case.

What do you think?

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -39,7 +39,6 @@ export class FirePageComponent implements OnDestroy, OnInit {
   public ngOnInit() {
     this.isLoading = true;
     this.deviceType = this.deviceService.getDeviceInfo().deviceType;
-
```
dtslvr, line 42: Please add the line break again 🙏🏻

libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
```diff
@@ -50,20 +51,24 @@ export class FireCalculatorComponent
...
 
   @ViewChild('chartCanvas') chartCanvas;
 
   public calculatorForm = this.formBuilder.group({
     annualInterestRate: new FormControl<number>(undefined),
     paymentPerPeriod: new FormControl<number>(undefined),
     principalInvestmentAmount: new FormControl<number>(undefined),
-    time: new FormControl<number>(undefined)
+    time: new FormControl<number>(undefined),
+    retirementNetWorth: new FormControl<number>(undefined)
```
dtslvr, line 66: Please move up (for an alphabetical order)

libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
```diff
@@ -86,7 +91,8 @@ export class FireCalculatorComponent
         annualInterestRate: 5,
         paymentPerPeriod: this.savingsRate,
         principalInvestmentAmount: 0,
-        time: 10
+        time: 10,
+        retirementNetWorth: this.targetNetWorth
```
dtslvr, line 95: Please move up (for an alphabetical order)

libs/ui/src/lib/fire-calculator/fire-calculator.service.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [FireCalculatorService]
+    }).compile();
+
+    fireCalculatorService = module.get<FireCalculatorService>(
+      FireCalculatorService
+    );
+  });
+
+  it('true', async () => {
```
dtslvr, line 19: ❓

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
```diff
@@ -40,4 +40,36 @@ export class FireCalculatorService {
...
+    r,
+    totalAmount
+  }: {
+    P: number;
+    PMT: number;
+    r: number;
+    totalAmount: number;
+  }) {
+    if (r == 0) {
+      //No compound interest
```
dtslvr, line 56: Please add a space after `//`

dtslvr (changes requested):
Please revert the changes in `yarn.lock`

robbert36:
> Hi @robbert36
> 
> First of all, I think this is a great idea! 👍🏻
> 
> From a usability perspective, I still find it a bit confusing. Shouldn't the user either choose _Investment Horizon_ or _Target Net Worth_? Both at the same time makes no sense to me. My idea would be to place these two fields at least together by moving _Investment Horizon_ just before _Target Net Worth_.
> 
> If _Target Net Worth_ has been defined, the chart should consider this and render all years until 2046. _Projected Total Amount_ can be hidden in that case.
> 
> What do you think?

Thanks, I'm using the application and thought it would be a nice addition.
Absolutely, sounds like a good idea. It makes sense to change the chart depending on the target net worth input.
The only thing I'm still wondering is how we can have a good UX making sure only one of the fields (investment horizon, target net worth) is being used. Having the 2 input fields below each other would still allow a user to fill in both which can be confusing. Maybe a toggle would be a good option to choose between the two?

I'll take a further look at these changes.

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
robbert36: Done

libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
robbert36: Done

libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
robbert36: Done

libs/ui/src/lib/fire-calculator/fire-calculator.service.spec.ts
robbert36: Removed the dummy test

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
robbert36: Done

robbert36:
Created a new commit. The chart will now show the years until the retirement date. If no target net worth is set, the investment horizon is used as it was before.
Screenshots are updated in the description, let me know what you think.

apps/api/src/app/user/update-user-setting.dto.ts
```diff
@@ -52,6 +52,10 @@ export class UpdateUserSettingDto {
   @IsOptional()
   savingsRate?: number;
 
+  @IsNumber()
+  @IsOptional()
+  targetNetWorth?: number;
```
dtslvr, line 57: I would rename to `projectedTotalAmount` (or do you know a better name? e.g. `retirementAmount`).

Reason: In my case it is not the net worth as I have an emergency fund (it is not included here, because the annual interest rate of the emergency fund is much lower).

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -109,6 +109,24 @@ export class FirePageComponent implements OnDestroy, OnInit {
       });
   }
 
+  public onTargetNetWorthChanged(targetNetWorth: number) {
```
dtslvr, line 112: Rename accordingly

apps/client/src/locales/messages.de.xlf
```diff
@@ -3481,6 +3481,22 @@
           <context context-type="linenumber">2</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="46d3a0f17b741c93f9e61aa7157820da41506f53" datatype="html">
+        <source>Target Net Worth</source>
+        <target state="new">Angestrebtes Nettovermögen</target>
```
dtslvr, line 3486: 😂 

You should also mark `state="translated"`

libs/common/src/lib/interfaces/user-settings.interface.ts
```diff
@@ -11,5 +11,6 @@ export interface UserSettings {
   language?: string;
   locale?: string;
   savingsRate?: number;
+  targetNetWorth?: number;
```
dtslvr, line 14: Rename accordingly

libs/ui/src/lib/fire-calculator/fire-calculator.component.html
```diff
@@ -17,24 +17,36 @@
...
             matInput
-            step="0.25"
+            step="100"
             type="number"
           />
-          <div class="ml-2" matTextSuffix>%</div>
+          <span class="ml-2" matTextSuffix>{{ currency }}</span>
         </mat-form-field>
 
         <gf-value
```
dtslvr, line 48: Always show the input (not as `gf-value`) that you can directly edit (either _Investment Horizon_ or _Projected Total Amount_)

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
```diff
@@ -40,4 +40,36 @@ export class FireCalculatorService {
...
+  }) {
+    if (r == 0) {
+      // No compound interest
+      return (totalAmount - P) / PMT;
+    } else if (totalAmount <= 0) {
+      return 0;
+    }
+
+    const periodInterest = new Big(r).div(this.COMPOUND_PERIOD);
+    const numerator1: number = Math.log10(
```
dtslvr, line 63: Do you need to transform to `number` here?

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
```diff
@@ -40,4 +40,36 @@ export class FireCalculatorService {
...
+      new Big(totalAmount).plus(new Big(PMT).div(periodInterest)).toNumber()
+    );
+    const numerator2: number = Math.log10(
+      new Big(P).plus(new Big(PMT).div(periodInterest)).toNumber()
+    );
+    const denominator: number = Math.log10(
+      new Big(1).plus(periodInterest).toNumber()
+    );
+
+    return (numerator1 - numerator2) / denominator;
```
dtslvr, line 73: What about
```
return numerator1.minus(numerator2).div(denominator).toNumber();
```

dtslvr (changes requested):
Looks better. I have an idea based on this:

Why not show _Projected Total Amount_ as an input field?

When you edit _Investment Horizon_, it will update _Projected Total Amount_ and vice versa.

dtslvr:
@robbert36 That sounds good! Instead of a date picker, maybe two selectors `[Month ▼] [Year ▼]`. I think a date is too specific 😄

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
robbert36: I didn't find a good library to take logarithms of big numbers. Probably it is not feasible to do for big numbers. That's why I needed to cast to a number here. For the numbers this is used for in our context, this should be fine though.

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
robbert36: See above, logarithms of Bignumbers don't seem feasible.

libs/ui/src/lib/fire-calculator/fire-calculator.component.html
robbert36: Changed Projected Total amount to an input field.

libs/common/src/lib/interfaces/user-settings.interface.ts
robbert36: used projectedTotalAmount in the new commit

apps/client/src/locales/messages.de.xlf
robbert36: updated :-)

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
robbert36: Changed to projectedTotalAmount in latest commit

apps/api/src/app/user/update-user-setting.dto.ts
robbert36: Done, used projectedTotalAmount in latest commit

robbert36:
Created a new commit where both retirementDate and projectedTotalAmount are input fields.
Both fields can be modified and are stored in the user settings so this is persisted. An extra boolean field is used to determine which field is set by the user, the other field is calculated automatically based on the user's portfolio.
See gif in description for the latest changes int he frontend.

libs/ui/src/lib/fire-calculator/fire-calculator.service.ts
dtslvr: Thanks for the clarification

dtslvr:
Thanks a lot for the update @robbert36. Especially the new gif 😎 

I have refactored the following on your branch:
* Eliminated `@angular/material-moment-adapter`
* Eliminated `projectedTotalAmountSet`
* Refactored `retirementDate` to type `Date`

The date picker currently shows a full date, which could be improved later. But I don't want to use `moment` as a dependency.

Can you try it?

apps/api/src/app/user/update-user-setting.dto.ts
```diff
@@ -48,6 +49,14 @@ export class UpdateUserSettingDto {
   @IsOptional()
   locale?: string;
 
+  @IsNumber()
+  @IsOptional()
+  projectedTotalAmount?: number;
+
+  @IsDate()
```
robbert36, line 56:
```suggestion
  @Type(() => Date)
  @IsDate()
```

robbert36 (commented):
Thanks for the refactor.
One thing is not working yet, when changing the retirementDate, the userSettings are pushed to the backend with a PUT call. This fails because the retirementDate string is not recognised as a date in the HTTP body.
Adding the extra annotation I suggested fixes this.

Other than that it looks great to me. So if the above issue is fixed, this can go in for me

apps/api/src/app/user/update-user-setting.dto.ts
dtslvr: Thanks a lot for the hint. I was not able to solve it with `@Type(() => Date)` though. But it's working now the same way as `date` in `CreateOrderDto`.

PR #1824, opened 2023-04-04
Feature/improve activities import for csv files of ibkr

dtslvr:


longstone:
Wow - this means we can import directly from lbkr? Is there documentation?

PR #1880, opened 2023-04-20
feature: allow to delete all activities from portfolio

x1c0:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Introduced a new button to delete all activities from the portfolio activities page
+
 ## 1.260.0 - 2023-04-23
 
 ### Added
@@ -32,7 +38,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the alignment of the performance column header in the holdings table
 - Removed the unnecessary sort header of the comment column in the activities table
 - Fixed the targets in `proxy.conf.json` from `http://localhost:3333` to `http://0.0.0.0:3333` for local development
-- Introduced a new button to delete all activities from the portfolio activities page
 
 ## 1.258.0 - 2023-04-20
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed the targets in `proxy.conf.json` from `http://localhost:3333` to `http://0.0.0.0:3333` for local development
+- Introduced a new button to delete all activities from the portfolio activities page
```
dtslvr, line 13: Please move to the section `### Added`

apps/api/src/app/order/order.controller.ts
```diff
@@ -62,6 +62,23 @@ export class OrderController {
     });
   }
 
+  @Delete()
```
dtslvr, line 65: Can you move this function up?

First `@Delete()`, then `@Delete(':id')`

apps/client/src/app/services/data.service.ts
```diff
@@ -150,6 +150,10 @@ export class DataService {
     return this.http.delete<any>(`/api/v1/order/${aId}`);
   }
 
+  public deleteAllOrders() {
```
dtslvr, line 153: Can you move this function up so the order is alphabetical?

libs/ui/src/lib/activities-table/activities-table.component.html
```diff
@@ -53,6 +53,14 @@
       <ion-icon class="mr-2" name="calendar-clear-outline"></ion-icon>
       <span i18n>Export Drafts as ICS</span>
     </button>
+    <button
+      class="align-items-center d-flex"
+      mat-menu-item
+      (click)="onDeleteAllActivities()"
+    >
+      <ion-icon class="mr-2" name="trash-outline"></ion-icon>
+      <span i18n>Delete all activities</span>
```
dtslvr, line 62: Delete all **A**ctivities

libs/ui/src/lib/activities-table/activities-table.component.ts
```diff
@@ -52,6 +52,7 @@ export class ActivitiesTableComponent implements OnChanges, OnDestroy, OnInit {
   @Output() export = new EventEmitter<string[]>();
   @Output() exportDrafts = new EventEmitter<string[]>();
   @Output() import = new EventEmitter<void>();
+  @Output() deleteAllActivities = new EventEmitter<void>();
```
dtslvr, line 55: Can you move up: before`@Output() export...`

dtslvr (changes requested):

apps/client/src/app/services/data.service.ts
x1c0: there is some kind of plugin that does this automatically :) maybe worth investigating

libs/ui/src/lib/activities-table/activities-table.component.html
x1c0: actually in English we don't capitalize names like in German :) but i'll change it :P

dtslvr (approved):
Thank you for the modifications. Looks fine! 👍🏻

PR #1882, opened 2023-04-20
Improved holdings table showing cash position also when the filter contains accounts

x1c0:
We have 2 accounts, Account A (Mine) has 2 stocks, Account B (Familly) 1 stock:
image

Then if I filter by Account A, I want to see my cash position and how much is allocated on that account:
image

The same for Account B:
image

And if we select both, should give is the same initial result when we had no filters, since we only have 2 accounts:
image

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -462,10 +462,17 @@ export class PortfolioService {
     });
 
     const holdings: PortfolioDetails['holdings'] = {};
-    const totalInvestmentInBaseCurrency = currentPositions.totalInvestment.plus(
+    const totalInvestmentInBaseCurrency = currentPositions.currentValue.plus(
       cashDetails.balanceInBaseCurrency
     );
-    let filteredValueInBaseCurrency = currentPositions.currentValue;
+
+    const isFilteredByAccount = filters.some(
+      (filter) => filter.type === 'ACCOUNT'
+    );
+
+    let filteredValueInBaseCurrency = isFilteredByAccount
+      ? totalInvestmentInBaseCurrency
+      : currentPositions.currentValue;
 
     if (
       filters?.length === 0 ||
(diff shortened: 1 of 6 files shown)


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -462,10 +462,17 @@ export class PortfolioService {
     });
 
     const holdings: PortfolioDetails['holdings'] = {};
-    const totalInvestmentInBaseCurrency = currentPositions.totalInvestment.plus(
+    const totalInvestmentInBaseCurrency = currentPositions.currentValue.plus(
```
dtslvr, line 465: `const totalValueInBaseCurrency = ...`

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -462,10 +462,17 @@ export class PortfolioService {
...
       cashDetails.balanceInBaseCurrency
     );
-    let filteredValueInBaseCurrency = currentPositions.currentValue;
+
+    const isFilteredByAccount = filters.some(
```
dtslvr, line 469: 👍🏻

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -462,10 +462,17 @@ export class PortfolioService {
...
+      (filter) => filter.type === 'ACCOUNT'
```
dtslvr, line 470: I'm always using the written out version. Can you adapt for harmonization:

`(filter) => { return filter.type === 'ACCOUNT'; }`

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -564,12 +571,11 @@ export class PortfolioService {
       };
     }
 
-    if (
-      filters?.length === 0 ||
-      (filters?.length === 1 &&
-        filters[0].type === 'ASSET_CLASS' &&
-        filters[0].id === 'CASH')
-    ) {
+    const isFilteredByCash = filters.some(
```
dtslvr, line 574: 👍🏻

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -564,12 +571,11 @@ export class PortfolioService {
...
+      (filter) => filter.type === 'ASSET_CLASS' && filter.id === 'CASH'
```
dtslvr, line 575: Can you use the written out form here too?

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -595,7 +601,7 @@ export class PortfolioService {
       filters[0].id === EMERGENCY_FUND_TAG_ID &&
       filters[0].type === 'TAG'
     ) {
-      const cashPositions = await this.getCashPositions({
+      const emergencyFundCashPositions = await this.getCashPositions({
```
dtslvr, line 604: 👍🏻

dtslvr (changes requested):
Cool 😎 Can please you rebase the branch afterwards?

dtslvr:
Superseded by #1897.

PR #1897, opened 2023-04-25
Fix/holdings always include cash position

x1c0:
not sure what happened with the previous PR.. this one is correct

PR #1908, opened 2023-04-30
Fix accounts page

dtslvr:


dtslvr:
The improvement with `CASH` in holdings (https://github.com/ghostfolio/ghostfolio/pull/1897) has broken the accounts page. I have added the missing guards, but now all the activities are missing there. Any idea @x1c0?

dtslvr:
> I have added the missing guards, but now all the activities are missing there. Any idea @x1c0?

Never mind, it should work with this fix.

Issue #1932, opened 2023-05-06
Improve check for duplicates in import


dtslvr:
If a user exports activities from a broker and imports the complete file multiple files, Ghostfolio should take care that the portfolio does not contain duplicate entries of the same activities.

To reproduce, successfully import any [test file](https://github.com/ghostfolio/ghostfolio/tree/main/test/import). Then import the same file again and the user should see this:

<img width="1188" alt="duplicate-activities-import-png" src="https://user-images.githubusercontent.com/4159106/236608180-e0ed5ae0-f053-45af-a3e9-de618d726758.png">

visrut7:
You can assign it to me I will work on this.

PR #1940, opened 2023-05-06
Feature/improve check for duplicates in import #1932

visrut7:


apps/api/src/app/import/import.service.ts
```diff
@@ -94,7 +97,8 @@ export class ImportService {
             value,
             assetProfile.currency,
             userCurrency
-          )
+          ),
+          isDuplicate: false
```
dtslvr, line 101: Can you please move it below `isDraft` to keep the alphabetical order?

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
           value,
           currency,
           userCurrency
-        )
+        ),
+        isDuplicate
```
dtslvr, line 340: Can you please move it below`...order,` to keep the alphabetical order?

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
...
       });
     }
-
+    console.log(activities);
```
dtslvr, line 343: Remove

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
...
     return activities;
   }
 
+  private async labelDuplicateActivities({
```
dtslvr, line 347: Rename to `markActivitiesAsDuplicates`?

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
...
+        dataSource,
+        date,
+        fee,
+        quantity,
+        symbol,
+        type,
+        unitPrice
+      } = activitiesDtoEntry;
+
+      const duplicateActivity = existingActivities.find((activity) => {
```
dtslvr, line 374: Rename to `isDuplicate`

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
...
+          isSameDay(activity.date, parseISO(<string>(<unknown>date))) &&
+          activity.fee === fee &&
+          activity.quantity === quantity &&
+          activity.SymbolProfile.symbol === symbol &&
+          activity.type === type &&
+          activity.unitPrice === unitPrice
+        );
+      });
+
+      if (duplicateActivity) {
```
dtslvr, line 387: Simplify to:

```
activitiesDtoWithDuplication.push({
  ...activitiesDtoEntry,
  isDuplicate
});
```

apps/api/src/app/order/create-order.dto.ts
```diff
@@ -8,13 +8,15 @@ import {
...
   IsArray,
+  IsBoolean,
   IsEnum,
   IsISO8601,
   IsNumber,
   IsOptional,
   IsString
 } from 'class-validator';
 import { isString } from 'lodash';
+import { CreateAccessDto } from '../access/create-access.dto';
```
dtslvr, line 19: Remove

apps/api/src/app/order/create-order.dto.ts
```diff
@@ -65,3 +67,56 @@ export class CreateOrderDto {
   @IsNumber()
   unitPrice: number;
 }
+
+export class OrderDto {
```
dtslvr, line 71: Can you create a new file: `order.dto.ts`

apps/api/src/app/order/create-order.dto.ts
```diff
@@ -65,3 +67,56 @@ export class CreateOrderDto {
...
+  tags?: Tag[];
+
+  @IsEnum(Type, { each: true })
+  type: Type;
+
+  @IsNumber()
+  unitPrice: number;
+
+  @IsBoolean()
+  isDuplicate: boolean;
```
dtslvr, line 121: Move below `fee` to keep the alphabetical order

apps/api/src/app/order/interfaces/activities.interface.ts
```diff
@@ -8,4 +8,5 @@ export interface Activity extends OrderWithAccount {
   feeInBaseCurrency: number;
   value: number;
   valueInBaseCurrency: number;
+  isDuplicate: boolean;
```
dtslvr, line 11: Move below `feeInBaseCurrency` to keep the alphabetical order

libs/ui/src/lib/activities-table/activities-table.component.html
```diff
@@ -85,12 +85,22 @@
       </th>
       <td *matCellDef="let element" class="px-1" mat-cell>
         <mat-checkbox
+          *ngIf="!element.isDuplicate"
```
dtslvr, line 88: Please only use one `mat-checkbox` and add the attribute:

```
[disabled]="element.isDuplicate"
```

libs/ui/src/lib/activities-table/activities-table.component.html
```diff
@@ -85,12 +85,22 @@
...
           class="mt-2"
```
dtslvr, line 89: Can you remove this line, I think it's not needed anymore

libs/ui/src/lib/activities-table/activities-table.component.html
```diff
@@ -85,12 +85,22 @@
...
           color="primary"
           [checked]="selectedRows.isSelected(element)"
```
dtslvr, line 91: Probably like this:

```
[checked]="element.isDuplicate ? false : selectedRows.isSelected(element)"
```

apps/api/src/app/import/import.service.ts
```diff
@@ -228,8 +237,9 @@ export class ImportService {
       quantity,
       symbol,
       type,
-      unitPrice
-    } of activitiesDto) {
+      unitPrice,
+      isDuplicate
```
dtslvr, line 241: Can you move up to keep the alphabetical order?

apps/api/src/app/import/import.service.ts
```diff
@@ -322,13 +336,70 @@ export class ImportService {
...
+    activitiesDto: Partial<CreateOrderDto>[];
+    userId: string;
+  }): Promise<Partial<OrderDto>[]> {
+    const existingActivities = await this.orderService.orders({
+      include: { SymbolProfile: true },
+      orderBy: { date: 'desc' },
+      where: { userId }
+    });
+
+    const activitiesDtoWithDuplication: Partial<OrderDto>[] = [];
```
dtslvr, line 360: Can you use `Partial<Activity>[]`? Then you could get rid of `OrderDto` completely

dtslvr (changes requested):
That looks promising @visrut7! I have added some comments, just ask if something is unclear.

apps/api/src/app/import/import.service.ts
dtslvr: Can you write: `isDuplicate: false // TODO: Use evaluated state`

In a follow up version, it should work the same way for dividend :)

apps/api/src/app/import/import.service.ts
visrut7: I have used `Partial<Activity>[]` but I think we will get a type error in the loop, because Activity doesn't contain `currency`, and some other fields maybe? Or Am I missing something?

image

I can clean up `OrderDto` like this if you prefer

image

visrut7: I can clean up data definition like this

image

dtslvr: You are right. Extending is fine. I would remove `@IsBoolean()` as we do not validate here.

PR #1963, opened 2023-05-12
feature/drip transaction

raskanskyz:
- chore: create prisma migration
- chore: add "DRIP" selection to activities page + table

### DIFF
--- a/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
+++ b/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
@@ -15,6 +15,7 @@
           <mat-option i18n value="DIVIDEND">Dividend</mat-option>
           <mat-option i18n value="ITEM">Item</mat-option>
           <mat-option i18n value="SELL">Sell</mat-option>
+          <mat-option i18n value="DRIP">DRIP</mat-option>
         </mat-select>
       </mat-form-field>
     </div>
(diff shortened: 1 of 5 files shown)


raskanskyz:
@dtslvr can you please review this?
I'm sure I missed a few things since i'm not too familiar with the project, yet :)

The only thing I tested was adding a "DRIP" transactoin both manually and by JSON import and I see that the transaction is added as expected and it also appears in the "Dividend Timeline".
What else needs to be verified?

dtslvr:
Hi @raskanskyz

Thanks for your suggestion. I have finally found some time to review your PR.

I would like to understand exactly why a new `DRIP` type is needed. Can't the same scenario already be represented with a `DIVIDEND` and a `BUY` (with `unitPrice` 0) activity?

raskanskyz:
Hey @dtslvr,
Thanks for the review/reply!
You are correct although I think the `unitPrice ` should be the price in which the stock was repurchased rather than `0`.
This PR might actually be unnecessary. I was mainly looking for a convinient way to represent "DRIP" since it's a common term/type in most brokerages.
What I can do alternatively is update the JSON I want to import and split the "DRIP" transactions into `DIVIDEND` and `BUY` transactions.

dtslvr:
Maybe you can create a test scenario @raskanskyz [here](https://github.com/ghostfolio/ghostfolio/tree/main/test/import) to verify?

Something like `ok-drip.json` including `BUY` (initial buy), `DIVIDEND`, `BUY` (repurchase).

bats77:
Hello, I had some comments about DRIP transaction, as example :
real dividend (based on published data) 100 = 
first option : dividend = 100
second option : reinvestment on 15 stocks with a lasting dividend of 1,20 
last option :  revinvestment on 16 stocks with a buy operation consuming 10,4 cash

very close to Merge&Split transaction : 
1 Google = 100 (new Google) at date + cash (bonus or reinvestment or nothing) 
/ or / 19 At&t at date = 15 (new)At&t + 3 WarnerMedia + cash (bonus or reinvestment);

Nice and promising application by the way, congrats 👍

PR #1984, opened 2023-05-19
Feature/add error message if importing duplicates

dtslvr:


dtslvr:
I have continued your preliminary work @visrut7. Now an icon with an error message is displayed when you try to import a duplicate.

import-duplicate

PR #1998, opened 2023-05-22
Improve style on about page

secondl1ght:
Hello, I just recently started self-hosting and checking out Ghostfolio and it is a cool project! One thing I noticed on the About page is that the Changelog & License button is not centered. I decided to put in a PR to fix this. Thanks :D

BEFORE:
image

AFTER:
image


dtslvr (approved):
Thank you very much for the improvement @secondl1ght 😄

PR #2075, opened 2023-06-15
Feature/added ability to add Asset Profile from Admin Market data

aGhosh02:


apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
```diff
@@ -241,4 +245,46 @@ export class AdminMarketDataComponent implements OnDestroy, OnInit {
...
+            const dataSource = data?.dataSource;
+            const symbol = data?.symbol;
+
+            if (dataSource && symbol) {
+              this.adminService
+                .addAssetProfile({ dataSource, symbol })
+                .pipe(takeUntil(this.unsubscribeSubject))
+                .subscribe({
+                  next: () => {
+                    window.location.reload();
```
aGhosh02, line 281: Reloading page here. Let me know if you want to call the fetchAdminMarketData again instead.

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
```diff
@@ -241,4 +245,46 @@ export class AdminMarketDataComponent implements OnDestroy, OnInit {
           });
       });
   }
+
+  private openCreateAssetProfileDialog() {
+    this.dataService
+      .fetchAdminMarketData({})
```
dtslvr, line 251: Hmm, I would not fetch market data here. I noticed it right away because it took over 6 seconds to open the dialog for me 🙈 (yes, there are too many asset profiles without paging)

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
dtslvr: I also did it like this. I think it's okay for the moment

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
```diff
@@ -241,4 +245,46 @@ export class AdminMarketDataComponent implements OnDestroy, OnInit {
...
+      .pipe(
+        combineLatestWith(this.userService.get()),
+        takeUntil(this.unsubscribeSubject)
+      )
+      .subscribe(([adminMarketData, user]) => {
+        this.user = user;
+        const dialogRef = this.dialog.open(CreateAssetProfileDialog, {
+          autoFocus: false,
+          data: <CreateAssetProfileDialogParams>{
+            existingSymbols: adminMarketData.marketData,
```
dtslvr, line 261: I would not pass `existingSymbols`

apps/client/src/app/components/admin-market-data/admin-market-data.module.ts
```diff
@@ -8,17 +8,21 @@ import { GfActivitiesFilterModule } from '@ghostfolio/ui/activities-filter/activ
...
     CommonModule,
     GfActivitiesFilterModule,
     GfAssetProfileDialogModule,
+    GfCreateAssetProfileDialogModule,
     MatButtonModule,
     MatMenuModule,
     MatSortModule,
-    MatTableModule
+    MatTableModule,
+    RouterLink
```
dtslvr, line 25: I would import `RouterModule` here for consistency

apps/client/src/app/components/admin-market-data/create-asset-profile-dialog/create-asset-profile-dialog.component.ts
```diff
@@ -0,0 +1,79 @@
...
+@Component({
+  host: { class: 'h-100' },
+  selector: 'gf-create-asset-profile-dialog',
+  templateUrl: 'create-asset-profile-dialog.html',
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CreateAssetProfileDialog implements OnInit, OnDestroy {
+  public createAssetProfileForm: FormGroup;
+
+  constructor(
```
dtslvr, line 32: Please explicitly add `public` to the methods: `public constructor()...`

apps/client/src/app/components/admin-market-data/create-asset-profile-dialog/create-asset-profile-dialog.component.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.createAssetProfileForm.controls['searchSymbol'].value.dataSource,
+      symbol: this.createAssetProfileForm.controls['searchSymbol'].value.symbol
+    });
+  }
+
+  onCancel() {
+    this.dialogRef.close();
+  }
+
+  validateSymbolAlreadyExists(
```
dtslvr, line 63: I would not validate in the frontend. In the future, the client will no longer have the entire data at its disposal.

apps/api/src/app/admin/admin.service.ts
```diff
@@ -35,6 +37,38 @@ export class AdminService {
     this.baseCurrency = this.configurationService.get('BASE_CURRENCY');
   }
 
+  public async addAssetProfile({
+    dataSource,
+    symbol
+  }: UniqueAsset): Promise<SymbolProfile | never> {
+    try {
+      const assetProfile = await this.dataProviderService.getAssetProfiles([
```
dtslvr, line 45: `const assetProfiles =`

apps/api/src/app/admin/admin.service.ts
```diff
@@ -35,6 +37,38 @@ export class AdminService {
...
+        { dataSource, symbol }
+      ]);
+
+      if (!assetProfile[symbol].currency) {
```
dtslvr, line 49: Why do you check for `currency`?

```
if (!assetProfiles[symbol]) {
```

apps/api/src/app/admin/admin.controller.ts
```diff
@@ -328,6 +328,27 @@ export class AdminController {
     });
   }
 
+  @Post('profile-data/:dataSource/:symbol')
+  @UseGuards(AuthGuard('jwt'))
```
dtslvr, line 332: Can you add this line below `@UseGuards(AuthGuard('jwt'))`:

```
@UseInterceptors(TransformDataSourceInRequestInterceptor)
```

dtslvr (changes requested):
This will be a nice addition 😎 
I have added some comments, mainly because the validation cannot be done in the frontend.

apps/api/src/app/admin/admin.service.ts
aGhosh02: I am checking for currency because the function returns the below son even when it does not get any data.
{
symbolName:{
datasourse: "providedDatasource"
symbol: "symbolName",
currency: null
}
}

apps/api/src/app/admin/admin.service.ts
aGhosh02: In this case it is always going to be only one assetProfile. So assetProfile makes more sense to me.
But if you want it to be assetProfiles let me know.

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
aGhosh02: Did it the right way now

dtslvr: Nice!

apps/api/src/app/admin/admin.service.ts
dtslvr: Ok

dtslvr (approved):
I have made some changes.
Thanks again for implementing this feature!

PR #2076, opened 2023-06-15
Feature/setup language localization for portuguese

dtslvr:


apps/client/src/app/app.component.html
```diff
@@ -150,6 +150,9 @@
           <li>
             <a href="../nl" title="Ghostfolio in Nederlands">Nederlands</a>
```
luispinho, line 151: Even though this wasn't changed in this PR,  given all the other options mention the language itself (not the country) should this be `Ghostfolio in Dutch`?

dtslvr: I think it's fine like this 😃  See [here](https://translate.google.com/?sl=en&tl=nl&text=Dutch&op=translate)!

luispinho: Ah. Did not know this. Thanks for sharing 🙌

Issue #1322, opened 2022-10-04
Autofill token field with password managers


martinjuhasz:
**Describe the bug**

Currently my password manager (1password) is unable to automatically pick up the security token field and fill it (or store it automatically as the password).

**To Reproduce**
1. use 1Password as exprected
2. no popup or autofill will work

**Expected behavior**

my password manager should be able to guess that the security token is basically a password and treat it like that.

**Environment (please complete the following information):**
 - Ghostfolio Version v1.201.0
 - FIrefox Latest
 - macOS Monterey Latest
 - 1Password Latest

1Password provides some docs here but i guess this is true for all password managers out there:
https://developer.1password.com/docs/web/compatible-website-design/

dtslvr:
Thank you for your request @martinjuhasz. Another conversation in _Ideas_ about this topic has been started some time ago: https://github.com/ghostfolio/ghostfolio/discussions/906

I am not affected by it, so I will not work on it. Do you want to implement it?

daxter44:
I'm thinking about this solution, I use KeePass and the built-in password manager on MacOs, both have fields for Password and Login, no field for token. 
In addition, according to the docs of 1Pass, looks for forms with username and password on the page. We dont have them. The only solution I figured out is to add to the token field:  "id="username-field""and see what will heppened

dtslvr:
Hey @daxter44

What would happen if there was a (hidden) field for the username and you just add the password (token) in _KeePass_? Have you already played around with the code?

daxter44:
@dtslvr
So i prepared some PoC. I downloaded and configured 1Pass, its works similiar to KeePass.
Both programs needs only field with Password type, username is not neccesary.
But on the login modal dialog we must replace <textarea/> to < input/>. Password managers don't work with <textarea/> tag.

Its still not ideal aproach beacouse user has toadd manualy entry into password manager (and put the token into password field). But its work in further logs in attempts.

If You approve this solution i can prepare PR :)

dtslvr:
Yes, please create a PR with your initial solution and we can build on it.

daxter44:
@dtslvr  ready for review :) not a lot of code but its my begining :)

Sl1van:
1password now works as intended:
![image](https://github.com/ghostfolio/ghostfolio/assets/58076574/e45f80f6-8a31-4128-920a-366a8240ccdc)

Tested with 1password version 1.10.30 and Ghostfolio 2.76.0
[...]

PR #2124, opened 2023-07-05
Feature/Better login dialog

bptrgx:
Better login dialog:
- the token is hidden by default,
- hide/show button with the usual eye icon,
- hitting the enter key send the login request,
- the input field should be detected by password manager (I only tested Kee), because of `<form id="login">` and `<input name="password">`.

Now:
Screenshot

With this PR:
Screenshot
Screenshot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


bptrgx:
It might close #1322.

dtslvr (approved):

dtslvr:
Thank you for the groundwork @bptrgx 🤓

PR #2133, opened 2023-07-08
Feature/hide average buy price in position detail chart if no holding

dtslvr:


libs/ui/src/lib/line-chart/line-chart.component.ts
```diff
@@ -123,7 +123,7 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
     const marketPrices = [];
 
     this.historicalDataItems?.forEach((historicalDataItem, index) => {
-      benchmarkPrices.push(this.benchmarkDataItems?.[index]?.value);
+      benchmarkPrices.push(this.benchmarkDataItems?.[index]?.value || null);
```
bptrgx, line 126: I wouldn't set null values in the line-chart component, as this component is not only used for the average unit price, but also for the net performance on the home page where zero shouldn't be hidden.

libs/ui/src/lib/line-chart/line-chart.component.ts
```diff
@@ -154,7 +154,8 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
           data: benchmarkPrices,
           fill: false,
           label: this.benchmarkLabel,
-          pointRadius: 0
+          pointRadius: 0,
+          spanGaps: false
```
bptrgx, line 158: Shorter and easier to understand than my complex dataset split. 👍

libs/ui/src/lib/line-chart/line-chart.component.ts
dtslvr: Good point

bptrgx (approved):
Looks good to me. 👍

PR #2145, opened 2023-07-12
Added Stake order type for better crypto currency handling

dandevaud:
Hi @dtslvr 

First of all thank you very much for that awesome project you have created and all the effort you have put in.

While using ghostfolio I stumbled accross strange performance graphs:
Performance-Before

Hence, I started debugging ghostfolio locally and saw that for the performance calculation it uses the orders and their prices.
Unfortunately, I have a lot of BUY orders with unit prices of 0, it was a workaround to reflect stake rewards earned on crypto currencies. 
This was the reason for the performance graph to be skewed, as on the days with stake rewards the total investment of the specific single currency was 0 (quantity * orderPrice for that day).

I saw that one person already requested to add stake rewards as order type. Therefore, I started working on implementing those and added the following features:

- Support for order type STAKE in DB and controllers (CRUD)
- Updated the stake handling when calculating performance graphs (no impact on summary and total performance, yet)
- In the position detail dialog instead of the dividends the stake rewards are displayed if there are stake rewards but no dividends

The new order type could as well be used for Stock dividends, not sure if one could reflect split as well.

Let me know if you have any doubts or question, happy to help and it would of course be awesome if the pull request gets approved


dtslvr:
Hi @dandevaud

First of all, thank you very much for your contribution. It's great to see that you were able to set up and tweak the project.

I have a question I'd like to further discuss regarding the new type `STAKE`. What do you think is the advantage over recording a `BUY` activity at the day price for staking and adding a tag (e.g. _Staking_) to the activity? I already use this concept for gifted stocks.

Again, thanks for your contribution. Looking forward to your response.

dandevaud:
Hi @dtslvr 

Thanks for your swift reply.

I honestly have never used the tag functionality on transactions and hence might have missed some of its functionality.

Wouldn't adding staking or gifted stocks using your concept affect the average unit price, the holding performance and eventually the overall performance as well?
Assuming you bought 1 ETH @ 1000, received 1 ETH when it was @ 3000 through staking and the price is now @ 2000.

Recording the staking as buy at the daily price you would have 2 ETH with the average unit price of 2000 and a performance of 0%.
The new staking feature would handle them as buy but with no price (only used in the performance graphs).
Hence, you would have 2 ETH at an average price of 500 (1 @1000 + 1 @0) and performance of 300%.

Latter makes more sense in terms of performance, as you effectively only have paid 1000 but now hold 2 ETH.
As well as your gifted stocks are recorded as gifts and the actual unit price they have is all performance.
Indeed this can lead to infitive performance on a single holding (only gifted stocks), in my opinion this is fine as you never really bought the stock.

Let me know if that makes sense or if I missed something.

dandevaud:
Hi @dtslvr 

That concept should work for stock dividends as well.
it's just the naming of the new type which was easier to be STAKE than something with stockdividend.

Basically, Staking and Stock Dividends are the same in terms of revenue and performance in my opinion.

The other solution without an additional type would be to use 0 unit price buys, but then the performance graph need to refactored to use the daily prices instead of the order prices. This would remove the volatility of the graph in the picture above.

VibeNL:
Any progress on this 👼? Most of my crypto positions will be incorrect without something like this 😄

Joren-vanGoethem:
don't forget that dividends can be taxed ( and i would like to see the option to automatically deduct a specific amount of tax to reflect my real-world income and not the before-tax income)

stake rewards not really

VibeNL:
@dandevaud , I don't think this should be closed. A workaround in another project has been completed. This still should be fixed on ghostfolio itself 😅

dandevaud:
Hi @VibeNL 

Indeed, I just deleted the branch in the my fork of the repo to clean it up.
I forgot that this PR is still pending and that it will automatically be closed if I delete the branc
Meanwhile the PR had merge conflicts and I have moved forward using my own implementation, hence this PR is not up to date anymore.

But all changes can be found in the commit https://github.com/dandevaud/ghostfolio/commit/5d06b131418a0d650f8436d19f9eb54f3afc4b48.

Feel free to cherry pick it on any implementation you want. You might need to create a prisma migration when using it.

albertsalles4:
Any advances on this? Maybe instead of adding a new STAKE option for rewards we could choose to add Dividends in the same currency (crypto or stock) instead of cash (which would essentially be acquiring it at the price of 0), that would just require a selection button in the Dividend dialog. How would you see this option?

PR #2156, opened 2023-07-15
Feature/upgrade to angular 16

dtslvr:


dtslvr:
```
Error: error TS2688: Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point of type library 'node' specified in compilerOptions
```

dtslvr:
Current issue with `{workspaceRoot}`:
favicon

Issue #2181, opened 2023-07-25
[BUG] Liabilities do not subtract value from assets


Hugo-Persson:
The Issue tracker is **ONLY** used for reporting bugs. New features should be discussed in our [Slack](https://join.slack.com/t/ghostfolio/shared_invite/zt-vsaan64h-F_I0fEo5M0P88lP9ibCxFg) community or in [Discussions](https://github.com/ghostfolio/ghostfolio/discussions).

**Bug Description**

**To Reproduce**

1. Create an account with one value item
2. Then a couple of Liabilities with negative balance
3. The total will not be value item - liabilities

**Expected behavior**

The total should be the total value minus the liabilities
**Screenshots**

![image](https://github.com/ghostfolio/ghostfolio/assets/47189713/0774ab90-e724-44b0-a2e9-58757dcb6101)

**Logs**

**Environment**

- Cloud or Self-hosted: Self hosted
- Ghostfolio Version X.Y.Z:  Ghostfolio v1.292.0.
- Browser: Edge
- OS: Windows

**Additional context**

Hugo-Persson:
Started work on Issue here https://github.com/Hugo-Persson/ghostfolio/tree/Hugo-Persson/issue2181

PR #2184, opened 2023-07-26
Fix total account value calculation for liabilities

Hugo-Persson:
**Still in draft**

Found way to change total calculation, still need to tweak account modal and check other places


Hugo-Persson:
See https://github.com/ghostfolio/ghostfolio/issues/2181

dtslvr (approved):
Thank you very much for your contribution, @Hugo-Persson 🙂

I have slightly refactored your code (see commits if you are interested).

PR #2302, opened 2023-09-07
Improve import error message

httpiga:
Yahoo once used `BTCEUR` as a ticker for BTC expressed in €, but recently they changed it to `BTC-EUR`. 
Old Ghostfolio backups having `BTCEUR` as activity symbol caused the import to fail due to a `activities.2.currency ("EUR") does not match with "undefined"` error.
This error is misleading, since the problem relies in the (now) wrong symbol, and not in the currency, so the `activities.${index}.symbol ("${symbol}") is not valid for the specified data source ("${dataSource}")` error should be thrown instead.
The motivation why the wrong error is thrown is because, in case of missing/wrong symbol, `assetProfile` has this shape: 
```
{
  assetClass: undefined,
  assetSubClass: undefined,
  currency: undefined,
  name: undefined,
  dataSource: 'YAHOO',
  symbol: 'BTCEUR'
}
```
so the `if (!assetProfile)` condition is never met.
I extended the condition to one of its key (`name`, but I chosen it in an arbitrary way), so now the correct error is displayed (`activities.2.symbol ("BTCEUR") is not valid for the specified data source ("YAHOO")`).

Moreover in case of an import error the activity displayed in the frontend may be different from the one which actually caused the error in the server: in fact, in my case it showed a non-BTC activity. 
I think this is caused byLodash's `uniqBy` which doesn't preserve the array order. Since I don't know why it was introduced, I haven't touched it but just left a comment as a reminder.

_Sorry for the irrelevant commits, I already forked Ghostfolio to do some personal customisation and GitHub doesn't allowed me to create another fork. You can check the "files changed" tab._


dtslvr:
> Thank you for your quick feedback, @dtslvr 🙏

I even tried to set you as an author of #2305, not sure if it worked 😕

> Shall I open an issue for the bug I described?

Yes, if you can reproduce the bug. According to ChatGPT, the "`_.uniqBy()` function preserves the order of the elements in the original array."

PR #2316, opened 2023-09-10
Speeding up importing of activities

rafacla:
Dears,

Adding a auxiliary list to avoid repetitive access to Yahoo API during importing process I was able to import 611 activities in my local testing machine, this may speed up the process specially for those that make recurring purchases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added support to drop a file in the import activities dialog
+
 ### Changed
 
 - Harmonized the logger output: <symbol> (<dataSource>)
--- a/apps/client/src/app/directives/file-drop/file-drop.directive.ts
+++ b/apps/client/src/app/directives/file-drop/file-drop.directive.ts
@@ -1,0 +1,28 @@
+import { Directive, HostListener, Output, EventEmitter } from '@angular/core';
+
+@Directive({
+  selector: '[gfFileDrop]'
+})
+export class FileDropDirective {
+  @Output() filesDropped = new EventEmitter<FileList>();
+
+  @HostListener('dragenter', ['$event']) onDragEnter(event: DragEvent) {
+    event.preventDefault();
+    event.stopPropagation();
+  }
+
+  @HostListener('dragover', ['$event']) onDragOver(event: DragEvent) {
+    event.preventDefault();
+    event.stopPropagation();
+  }
+
+  @HostListener('drop', ['$event']) onDrop(event: DragEvent) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    // Prevent the browser's default behavior for handling the file drop
+    event.dataTransfer.dropEffect = 'copy';
+
+    this.filesDropped.emit(event.dataTransfer.files);
+  }
+}
--- a/apps/client/src/app/directives/file-drop/file-drop.module.ts
+++ b/apps/client/src/app/directives/file-drop/file-drop.module.ts
@@ -1,0 +1,9 @@
+import { NgModule } from '@angular/core';
+
+import { FileDropDirective } from './file-drop.directive';
+
+@NgModule({
+  declarations: [FileDropDirective],
+  exports: [FileDropDirective]
+})
+export class GfFileDropModule {}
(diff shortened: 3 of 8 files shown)


apps/api/src/app/import/import.service.ts
```diff
@@ -555,11 +555,16 @@ export class ImportService {
       }
     );
 
+    var listOfValidatedAssets: string[] = [];
```
dtslvr, line 558: I think a map is more intuitive for this kind of check:
```
const validatedAssetsMap: { [assetKey: string]: boolean } = {};
```

apps/api/src/app/import/import.service.ts
```diff
@@ -580,12 +585,13 @@ export class ImportService {
...
         ) {
           throw new Error(
-            `activities.${index}.currency ("${currency}") does not match with "${assetProfile.currency}" and no exchange rate is available from "${currency}" to "${assetProfile.currency}"`
+            `activities.${index}.symbol("${symbol}").currency ("${currency}") does not match with "${assetProfile.currency}" and no exchange rate is available from "${currency}" to "${assetProfile.currency}."`
           );
         }
 
         assetProfiles[getAssetProfileIdentifier({ dataSource, symbol })] =
           assetProfile;
+        listOfValidatedAssets.push(symbol);
```
dtslvr, line 594: `validatedAssetsMap[getAssetProfileIdentifier({ dataSource, symbol })] = true;`

apps/api/src/app/import/import.service.ts
```diff
@@ -555,11 +555,16 @@ export class ImportService {
...
+    var listOfValidatedAssets: string[] = [];
+
     for (const [
       index,
       { currency, dataSource, symbol }
     ] of uniqueActivitiesDto.entries()) {
-      if (dataSource !== 'MANUAL') {
+      if (
+        dataSource !== 'MANUAL' &&
+        listOfValidatedAssets.includes(symbol) === false
```
dtslvr, line 566: `!validatedAssetsMap[getAssetProfileIdentifier({ dataSource, symbol })]`

apps/api/src/app/import/import.service.ts
```diff
@@ -580,12 +585,13 @@ export class ImportService {
           )
         ) {
           throw new Error(
-            `activities.${index}.currency ("${currency}") does not match with "${assetProfile.currency}" and no exchange rate is available from "${currency}" to "${assetProfile.currency}"`
+            `activities.${index}.symbol("${symbol}").currency ("${currency}") does not match with "${assetProfile.currency}" and no exchange rate is available from "${currency}" to "${assetProfile.currency}."`
```
dtslvr, line 588: This looks strange 😮

dtslvr (changes requested):
Great optimization, @rafacla! 🙌🏻 

I have added some comments.

apps/api/src/app/import/import.service.ts
rafacla: ooops, this one was an initial change I had made to identify which item was not being found at Yahoo, but I saw you already had fixed it a few lines before, we can drop this one.

dtslvr (changes requested):
I'm confused. There was already an asset profiles map (see `assetProfiles`). Shouldn't we just use that in the check instead of introducing `validatedAssetsMap`?

dtslvr:
What happened now? I think you were on the right track @rafacla! 🙃

This could already have an effect, but you don't need to populate the map again.

```
if (dataSource !== 'MANUAL') {
```
to
```
if (dataSource !== 'MANUAL' && !assetProfiles[getAssetProfileIdentifier({ dataSource, symbol })]) {
```

Can you give it a last try?

rafacla:
Hi Thomas,

Actually this PR closed itself, because I had to rebase my git and open different branches to deal with the memory leak of TZ before cotinuing with this one, because the memory leak was causing me problems to debug the code after importing the first activities (otherwise I'd need to reset the database).

But, yesterday I was reading the code more troughly and acutally your code is already filtering unique assets in the process, so the changes I had initially proposed would not improve it afterall, but the thing is, I had installed Ghostfol.io in a Raspberry Pi 4, so in the Pi my former code indeed improved the time of importing, but I'll try to replicate the cases in the Pi in this weekend to understand what indeed happened to improve the speed.

As the Pi hardware is way weaker than my laptop, any improvement has more impact on it than on my dev machine.

Thanks!

dtslvr:
Ah okay, I see. It's impossible to reopen this PR. So please create a new one if you have some insights 🙂

Issue #2287, opened 2023-09-02
Support dd.MM.yyyy date format in activities import


gogo199432:
**Description**

Unable to import activity with date format: 01.09.2023 (dd.MM.yyyy)
I get "activities.0.date is not valid" message when trying to import via CSV

**Environment**

- Self-hosted
- Ghostfolio Version latest
- Firefox
- K8S

dtslvr:
Thanks for reporting @gogo199432.

Are you interested in extending the `parseDate()` function [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/services/import-activities.service.ts#L226)?

FraneCaleta:
@dtslvr Feel free to asign me on this one. I'm already taking a look at it.

dtslvr:
@FraneCaleta can you extend [ok.csv](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok.csv) with that date format that it can be tested easily?

FraneCaleta:
> @FraneCaleta can you extend [ok.csv](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok.csv) with that date format that it can be tested easily?

Maybe, let me finish the basics first, it got more complicated than I expected :)

PR #2362, opened 2023-09-21
Feature/improve supported date formats in activities import

FraneCaleta:
As mentioned in issue #2287 
_Unable to import activity with date format: 01.09.2023 (dd.MM.yyyy)
I get "activities.0.date is not valid" message when trying to import via CSV_

It should not be a problem anymore once this PR gets merged. 
Some different date formats have been added as well + impoved code reusability.

I have made common function inside `helper.ts` and then reused it in `import-activities.service.ts`. Is this a good approach or you would prefer to keep old basic fuction inside of `helper.ts` and then write all of the logic inside `import-activities.service.ts`?
Any opinion is welcome :)


FraneCaleta:
Need to adapt test case examples still

FraneCaleta:
Added support for [ok-vti-buy-on-ibkr.csv](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok-vti-buy-on-ibkr.csv) example -> yyyyMMdd format

dtslvr (approved):

Issue #2363, opened 2023-09-21
[BUG] Total value of activities not showing when DIVIDEND activity is added


FraneCaleta:
**Bug Description**

As the title says, total value of activities not showing at all when _DIVIDEND_ activity is added.

**To Reproduce**

1. Navigate to _portfolio/activities_ tab
2. Add a few buy orders of any stock
3. Total value should be counter properly now
4. Add a dividend from some stock (E.G. AAPL dividend)
5. Total value is not shown anymore

**Screenshots**

![image](https://github.com/ghostfolio/ghostfolio/assets/37557775/13e719b5-8248-4c53-b8ee-e405cc02cf1e)

**Environment**

- Self-hosted
- Ghostfolio Version 2.4.0
- Google Chrome
- Windows

brhahlen:
Did not notice it before, but can confirm.
Also running 2.4.0, using Firefox

FraneCaleta:
@dtslvr You can asign me to this issue, I'll have a look at it.

PR #2364, opened 2023-09-21
Bugfix - issue with total value of activites not showing

FraneCaleta:
For more informations it's best to check issue #2363.

It's a small fix which prevents "null" from showing to the user.
So basically, if we have one of these activities: **Dividend, Fee, Interest, Liability** or **Sell** we will just ignore them and continue with our loop which counts the total value of types **Buy** and **Item**.

image

In this picture we can see that totalValue is now being properly displayed and counted.
You can find a picture with bug in the issue #2363.

### DIFF
--- a/libs/ui/src/lib/activities-table/activities-table.component.ts
+++ b/libs/ui/src/lib/activities-table/activities-table.component.ts
@@ -399,7 +399,7 @@ export class ActivitiesTableComponent implements OnChanges, OnDestroy, OnInit {
           type === 'LIABILITY' ||
           type === 'SELL'
         ) {
-          return null;
+          continue;
         }
       } else {
         return null;


FraneCaleta:
> The problem with this implementation I see is the following:
> 
> ```
> BUY  1 Alphabet $1000
> SELL 1 Alphabet $1200
> -----------------------
>                 $1000 <-- ???
> ```

I guess the main question is how exactly do we want this to function, the implementation part seems pretty straight forward :)

In the example you showed, do you think the best behaviour would be to display 2200? Or -200?

Should we perhaps add all **Item, Dividend** and **Buy** types and then substract them with Liabilities, Fees, Interests and Sell types?

I think it's better to discuss and choose one approach rather than displaying nothing.

PR #2405, opened 2023-09-30
Add additional time horizon options for portfolio performance.

prashmohan:
This mirrors the time horizons used in popular finance tracking websites online.


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1505,6 +1506,24 @@ export class PortfolioService {
           subDays(new Date().setHours(0, 0, 0, 0), 1)
         ]);
         break;
+      case '5d':
```
dtslvr, line 1509: Is this the correct logic? I would rather expect it to be the range from Monday to today (aka _Week to date_).

dtslvr (changes requested):
Thank you for you contribution @prashmohan. Your changes look solid.

However, I don't like the user interface and I don't want the app to be unnecessarily complicated.
ghostfolio-homescreen

My idea is to combine the filter function and date range selection in the future.

apps/api/src/app/portfolio/portfolio.service.ts
prashmohan: WTD is fair logic to have. But the way that Yahoo finance (a portfolio tracking system that I really like) implements it appears to be the last 5 trading days - so not exactly what is implemented here either.

PR #2406, opened 2023-10-01
Use port numbers from environment file.

prashmohan:
Allow overriding of the postgres and redis forwarded ports in the docker configuration to allow running on alternate ports where the default ports may already be in use.


docker/docker-compose.dev.yml
dtslvr, line 1: Can you please adopt the change(s) also to `docker-compose.yml` and `docker-compose.build.yml`?

dtslvr (changes requested):

docker/docker-compose.dev.yml
prashmohan: docker-compose.yml and docker-compose.build.yml should not require any modification, since they do not export the postgres and redis ports on the host machine.

dtslvr (changes requested):

dtslvr:
What do you think of my changes @prashmohan?

prashmohan:
> What do you think of my changes @prashmohan?

I dont think this change would work though. This is overriding the port number that postgres is listening on inside its container. Looking at the environment variables that postgres uses (in https://hub.docker.com/_/postgres), it does not appear that the postgres container can be configured to have the service listening on a different port. In this case, I assume the ghostfolio server will be unable to reach postgres.

The way I am intending to use the environment variable specified in ghostfolio's config is to configure the port being used on the host machine without tampering with the port configuration used within the containers.

dtslvr:
> I dont think this change would work though.

Thanks for your explanation @prashmohan. I have reverted my changes. Now, `${POSTGRES_PORT}` is only used (optionally) in `docker-compose.dev.yml`. You are free to define `POSTGRES_PORT` in your `.env` file. What do you think?

prashmohan:
> What do you think?

Yup. I think that works best for development.

dtslvr (approved):

Issue #2383, opened 2023-09-24
Support copy-assets Nx target on Windows


dtslvr:
ℹ️ This task is suitable for Windows users!

The `copy-assets` _Nx_ target is currently not working on _Windows_.

Error on Windows when running `yarn start:client`:

```
The syntax of the command is incorrect.
cp: cannot overwrite non-directory 'dist/apps/client/assets' with directory 'apps/client/src/assets'
Warning: run-commands command "mkdir -p dist/apps/client" exited with non-zero status codeWarning: run-commands command "cp -r apps/client/src/assets dist/apps/client" exited with non-zero status code
```

Come up with a solution that works on _Unix_ and _Windows_ systems.

Here is a first try as an inspiration that does not seem to work: https://github.com/ghostfolio/ghostfolio/pull/2296

PR #2410, opened 2023-10-01
Support copy-assets Nx target on Windows

KK4ito:
Hi there, 

I have a possible solution for the issue #2383. The plugin run-script-os allows us to write npm platform specific scripts.

I adjusted every npm script which contains the "nx run client:copy-assets" to use the platform specific command.
I also had to duplicate the nx command "copy-assets".

Thanks for a feedback.


dtslvr (changes requested):
Thanks a lot for providing a solution @KK4ito.

Isn't there a more elegant solution? 🤔 I don't like so much that the paths are in there twice now. This way, unfortunately, it is often forgotten that it must be adjusted in both locations.

KK4ito:
Thanks for the feedback @dtslvr.
I changed the solution by using the "shx" plugin.

package.json
```diff
@@ -194,6 +194,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "replace-in-file": "7.0.1",
+    "shx": "^0.3.4",
```
dtslvr, line 197: That looks promising. Can you fix the version (remove `^`) and run `yarn install` again?

dtslvr (changes requested):

package.json
KK4ito: done, see commit c0d89362625faf4dd0e00e0fb94df6d57fd4aa3e

dtslvr (approved):
Thank you Kevin! 👍🏻

PR #2421, opened 2023-10-03
Update object.helper.ts

virus231:


apps/api/src/helper/object.helper.ts
```diff
@@ -3,7 +3,7 @@ import { cloneDeep, isArray, isObject } from 'lodash';
 
 export function hasNotDefinedValuesInObject(aObject: Object): boolean {
   for (const key in aObject) {
-    if (aObject[key] === null || aObject[key] === null) {
+    if (aObject[key] === null || aObject[key] === undefined) {
```
dtslvr, line 6: Wow, this looked strange. Thanks for creating a patch @virus231. How did you notice that?

dtslvr (approved):

apps/api/src/helper/object.helper.ts
virus231: Just looking at your code

Issue #2424, opened 2023-10-04
Use current version from AdminData (endpoint) in Admin Control panel


dtslvr:
Change the current version of Ghostfolio in the [Overview](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-overview) of the Admin Control panel.

<img width="1225" alt="version-in-admin-control-panel" src="https://github.com/ghostfolio/ghostfolio/assets/4159106/447e056b-0aef-4997-965d-cac03ca0c45e">

The version should be fetched from `AdminData` instead of `environment.version` (deprecated for security reasons).

The template should be changed to a `gf-value` component (for the loading indicator):

```
<gf-value [value]="version" />
```

Franqsanz:
Hello @dtslvr , that would be the only change?

dtslvr:
> Hello @dtslvr , that would be the only change?

Yes, this is a relatively small change. Good as a first issue 🙂

Franqsanz:
Where to import AdminData from?

dtslvr:
> Where to import `AdminData` from?

You can get the version from [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-overview/admin-overview.component.ts#L207).

PR #2427, opened 2023-10-04
fix current version from AdminData #2424

akshaywadatkar:
fix the version bug for #2424

### DIFF
--- a/apps/client/src/app/components/admin-overview/admin-overview.component.ts
+++ b/apps/client/src/app/components/admin-overview/admin-overview.component.ts
@@ -43,7 +43,7 @@ export class AdminOverviewComponent implements OnDestroy, OnInit {
   public transactionCount: number;
   public userCount: number;
   public user: User;
-  public version:any
+  public version: string;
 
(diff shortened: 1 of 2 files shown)


apps/client/src/app/components/admin-overview/admin-overview.component.ts
```diff
@@ -43,7 +43,7 @@ export class AdminOverviewComponent implements OnDestroy, OnInit {
   public transactionCount: number;
   public userCount: number;
   public user: User;
-  public version = environment.version;
+  public version:any
```
dtslvr, line 46: `public version: string;`

dtslvr (changes requested):
Can you change `{{ version }}` to `<gf-value [value]="version" />` in the template (`admin-overview.component.html`)?

apps/client/src/app/components/admin-overview/admin-overview.component.ts
akshaywadatkar: sure

dtslvr (approved):

PR #2428, opened 2023-10-04
fix: prevent empty form submission in account dialog

### LINKED ISSUE #2418: [BUG] Invalid account form is submitted by pressing enter
**Bug Description**

When I press the enter key in the state of the `CreateOrUpdateAccountDialog` as shown below, an http request is sent to the backend and the dialog is closed.

**Expected behavior**

Should not send an http request to the backend and keep the dialog open if the form is invalid.

**Screenshots**

add-account-press-enter

**Logs**

```
POST http://localhost:4200/api/v1/account 400 (Bad Request)
```

anirudhsudhir:
Fixes #2418

Changed default behavior of {keydown.enter} in `<input>` which previously led to form submissions when fields were empty

### DIFF
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
@@ -13,7 +13,7 @@
         <input
           formControlName="name"
           matInput
-          (keydown.enter)="$event.preventDefault()"
+          (keydown.enter)="$event.stopPropagation()"
         />
       </mat-form-field>
     </div>
@@ -34,7 +34,7 @@
           formControlName="balance"
           matInput
           type="number"
-          (keydown.enter)="$event.preventDefault()"
+          (keydown.enter)="$event.stopPropagation()"
         />
         <span class="ml-2" matTextSuffix>{{ data.account.currency }}</span>
       </mat-form-field>
@@ -75,7 +75,7 @@
     </div>
   </div>
   <div class="justify-content-end" mat-dialog-actions>
-    <button i18n mat-button (click)="onCancel()">Cancel</button>
+    <button i18n mat-button type="button" (click)="onCancel()">Cancel</button>
     <button
       color="primary"
       mat-flat-button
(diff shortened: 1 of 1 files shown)


apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
```diff
@@ -10,7 +10,11 @@ <h1 *ngIf="!data.account.id" i18n mat-dialog-title>Add account</h1>
     <div>
       <mat-form-field appearance="outline" class="w-100">
         <mat-label i18n>Name</mat-label>
-        <input formControlName="name" matInput />
+        <input
+          formControlName="name"
+          matInput
+          (keydown.enter)="$event.preventDefault()"
```
dtslvr, line 16: Would `(keyup.enter)="$event.stopPropagation()"` also work? Because then it would be consistent with the _Note_ textarea.

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
```diff
@@ -26,7 +30,12 @@ <h1 *ngIf="!data.account.id" i18n mat-dialog-title>Add account</h1>
     <div>
       <mat-form-field appearance="outline" class="w-100">
         <mat-label i18n>Cash Balance</mat-label>
-        <input formControlName="balance" matInput type="number" />
+        <input
+          formControlName="balance"
+          matInput
+          type="number"
+          (keydown.enter)="$event.preventDefault()"
```
dtslvr, line 37: Would `(keyup.enter)="$event.stopPropagation()"` also work? Because then it would be consistent with the _Note_ textarea.

dtslvr (changes requested):

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
anirudhsudhir: I tried replacing `(keydown.enter)="$event.preventDefault()"` with `(keyup.enter)="$event.stopPropagation()"`. Unfortunately, it doesn't seem to work and the empty form submission occurs on pressing Enter.

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
anirudhsudhir: I tried replacing `(keydown.enter)="$event.preventDefault()"` with `(keyup.enter)="$event.stopPropagation()"`. Unfortunately, it doesn't seem to work and the empty form submission occurs on pressing Enter.

dtslvr (approved):

Issue #2419, opened 2023-10-03
Change platform control from select to autocomplete in account dialog


dtslvr:
As platforms are growing, a simple select is not enough anymore: Change the platform form control from [Select](https://material.angular.io/components/select/overview) to [Autocomplete](https://material.angular.io/components/autocomplete/overview) in the [account dialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/pages/accounts/create-or-update-account-dialog). Make sure that only a platform from the list is a valid value.

<img width="795" alt="account-form" src="https://github.com/ghostfolio/ghostfolio/assets/4159106/7ed5aaa2-1b6f-4eef-abfb-958a985e39de">

KK4ito:
Hi @dtslvr , I can work on this if its still unassigned

PR #2429, opened 2023-10-04
replaced select component with autocomplete

KK4ito:
for #2419

- add validation to allow only values from list
- add new interface type: platform
- autoselect value if input matches with an entry from list

### DIFF
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
@@ -18,10 +18,8 @@ import { DataService } from '@ghostfolio/client/services/data.service';
 import { Observable, Subject } from 'rxjs';
 import { map, startWith } from 'rxjs/operators';
 
-import {
-  CreateOrUpdateAccountDialogParams,
-  Platform
-} from './interfaces/interfaces';
+import { CreateOrUpdateAccountDialogParams } from './interfaces/interfaces';
+import { Platform } from '@prisma/client';
 
 @Component({
   host: { class: 'h-100' },
@@ -33,8 +31,8 @@ import {
 export class CreateOrUpdateAccountDialog implements OnDestroy {
   public accountForm: FormGroup;
   public currencies: string[] = [];
-  public platforms: Platform[];
   public filteredPlatforms: Observable<Platform[]>;
+  public platforms: Platform[];
 
   private unsubscribeSubject = new Subject<void>();
 
@@ -49,7 +47,9 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
     const { currencies, platforms } = this.dataService.fetchInfo();
 
     this.currencies = currencies;
-    this.platforms = platforms;
+    this.platforms = platforms.map((platform) => {
+      return { ...platform, url: '' };
+    });
 
     this.accountForm = this.formBuilder.group({
       accountId: [{ disabled: true, value: this.data.account.id }],
@@ -59,9 +59,9 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
       isExcluded: [this.data.account.isExcluded],
       name: [this.data.account.name, Validators.required],
       platformId: [
-        this.platforms.find(
-          (platform) => platform.id === this.data.account.platformId
-        ),
+        this.platforms.find(({ id }) => {
+          return id === this.data.account.platformId;
+        }),
         this._autocompleteObjectValidator()
       ]
     });
@@ -106,22 +106,23 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
 
   public autoCompleteCheck() {
     const inputValue = this.accountForm.controls['platformId'].value;
-    const matchingEntry = this.platforms.find(
-      (platform) => platform.name === inputValue
-    );
-    this.accountForm.controls['platformId'].setValue(matchingEntry);
+    if (typeof inputValue === 'string') {
+      const matchingEntry = this.platforms.find(({ name }) => {
+        return name === inputValue;
+      });
+      this.accountForm.controls['platformId'].setValue(matchingEntry);
+    }
   }
 
   displayFn(platform: Platform) {
-    return platform && platform.name ? platform.name : '';
+    return platform?.name ? platform.name : '';
   }
 
   private _filter(value: string): Platform[] {
     const filterValue = value.toLowerCase();
-
-    return this.platforms.filter((platformEntry) =>
-      platformEntry.name.toLowerCase().includes(filterValue)
-    );
+    return this.platforms.filter(({ name }) => {
+      return name.toLowerCase().includes(filterValue);
+    });
   }
 
   private _autocompleteObjectValidator(): ValidatorFn {
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
@@ -115,7 +115,7 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
   }
 
   public displayFn(platform: Platform) {
-    return platform?.name ? platform.name : '';
+    return platform?.name ?? '';
   }
 
   private _filter(value: string): Platform[] {
@@ -127,7 +127,7 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
 
   private _autocompleteObjectValidator(): ValidatorFn {
     return (control: AbstractControl) => {
-      if (typeof control.value === 'string') {
+      if (control.value && typeof control.value === 'string') {
         return { invalidAutocompleteObject: { value: control.value } };
       }
       return null;
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
@@ -43,7 +43,7 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
     private formBuilder: FormBuilder
   ) {}
 
-  ngOnInit() {
+  public ngOnInit() {
     const { currencies, platforms } = this.dataService.fetchInfo();
 
     this.currencies = currencies;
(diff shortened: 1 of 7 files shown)


apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -4,14 +4,24 @@ import {
...
 import { UpdateAccountDto } from '@ghostfolio/api/app/account/update-account.dto';
 import { DataService } from '@ghostfolio/client/services/data.service';
-import { Subject } from 'rxjs';
+import { Observable, Subject } from 'rxjs';
+import { map, startWith } from 'rxjs/operators';
 
-import { CreateOrUpdateAccountDialogParams } from './interfaces/interfaces';
+import {
+  CreateOrUpdateAccountDialogParams,
+  Platform
```
dtslvr, line 23: Can you reuse `import { Platform } from '@prisma/client';` instead of your own interface?

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -23,7 +33,8 @@ import { CreateOrUpdateAccountDialogParams } from './interfaces/interfaces';
 export class CreateOrUpdateAccountDialog implements OnDestroy {
   public accountForm: FormGroup;
   public currencies: string[] = [];
-  public platforms: { id: string; name: string }[];
+  public platforms: Platform[];
+  public filteredPlatforms: Observable<Platform[]>;
```
dtslvr, line 37: Can you move up ⬆️ to keep the alphabetical order?

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -47,8 +58,22 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
       currency: [this.data.account.currency, Validators.required],
       isExcluded: [this.data.account.isExcluded],
       name: [this.data.account.name, Validators.required],
-      platformId: [this.data.account.platformId]
+      platformId: [
+        this.platforms.find(
+          (platform) => platform.id === this.data.account.platformId
```
dtslvr, line 63: Can you use the more explicit form for consistency?

```
({id}) => { return id === this.data.account.platformId; }
```

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,33 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
     this.unsubscribeSubject.next();
     this.unsubscribeSubject.complete();
   }
+
+  public autoCompleteCheck() {
+    const inputValue = this.accountForm.controls['platformId'].value;
+    const matchingEntry = this.platforms.find(
+      (platform) => platform.name === inputValue
```
dtslvr, line 110: `({ name }) => { return name === inputValue; }`

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,33 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+    );
+    this.accountForm.controls['platformId'].setValue(matchingEntry);
+  }
+
+  displayFn(platform: Platform) {
+    return platform && platform.name ? platform.name : '';
```
dtslvr, line 116: `return platform?.name ? platform.name : '';`

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,33 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+  }
+
+  private _filter(value: string): Platform[] {
+    const filterValue = value.toLowerCase();
+
+    return this.platforms.filter((platformEntry) =>
+      platformEntry.name.toLowerCase().includes(filterValue)
```
dtslvr, line 123: `{ return name.toLowerCase().includes(filterValue); }`

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
```diff
@@ -33,12 +33,23 @@ <h1 *ngIf="!data.account.id" i18n mat-dialog-title>Add account</h1>
...
+          [matAutocomplete]="auto"
+          (blur)="autoCompleteCheck()"
+        />
+        <mat-autocomplete #auto="matAutocomplete" [displayWith]="displayFn">
+          <mat-option
+            *ngFor="let platformEntry of filteredPlatforms | async"
+            [value]="platformEntry"
           >
-        </mat-select>
+            {{platformEntry.name}}
```
dtslvr, line 50: `{{ platformEntry.name }}`

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.module.ts
```diff
@@ -21,6 +22,7 @@ import { CreateOrUpdateAccountDialog } from './create-or-update-account-dialog.c
     MatFormFieldModule,
     MatInputModule,
     MatSelectModule,
+    MatAutocompleteModule,
```
dtslvr, line 25: Can you move up ⬆️  to keep the alphabetical order?

dtslvr (changes requested):
Cool 😎  I have added some comments.

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
KK4ito: yes sure, bd2b36622d4ccea0b9086ba6c7ba4f66be2010b3

because of the property "url" i had to map the infos from
`this.dataService.fetchInfo();`

with the following adjustment

```
 this.platforms = platforms.map((platform) => {
      return { ...platform, url: '' };
    });
```

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
KK4ito: done bd2b36622d4ccea0b9086ba6c7ba4f66be2010b3

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
KK4ito: thanks for the feedback, didnt know about the explicit form :D quite a long time since my last work with typescript

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
KK4ito: during testing, I encountered a bug with my previous implementation. I fixed it by checking the input value.

only for inputvalue of type string do the stuff (find, setting value)

inputValue can be of type platform, when a value from the list is selected by user interaction

KK4ito: additional thoughts: this approach (matching by platform name) only works if we can guarantee that there are no multiple platforms with the same name.
or else the current implementation returns the first match from the given array

KK4ito:
thanks for the fast review. I applied all suggestions,

KK4ito:
> Thank you very much for the changes, @KK4ito!
> 
> I have played around a bit and added some more comments. It happened to me several times that I clicked a search result, but the value was not applied. I can't reproduce it reliably and unfortunately there is no error in the console. Can you take a closer look?

Thank you for the feedback.
I tested it and I think I was be able to reproduce & fix it. This behaviour occured when there is a input in the text field during user click. In such cases the input value was from type string and the value was forcibly set to undefined. 94a99d188819fe371abc7649b75890e2c204e69a

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,34 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+    const inputValue = this.accountForm.controls['platformId'].value;
+    if (typeof inputValue === 'string') {
+      const matchingEntry = this.platforms.find(({ name }) => {
+        return name === inputValue;
+      });
+      this.accountForm.controls['platformId'].setValue(matchingEntry);
+    }
+  }
+
+  displayFn(platform: Platform) {
```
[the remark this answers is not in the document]
KK4ito: done

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,34 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+
+  private _filter(value: string): Platform[] {
+    const filterValue = value.toLowerCase();
+    return this.platforms.filter(({ name }) => {
+      return name.toLowerCase().includes(filterValue);
+    });
+  }
+
+  private _autocompleteObjectValidator(): ValidatorFn {
+    return (control: AbstractControl): { [key: string]: any } | null => {
```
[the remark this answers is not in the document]
KK4ito: done

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +103,34 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+  }
+
+  displayFn(platform: Platform) {
+    return platform?.name ? platform.name : '';
+  }
+
+  private _filter(value: string): Platform[] {
+    const filterValue = value.toLowerCase();
+    return this.platforms.filter(({ name }) => {
+      return name.toLowerCase().includes(filterValue);
```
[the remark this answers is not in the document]
KK4ito: done

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
```diff
@@ -79,4 +101,36 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
...
+        return name === inputValue;
+      });
+      if (matchingEntry) {
+        this.accountForm.controls['platformId'].setValue(matchingEntry);
+      }
+    }
+  }
+
+  public displayFn(platform: Platform) {
+    return platform?.name ? platform.name : '';
```
dtslvr, line 118: `return platform?.name ?? '';`

dtslvr (changes requested):
I have simplified the platforms (type) a bit. Please fetch my update.

Unfortunately, I noticed one more problem: When I clear the platform, the form component gets red and the form is invalid. Do you have an idea?

KK4ito:
> I have simplified the platforms (type) a bit. Please fetch my update.
> 
> Unfortunately, I noticed one more problem: When I clear the platform, the form component gets red and the form is invalid. Do you have an idea?

Thanks for testing and feedback. This error happens because of the autocompelteObjectValidator. Fixed it with 473d466656e436977cd21241a2856bd03912ee5d

dtslvr (approved):
Had to fix an issue with removing a platform when you edit an account. Now I think it works fine. Thanks a lot for your contribution, Kevin! 😎

Issue #2442, opened 2023-10-07
[BUG] Endpoint api/v1/portfolio/positions is requested twice when loading Holdings in Overview


dtslvr:
**Bug Description**

The API endpoint `GET api/v1/portfolio/positions` is requested twice when loading the holdings tab from the overview.

**To Reproduce**

1. Login to Ghostfolio
2. Select _Holdings_
3. Check _Network_ in your browser's web developer tools

**Expected behavior**

`GET api/v1/portfolio/positions` is only requested once.

**Environment**

- Cloud and Self-hosted
- Ghostfolio Version 2.8.0

Yasir761:
I want to give it a try, Please do assign me

Yasir761:
Hey, @dtslvr I give it a try, but I didn't able to solve it, My bad.

Basimohd:
@dtslvr Could you please assign this to me?

Fullstack-Arshad:
@dtslvr  just one hour back I saw Yasir761 removed this task assignment, So I simply started working on it. Now when I have fixed it, It's assigned to Basimohd. My bad, I should have asked first to assign it to me 😕
But the solution is simple only

https://github.com/ghostfolio/ghostfolio/assets/61023134/197be73d-41b4-4d8b-ab69-24dea1ebb21f

Basimohd:
@dtlsr, if the commit by @Fullstack-Arshad is correct, kindly assign him the issue; otherwise, I will check it out by setting up.

Fullstack-Arshad:
Hi @Basimohd, I am so sorry I committed and pushed that code just as a backup so that I can start working on other bug, I did not mean to take this task from you. But thanks for allowing me I have added a PR @dtslvr

PR #2453, opened 2023-10-08
Fix:#2442 positions is requested twice when loading Holdings

Fullstack-Arshad:
Fix: #2442

### DIFF
--- a/apps/client/src/app/components/home-holdings/home-holdings.component.ts
+++ b/apps/client/src/app/components/home-holdings/home-holdings.component.ts
@@ -66,8 +66,6 @@ export class HomeHoldingsComponent implements OnDestroy, OnInit {
             this.user.permissions,
             permissions.createOrder
           );
-
-          this.update();
         }
       });
   }
--- a/apps/client/src/app/components/home-holdings/home-holdings.component.ts
+++ b/apps/client/src/app/components/home-holdings/home-holdings.component.ts
@@ -66,6 +66,8 @@ export class HomeHoldingsComponent implements OnDestroy, OnInit {
             this.user.permissions,
             permissions.createOrder
           );
+
+          this.update();
         }
       });
   }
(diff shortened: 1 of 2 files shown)


Fullstack-Arshad:
what I found was update function was making that API call and it was being called once from the constructor and once from the ngOninit, removing the removing the API call withing ngOnint was crashing the API call, hence I decided to remove the one in the constructor...

dtslvr:
Thanks for your investigations @Fullstack-Arshad.

I have tested it, but it does not seem to work like this: When you click the `gf-toggle` component (e.g. `Today` or `Max`) now, there is no further request to reload the holdings with the date range parameter. This happened before because the user settings were changed with the click.

apps/client/src/app/components/home-holdings/home-holdings.component.ts
```diff
@@ -66,8 +66,6 @@ export class HomeHoldingsComponent implements OnDestroy, OnInit {
             this.user.permissions,
             permissions.createOrder
           );
-
-          this.update();
```
dtslvr, line 70: Without it, holdings are not reloaded when you change the date range.

dtslvr (changes requested):

apps/client/src/app/components/home-holdings/home-holdings.component.ts
Fullstack-Arshad: Thanks for this information @dtslvr, I will further dig into this problem and try to resolve it. But I work in a company and this week I will be busy, therefore will try to do it in my free time.

Fullstack-Arshad: Hi @dtslvr, hope you are doing well, I have added a commit, this should hopefully fix the issue, attaching a video for your reference

https://github.com/ghostfolio/ghostfolio/assets/61023134/4d0774be-065a-4c94-83eb-cdfd292a4425

please let me know if there is still any issue....
Thanks 🙂

Fullstack-Arshad:
Hi @dtslvr , yesterday I added one commit that should hopefully fix this bug. Please check once and let me know if anything else is required, attaching a video clip for your reference

https://github.com/ghostfolio/ghostfolio/assets/61023134/dcd1eedc-f19a-4a59-8383-5691e21867fc

Thank you 🙂

dtslvr (approved):
Great that you solved the problem, @Fullstack-Arshad! 🙌🏻

PR #2455, opened 2023-10-09
feat: Add endpoint to transfer cash between accounts

### LINKED ISSUE #2446: Add endpoint to transfer cash between accounts
This is a follow up task of #2432.

### Todo

- [ ] Create endpoint `POST api/v1/account/transfer-balance` using `TransferBalanceDto`
- [ ] Add logic which calls [updateAccountBalance()](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/account/account.service.ts#L217) for both affected accounts 
- [ ] Enable transfer button [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/accounts-table/accounts-table.component.html#L1)
- [ ] Call endpoint on submit from client

Dhoni77:
Closes #2446

- Enabled transfer button on UI
- Created transfer api on client
- Implemented post transfer balance on submit
- Created the server side end point

### DIFF
--- a/apps/api/src/app/account/account.controller.ts
+++ b/apps/api/src/app/account/account.controller.ts
@@ -40,7 +40,7 @@ export class AccountController {
     private readonly impersonationService: ImpersonationService,
     private readonly portfolioService: PortfolioService,
     @Inject(REQUEST) private readonly request: RequestWithUser
-  ) { }
+  ) {}
 
   @Delete(':id')
   @UseGuards(AuthGuard('jwt'))
@@ -158,11 +158,9 @@ export class AccountController {
 
   @Post('transfer-balance')
   @UseGuards(AuthGuard('jwt'))
-  public async transferAccountBalance(@Body() {
-    accountIdFrom,
-    accountIdTo,
-    balance
-  }: TransferBalanceDto) {
+  public async transferAccountBalance(
+    @Body() { accountIdFrom, accountIdTo, balance }: TransferBalanceDto
+  ) {
     if (
       !hasPermission(this.request.user.permissions, permissions.updateAccount)
     ) {
@@ -191,7 +189,6 @@ export class AccountController {
       userId: this.request.user.id
     });
   }
-
 
   @Put(':id')
   @UseGuards(AuthGuard('jwt'))
--- a/apps/api/src/app/account/account.controller.ts
+++ b/apps/api/src/app/account/account.controller.ts
@@ -170,12 +170,27 @@ export class AccountController {
       );
     }
 
+    const currentAccountIds = (
+      await this.accountService.getAccounts(this.request.user.id)
+    ).map((account) => account.id);
+
+    if (
+      ![accountIdFrom, accountIdTo].every((id) =>
+        currentAccountIds.includes(id)
+      )
+    ) {
+      throw new HttpException(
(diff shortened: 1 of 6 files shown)


Dhoni77:
cc @dtslvr 
Hi, I just had a doubt regarding the updateAccountBalance() logic for updating the account balance.
Since we are not reducing the amount from the fromAccount, instead we are adding the entered amount to both the from and to accounts.
Is this expected or should I make any logic changes ?

apps/client/src/app/services/data.service.ts
```diff
@@ -471,6 +472,18 @@ export class DataService {
     return this.http.post<OrderModel>(`/api/v1/account`, aAccount);
   }
 
+  public postTransferAccountBalance({
```
dtslvr, line 475: Can you move down ⬇️ to keep the alphabetical order?

apps/client/src/app/pages/accounts/accounts-page.component.ts
```diff
@@ -298,9 +298,18 @@ export class AccountsPageComponent implements OnDestroy, OnInit {
           console.log(
```
dtslvr, line 298: Please remove this

apps/api/src/app/account/account.controller.ts
```diff
@@ -154,6 +156,40 @@ export class AccountController {
...
+        StatusCodes.FORBIDDEN
+      );
+    }
+
+    const today = resetHours(new Date());
+    const userCurrency = this.request.user.Settings.settings.baseCurrency;
+
+    await this.accountService.updateAccountBalance({
+      accountId: accountIdFrom,
+      amount: balance,
```
dtslvr, line 178: > Is this expected or should I make any logic changes?

Adding the negative balance here should work:

```
amount: -balance,
```

apps/api/src/app/account/account.controller.ts
```diff
@@ -154,6 +156,40 @@ export class AccountController {
...
+  ) {
+    if (
+      !hasPermission(this.request.user.permissions, permissions.updateAccount)
+    ) {
+      throw new HttpException(
+        getReasonPhrase(StatusCodes.FORBIDDEN),
+        StatusCodes.FORBIDDEN
+      );
+    }
+
```
dtslvr, line 172: For security reasons, a check is needed that the accounts with id `accountIdFrom` and `accountIdTo` belong to the user.

Can you get the accounts like this:
```
this.accountService.getAccounts(this.request.user.id);
```
And throw a new `NOT_FOUND` exception if the check fails?

dtslvr (changes requested):
Thanks for your contribution, @Dhoni77 🙂 I have added some comments.

apps/client/src/app/pages/accounts/accounts-page.component.ts
Dhoni77: done

apps/api/src/app/account/account.controller.ts
Dhoni77: yes

dtslvr (approved):
Thank you for the modifications. I have slightly refactored the code, ready to merge now.

PR #2487, opened 2023-10-14
feat: Add reusable currency selector

Dhoni77:
- Added reusable currency selector component
- Use currency selector component in create or account dialog component

### DIFF
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.component.ts
@@ -20,6 +20,7 @@ import { Observable, Subject } from 'rxjs';
 import { map, startWith } from 'rxjs/operators';
 
 import { CreateOrUpdateAccountDialogParams } from './interfaces/interfaces';
+import { Currency } from '@ghostfolio/common/interfaces/currency.interface';
 
 @Component({
   host: { class: 'h-100' },
@@ -30,7 +31,7 @@ import { CreateOrUpdateAccountDialogParams } from './interfaces/interfaces';
 })
 export class CreateOrUpdateAccountDialog implements OnDestroy {
   public accountForm: FormGroup;
-  public currencies: string[] = [];
+  public currencies: Currency[] = [];
   public filteredPlatforms: Observable<Platform[]>;
   public platforms: Platform[];
 
@@ -46,7 +47,10 @@ export class CreateOrUpdateAccountDialog implements OnDestroy {
   public ngOnInit() {
     const { currencies, platforms } = this.dataService.fetchInfo();
 
-    this.currencies = currencies;
+    this.currencies = currencies.map((currency) => ({
+      label: currency,
+      value: currency
+    }));
     this.platforms = platforms;
 
     this.accountForm = this.formBuilder.group({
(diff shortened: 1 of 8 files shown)


dtslvr (changes requested):
Thanks for your contribution, @Dhoni77.

I have slightly refactored the code and found out that the following is not yet working as expected:

* Set `disabled: true` in `accountForm`does not disable the currency selector
* When you edit an account, the current currency is not preselected

Can you please have a look?

Dhoni77:
> Thanks for your contribution, @Dhoni77.
> 
> I have slightly refactored the code and found out that the following is not yet working as expected:
> 
> * Set `disabled: true` in `accountForm`does not disable the currency selector
> * When you edit an account, the current currency is not preselected
> 
> Can you please have a look?

Sure

Dhoni77:
> > Thanks for your contribution, @Dhoni77.
> > I have slightly refactored the code and found out that the following is not yet working as expected:
> > 
> > * Set `disabled: true` in `accountForm`does not disable the currency selector
> > * When you edit an account, the current currency is not preselected
> > 
> > Can you please have a look?
> 
> Sure

Hi @dtslvr,

I tried adding the disabled property to the **accountForm** group and the form control is disabled.

Attachment:

disabled

Code:

```typescript
    this.accountForm = this.formBuilder.group({
      accountId: [{ disabled: true, value: this.data.account.id }],
      balance: [this.data.account.balance, Validators.required],
      comment: [this.data.account.comment],
      currency: [{ value: this.data.account.currency, disabled: true }, Validators.required],
      isExcluded: [this.data.account.isExcluded],
      name: [this.data.account.name, Validators.required],
      platformId: [
        this.platforms.find(({ id }) => {
          return id === this.data.account.platformId;
        }),
        this.autocompleteObjectValidator()
      ]
    });
```

dtslvr:
> I tried adding the disabled property to the accountForm group and the form control is disabled.

Thanks for checking, maybe I did something wrong in my configuration.

dtslvr:
> * When you edit an account, the current currency is not preselected

What about this one @Dhoni77?

Dhoni77:
> > * When you edit an account, the current currency is not preselected
> 
> What about this one @Dhoni77?

Will check this @dtslvr

Dhoni77:
cc @dtslvr

dtslvr:
@Dhoni77 I have found a way to directly access the form control. Then, there is no `defaultValue` needed.

But I have also found another weird problem:

1. Open create account dialog
2. Delete currency (that the form field gets invalid)
3. Reselect (default) currency
4. There is the following error:
```
ERROR TypeError: Cannot read properties of null (reading 'toLowerCase')
    at currency-selector.component.ts:153:29
```

It looks like `value` somehow gets removed from the entry of `this.currencies` 😜

Dhoni77:
> @Dhoni77 I have found a way to directly access the form control. Then, there is no `defaultValue` needed.
> 
> But I have also found another weird problem:
> 
> 1. Open create account dialog
> 2. Delete currency (that the form field gets invalid)
> 3. Reselect (default) currency
> 4. There is the following error:
> 
> ```
> ERROR TypeError: Cannot read properties of null (reading 'toLowerCase')
>     at currency-selector.component.ts:153:29
> ```
> 
> It looks like `value` somehow gets removed from the entry of `this.currencies` 😜

Will look into it 👍

Dhoni77:
cc @dtslvr

dtslvr (approved):
Awesome that you found a solution, @Dhoni77! 🙌🏻

PR #2496, opened 2023-10-15
Csv importing enhancement

### LINKED ISSUE #2375: [BUG] Importing Sell Activities with Negative Quantity will increase the number of Shares Purchased
**Bug Description**

When importing a CSV file with Activities, if you define the Activity as Sell, but keep the Quantity negative, it'll sum as a Purchase (it'll add instead of subtract)

**To Reproduce**

1. Import a CSV file with a Sell activity and a Negative quantity
2. Check your holds and you'll see that now you have this quantity as purchased instead of sell

**Expected behavior**

When importing or creating an Activity of the Sell type, no matter if the Quantity is negative in the input, it should be considered as a Sell of that Quantity not a Purchase

**Screenshots**

image
image

**Environment**

- Self-hosted
- Ghostfolio Version 2.4.0
- Chrome
- Linux

**Additional Context**
Probably, the easiest solution would be to implement a validation when importing the CSV file, but maybe it should also treat this in the calculation?

I can help on the solution of this issue, but probably we should discuss this first?

rafacla:
Proposed changes to close #2375: unitPrice and quantity should be greater than zero, and client will make it absolute when importing a csv file.


dtslvr:
Hi @rafacla

Should this still be a draft or ready for review?

rafacla:
> Hi @rafacla
> 
> Should this still be a draft or ready for review?

Hi @dtslvr it's ready!

dtslvr (approved):
Thank you for this improvement, @rafacla! Looks fine 👍🏻

apps/client/src/app/services/import-activities.service.ts
dtslvr, line 1: ℹ️  `parseFee()` already uses `Math.abs()`

PR #2503, opened 2023-10-17
feat: Implement i18n service to query XML files

### LINKED ISSUE #2494: Implement i18n service to query XML files
This issue will enable [Angular Internationalization](https://angular.io/guide/i18n-overview) also in the backend.

Create a new `I18nService` in `apps/api/src` (with a module) which loads all `XML` locale files from `dist/apps/api/assets/locales/messages.*.xlf` in the `constructor` and keep them in the service at runtime.

The service should provide the following function:
```
public getTranslation({ id, locale }: { id: string; locale: string }): string {
  ...
}
```
e.g.
```
i18nService.getTranslation({ id: 'e19fcf996343543e13789f17b550ab0c08124b5c', locale: 'de' });
```
should return `Registrieren` (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.de.xlf#L6)).

For the query, you should use the library [cheerio](https://www.npmjs.com/package/cheerio) which is already used in the project.

Dhoni77:
Closes #2494

- Implemented I18nService
- Load files from dist/apps/api/assets/locales/messages.*.xlf
- Select translation based on given query

### DIFF
--- a/apps/api/src/services/i18n/i18n.service.ts
+++ b/apps/api/src/services/i18n/i18n.service.ts
@@ -26,7 +26,11 @@ export class I18nService {
       Logger.warn(`Translation not found for locale '${locale}'`);
     }
 
-    const translatedText = $(`trans-unit[id="${id}"] > target`).text();
+    const translatedText = $(
+      `trans-unit[id="${id}"] > ${
+        locale === DEFAULT_LANGUAGE_CODE ? 'source' : 'target'
+      }`
+    ).text();
 
     if (!translatedText) {
       Logger.warn(`Translation not found for id '${id}' in locale '${locale}'`);
--- a/apps/api/src/services/i18n/i18n.service.ts
+++ b/apps/api/src/services/i18n/i18n.service.ts
@@ -1,0 +1,59 @@
+import * as fs from 'fs';
+import { join } from 'path';
+import { Injectable } from '@nestjs/common';
+import * as cheerio from 'cheerio';
+
+@Injectable()
+export class I18nService {
+  private localesPath = join(__dirname, 'assets/locales/');
+  private localeRegex = /^messages\.[a-z]{2}\.xlf$/;
+  private translations: { [locale: string]: cheerio.CheerioAPI } = {};
+
+  constructor() {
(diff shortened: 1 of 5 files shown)


apps/api/src/services/i18n/i18n.service.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (!this.localeRegex.test(file)) continue;
+        if (!fs.existsSync(`${this.localesPath}${file}`)) {
+          throw new Error(`File: ${file} not found`);
+        } else {
+          const xmlData = fs.readFileSync(`${this.localesPath}${file}`, 'utf8');
+          this.translations[file.split('.')[1]] = this.parseXml(xmlData);
+        }
+      }
+    } catch (error) {
+      console.log(error);
```
dtslvr, line 51: `Logger.error(error, 'I18nService');`

apps/api/src/services/i18n/i18n.service.ts
```diff
@@ -0,0 +1,59 @@
+import * as fs from 'fs';
```
dtslvr, line 1: `import { readdirSync } from 'fs';`

apps/api/src/services/i18n/i18n.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          this.translations[file.split('.')[1]] = this.parseXml(xmlData);
+        }
+      }
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
+  private parseXml(xmlData: string): cheerio.CheerioAPI {
+    const selectorFn = cheerio.load(xmlData, { xmlMode: true });
```
dtslvr, line 56: `return cheerio.load(xmlData, { xmlMode: true });`

dtslvr (changes requested):
Thanks for the implementation, @Dhoni77. I have added some feedback.

dtslvr (approved):
Thank you @Dhoni77 😎

I have slightly adapted the service and integrated it in the _HTML template middleware_.

PR #2541, opened 2023-10-25
fix/issue#2538: Fixed the analysis page error when deleting benchmark

Basimohd:
@dtlsr I've identified the root cause of the issue. When a user selects a benchmark from the "compare-with" dropdown, it triggers an update to user.setting.benchmark both in the frontend and backend. Consequently, I examined whether there is dataSource available for the selected benchmark in the updateBenchmarkDataItems() function.


dtslvr (approved):
Thanks for this fix, @Basimohd! I have slightly refactored the code. Now ready to merge!

Issue #2550, opened 2023-10-27
Improve Transfer Cash Balance


dtslvr:
There are some tasks planned to further improve the [TransferBalanceDialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/pages/accounts/transfer-balance) (open via `http://localhost:4200/en/accounts`).

- [x] Frontend: Remove height (`80vh`)
- [x] Frontend: Remove default value (`0`) of `balance`
- [x] Frontend: Treat form as invalid if `accountFrom` is equal to `accountTo`
- [x] Backend: Throw a `400 Bad Request` error if `accountFrom` is equal to `accountTo` in `api/v1/account/transfer-balance`
- [x] Backend: Throw a `400 Bad Request` error if the current balance of `accountFrom` is lower than the `balance` from the payload

iajaymk:
@dtslvr Can you please assign this to me?

PR #2552, opened 2023-10-27
bug fix: #2550

iajaymk:
#2550 
* Removed default value of balance
* Removed height
* Added validator to check accountFrom and accountTo are equal 
* Made 400 Bad request error if  accountFrom and accountTo are equal 
* Made 400 Bad request error if accountFrom balance is less than the transfer amount


dtslvr (approved):
Great contribution, @iajaymk! 🙌🏻 
I have made some minor refactorings... Ready to merge now.

PR #2557, opened 2023-10-28
feat: add decimation plugin to chart

### LINKED ISSUE #2553: Setup data decimation option in line chart component
The line chart looks a bit frazzled, especially rendered on small screens with many data points.

128614047-02585aca-8c36-49ea-81bf-14d1c7622421

Add `@Input() dataDecimation = false;` to the [LineChartComponent](https://github.com/ghostfolio/ghostfolio/blob/main/libs/ui/src/lib/line-chart) to enable / disable data decimation: the number of data points controlled via [Data Decimation plugin](https://www.chartjs.org/docs/latest/configuration/decimation.html) of chart.js. Preferred are the default options.

Dhoni77:
Closes #2553

- Setup data decimation plugin to chart with default option

### DIFF
--- a/libs/ui/src/lib/line-chart/line-chart.component.ts
+++ b/libs/ui/src/lib/line-chart/line-chart.component.ts
@@ -49,6 +49,7 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
   @Input() benchmarkLabel = '';
   @Input() colorScheme: ColorScheme;
   @Input() currency: string;
+  @Input() dataDecimation = false;
   @Input() historicalDataItems: LineChartItem[];
   @Input() isAnimated = false;
   @Input() locale: string;
@@ -209,6 +210,10 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
               tooltip: this.getTooltipPluginConfiguration(),
               verticalHoverLine: {
                 color: `rgba(${getTextColor(this.colorScheme)}, 0.1)`
+              },
+              decimation: {
+                enabled: this.dataDecimation,
+                algorithm: 'min-max'
               }
             },
             scales: {


libs/ui/src/lib/line-chart/line-chart.component.ts
```diff
@@ -209,6 +210,10 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
               tooltip: this.getTooltipPluginConfiguration(),
               verticalHoverLine: {
                 color: `rgba(${getTextColor(this.colorScheme)}, 0.1)`
+              },
+              decimation: {
```
dtslvr, line 214: Can you please move ⬆️  to keep the alphabetical order?

libs/ui/src/lib/line-chart/line-chart.component.ts
```diff
@@ -209,6 +210,10 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
...
+                enabled: this.dataDecimation,
+                algorithm: 'min-max'
```
dtslvr, line 216: According to the [documentation](https://www.chartjs.org/docs/latest/configuration/decimation.html), `min-max` is the default value, so not needed at all.

dtslvr (changes requested):
Thanks @Dhoni77 😎 I have added very few comments.

Can you enable the option (`[dataDecimation]="true"`) in
* `AdminMarketDataDetailComponent`
* `HomeOverviewComponent`
* `PositionDetailDialog`

Dhoni77:
cc @dtslvr

dtslvr (changes requested):
Thanks a lot for the update, @Dhoni77. The code looks fine, unfortunately when I test it, I can't see any difference:

data-decimation

The first chart is configured with `[dataDecimation]="false"` and the second one with `[dataDecimation]="true"`.

Ps: I have registered the plugin, but it also has no effect 🙁

Dhoni77:
> Thanks a lot for the update, @Dhoni77. The code looks fine, unfortunately when I test it, I can't see any difference:
> 
> data-decimation
> The first chart is configured with `[dataDecimation]="false"` and the second one with `[dataDecimation]="true"`.
> 
> Ps: I have registered the plugin, but it also has no effect 🙁

Will look into it @dtslvr

dtslvr:
> Will look into it @dtslvr

Were you able to find out anything, @Dhoni77?

Dhoni77:
> > Will look into it @dtslvr
> 
> Were you able to find out anything, @Dhoni77?

Hi @dtslvr Unfortunately I couldn't find a solution to this

dtslvr:
> Hi @dtslvr Unfortunately I couldn't find a solution to this

No problem. Thanks for the update.

karter16:
Just an FYI that the documentation on this states that the decimation will take effect when the threshold is exceeded. The Threshold by default is set to 4x the width of the canvas. @dtslvr I don't believe that there are enough data points between 2016 and 2023 in your example to exceed the Threshold.

Screenshot 2024-02-20 at 5 50 18 PM



For anyone picking this up in the future I'd suggest setting the Threshold to a lower value.

PR #2586, opened 2023-11-04
Add major version to docker tags

### LINKED ISSUE #2585: Add docker image major version tag
Hi there 👋

I'm hoping you'll consider tagging your image with a major version tag alongside the full version tag during release.

This is useful for those of us that use tools like watchtower or similar to keep our containers up to date on a schedule, but don't want to risk pulling in a new major version that may break the deployment.

So for example you'd tag a release with:

'latest', '2.30.0', '2'

ianschmitz:
Fixes #2585


dtslvr (approved):
Thank you for the contribution @ianschmitz! 😎

PR #2595, opened 2023-11-04
fix: CoinGecko GetQuotes

lmoe:
Found Ghostfolio today and wanted to try it out. 

I have noticed that I couldn't add crypto activities as `/api/v1/symbol/COINGECKO/<coin>` is always returning 404s. (The const `response` gets redeclared once the API call is made, which makes the function error out.) 

This should fix it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed an issue to get quotes in the _CoinGecko_ service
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thank you very much for providing this fix, @lmoe.

Looks like my [previous refactoring](https://github.com/dtslvr/ghostfolio/commit/a7043787025f8d831851227139715148d4894658#diff-6276162dd7ad841cc339e8711e1ec42e4fb65cf941a7f27653d5860766025d01) was not successful 😮‍💨

pvalfar:
I'm still getting this error

PR #2622, opened 2023-11-08
Feature/remove loading indicator of unit in overview of home page

dtslvr:


apps/client/src/app/components/home-overview/home-overview.component.ts
```diff
@@ -69,6 +70,10 @@ export class HomeOverviewComponent implements OnDestroy, OnInit {
       .subscribe((impersonationId) => {
         this.hasImpersonationId = !!impersonationId;
 
+        this.unit = this.hasImpersonationId
+          ? '%'
+          : this.user?.settings?.baseCurrency;
```
slagiewka, line 75: I'm not sure what the impersonation is in this lingo, but in 2.20 with zen mode I get:
image


Which is not true - the value is in %.

dtslvr: Thank you @slagiewka. I have created a bug report #2624.

PR #2704, opened 2023-12-01
Bugfix/fix too strict import deduplication logic

### LINKED ISSUE #2438: API import missing transactions
**Bug Description**

I try to write an importer using the import API for one trading platform that I am using. My first issue was that I don't have all transaction after running the script. After some analyzes I found that I have some trades on the same day, with same quantity and same market price. Only some minutes later than the first trade.

**To Reproduce**

1. having a list of trades, some trades having the same date (different time), same quantity, same market price
2. convert list to json
3. call import API

**Expected behavior**

Having the same number of transaction in ghostfolio as in the list.
Having the API reporting a duplicate transaction error. Instead the activities object in the response is empty for those imports.

**Screenshots**
-

**Logs**
-

**Environment**

- Self-hosted
- Ghostfolio Version 2.8.0

**Additional context**

In `ghostfolio/apps/api/src/app/import/import.service.ts` I found `isSameDay(activity.date, date) &&`. Is there some way to change the coding in my instance  that it is using the timestamp with time or to include the comment in the duplicate detection? I put the transaction id from my trading platform into the comment.

Thanks for your help :-)

ms32035:
Fixes #2438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Respected the `withExcludedAccounts` flag in the account balance time series
 
+### Changed
+
+- Relaxed the check for duplicates in the preview step of the activities import (allow same day)
+
 ## 2.27.1 - 2023-11-28
 
 ### Changed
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thank you for the contribution, @ms32035!

PR #2771, opened 2023-12-21
Feature/add permissions tests for controllers

underwater:
Updated the controllers to make use of the HasPermission guard and decorator instead of explicit verification of permissions.  each time.   There were some edge cases, which I've left alone in this initial p/r

- the case DEMO role in updateUserSettings user.controller.ts
- few cases where business rules are also being verified (as per our discussion)
- few cases where multiple permissions are being checked (example: ImportController)  

i suggest to deal with these in a seperate P/R.

### DIFF
--- a/apps/api/src/app/account-balance/account-balance.controller.ts
+++ b/apps/api/src/app/account-balance/account-balance.controller.ts
@@ -1,3 +1,5 @@
+import { HasPermission } from '@ghostfolio/api/decorators/has-permission.decorator';
+import { HasPermissionGuard } from '@ghostfolio/api/guards/has-permission.guard';
 import { permissions } from '@ghostfolio/common/permissions';
 import type { RequestWithUser } from '@ghostfolio/common/types';
 import {
@@ -10,11 +12,10 @@ import {
 } from '@nestjs/common';
 import { REQUEST } from '@nestjs/core';
 import { AuthGuard } from '@nestjs/passport';
-import { StatusCodes, getReasonPhrase } from 'http-status-codes';
 import { AccountBalance } from '@prisma/client';
-import { HasPermission } from '@ghostfolio/api/decorators/has-permission.decorator';
+import { StatusCodes, getReasonPhrase } from 'http-status-codes';
+
 import { AccountBalanceService } from './account-balance.service';
-import { HasPermissionGuard } from '@ghostfolio/api/guards/has-permission.guard';
 
 @Controller('account-balance')
 export class AccountBalanceController {
@@ -23,9 +24,9 @@ export class AccountBalanceController {
     @Inject(REQUEST) private readonly request: RequestWithUser
   ) {}
 
+  @HasPermission(permissions.deleteAccountBalance)
   @Delete(':id')
   @UseGuards(AuthGuard('jwt'), HasPermissionGuard)
-  @HasPermission(permissions.deleteAccountBalance)
   public async deleteAccountBalance(
     @Param('id') id: string
   ): Promise<AccountBalance> {
--- a/apps/api/src/app/account-balance/account-balance.controller.ts
+++ b/apps/api/src/app/account-balance/account-balance.controller.ts
@@ -14,6 +14,7 @@ import { StatusCodes, getReasonPhrase } from 'http-status-codes';
 import { AccountBalance } from '@prisma/client';
 import { HasPermission } from '@ghostfolio/api/decorators/has-permission.decorator';
 import { AccountBalanceService } from './account-balance.service';
+import { HasPermissionGuard } from '@ghostfolio/api/guards/has-permission.guard';
 
 @Controller('account-balance')
 export class AccountBalanceController {
@@ -23,7 +24,7 @@ export class AccountBalanceController {
   ) {}
 
   @Delete(':id')
-  @UseGuards(AuthGuard('jwt'))
+  @UseGuards(AuthGuard('jwt'), HasPermissionGuard)
   @HasPermission(permissions.deleteAccountBalance)
   public async deleteAccountBalance(
     @Param('id') id: string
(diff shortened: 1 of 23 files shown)


apps/api/src/app/account-balance/account-balance.controller.ts
```diff
@@ -10,8 +10,10 @@ import {
 } from '@nestjs/common';
 import { REQUEST } from '@nestjs/core';
 import { AuthGuard } from '@nestjs/passport';
-import { AccountBalance } from '@prisma/client';
+import { permissions } from '@ghostfolio/common/permissions';
```
dtslvr, line 13: This import seems to be a duplicate

apps/api/src/app/cache/cache.controller.ts
```diff
@@ -21,19 +15,8 @@ export class CacheController {
 
   @Post('flush')
   @UseGuards(AuthGuard('jwt'))
+  @HasPermission(permissions.accessAdminControl) // permission needed
```
dtslvr, line 18: Why this comment?

apps/api/src/app/import/import.controller.ts
```diff
@@ -35,17 +36,14 @@ export class ImportController {
 
   @Post()
   @UseGuards(AuthGuard('jwt'))
+  @HasPermission(permissions.createAccount)
```
dtslvr, line 39: I would change this to `permissions.createOrder`

apps/api/src/app/import/import.controller.ts
```diff
@@ -35,17 +36,14 @@ export class ImportController {
...
   public async import(
     @Body() importData: ImportDataDto,
     @Query('dryRun') isDryRun?: boolean
   ): Promise<ImportResponse> {
     if (
-      !hasPermission(
-        this.request.user.permissions,
-        permissions.createAccount
-      ) ||
       !hasPermission(this.request.user.permissions, permissions.createOrder)
```
dtslvr, line 47: I would change this to `permissions.createAccount`

dtslvr (changes requested):
Great how much boilerplate you can eliminate with that 👍🏻

I have added a few commens.

apps/api/src/app/cache/cache.controller.ts
underwater: forgot to remote it :-)

apps/api/src/app/account-balance/account-balance.controller.ts
underwater: agree

apps/api/src/app/import/import.controller.ts
underwater: OK, but then will do the excplicit check on the createAccount (as mentioned previously, I didn't test yet the permission guard / decorator on cases that check multiple permissions. (leaving that for seperate P/R).  OK with you ?

apps/api/src/app/import/import.controller.ts
underwater: OK, got it :-)

underwater:
Fixed to comply with suggestsin....
Also revisited few controllers, and removed unused constructor injection of RequestWithUser and its dependencies where this was no longer needed. So In controllers like account-balance, queue, benchmark, the "@Inject(REQUEST) private readonly request: RequestWithUser" is no longer needed.

dtslvr (changes requested):
Thank you for the changes, @underwater.
I've tried it, however I always get `403 Forbidden` when I open pages of the _Admin Control_ panel. Other endpoints (non-admin) seems to work, though. Any idea?

underwater:
Let me check this tomorrow and revert

dtslvr:
> Let me check this tomorrow and revert

Thanks. Do not revert if possible 🙂

underwater:
:-) I meant, check and get back to you :-)

underwater:
I suspect to have found the issue

I was registering HasPermissionGuard globally which means it runs earlier in the pipeline than the AuthGuard which is being applied explicitly to each controoler endpoint ... 

As a result HasPermissionGuard was receiving the request before the AuthGuard has had a chance to populate the 'user' object and the HasPermissionGuard was seeing 'user' object as 'undefined'.

To fix this, I would suggest I skip the global registration of the HasPermissionGuard (which I was doing in the app.module.ts) and instead just include it in our @UseGuards(...) like this @UseGuards(AuthGuard('jwt'), HasPermissionGuard)


So to summarize,
1) controller endpoints that do check permissions and indicate the required permissions using @HasPermissionDecorator will need to use @UseGuards(AuthGuard('jwt'), HasPermissionGuard)

while 

2) controller endpoints that don't check permissions (like getPlatforms()) will continue to use the simple
@UseGuards(AuthGuard('jwt'))


To keep things consistent and to avoid confusion, we could also always use the 
@UseGuards(AuthGuard('jwt'), HasPermissionGuard) 

This would work for all cases, the HasPermissionsGuard defaults to allowing the request to pass when no required permission is specified using the @HasPermission(....) 

let me know what is preferred and I will do the chance and push for you to have a look.

underwater:
We could eventually also consider registering both guards (AuthGuard, and HasPermissionGuard) globally and in the right order, to avoid having to think about this at each endpoint

dtslvr:
Using
```
@UseGuards(AuthGuard('jwt'), HasPermissionGuard)
```
is also fine. To ensure that it is forgotten less often, it may make sense to always use both of them.

underwater:
sure, I can update my p/r to make use of @UseGuards(AuthGuard('jwt'), HasPermissionGuard) everywhere for consistency...

underwater:
> Using
> 
> ```
> @UseGuards(AuthGuard('jwt'), HasPermissionGuard)
> ```
> 
> is also fine. To ensure that it is forgotten less often, it may make sense to always use both of them.

done ! pushed update

dtslvr (approved):
Looks fine now. Thanks @underwater 🙂

PR #2546, opened 2023-10-25
Provided support for Scraper Configuration.

M27afk:
With reference to https://github.com/ghostfolio/ghostfolio/issues/2292
@dtslvr 
Screenshot 2023-10-26 011351

Issue #2292, opened 2023-09-03
Add ability to test scraper configuration


dtslvr:
### Outline

1. Add a button to the [asset profile dialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog) to test a scraper configuration
2. Make a request to `/api/v1/gather/MANUAL/:symbol/:dateString` (where `dateString` is today)
3. Show the result (`marketPrice`) as an alert

### Mockup

<img width="795" alt="scraper-configuration" src="https://github.com/ghostfolio/ghostfolio/assets/4159106/0ad7e9f3-483b-4147-8b16-356c01ccefe3">

Siddharth9890:
i can work on this @dtslvr please?

dtslvr:
Any progress to share on this @Siddharth9890?

Siddharth9890:
Sorry @dtslvr currently i have exams so i am busy with it you can assign it to someone else if needed. Thank you !

M27afk:
Can I get assigned on this? @dtslvr

dtslvr:
> 1. The Scraper configuration is available only for Manual assets right?
> 2. Is @Post('gather/:dataSource/:symbol/:dateString') in admin.controller.ts the right route to be called?
> 3. How is the data in the scraper configuration input related to the API call?

1. Yes, scraper configuration is only used where `dataSource` is `MANUAL`
2. Yes
3. Behind the scenes, it calls `getHistorical()` with the relevant data provider ([ManualService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/data-provider/manual/manual.service.ts#L65) in this case) and uses the scraper configuration

M27afk:
Can I get sample Scraper Configuration of a Sample manual asset and expected output for testing purpose?

dtslvr:
> Can I get sample Scraper Configuration of a Sample manual asset and expected output for testing purpose?

Sure! This `scraperConfiguration` should return the most recent price:

```
{
  "url": "https://zkb-finance.mdgms.com/products/funds/zkb/detail.html?LANG=en&FI_ID_NOTATION=287639517",
  "selector": ".fi_table_module tr:nth-of-type(1) td:nth-of-type(3)"
}
```

[...]

PR #2777, opened 2023-12-23
Suggestion for fix to #2292

Hugo-Persson:
A suggested fix to https://github.com/ghostfolio/ghostfolio/issues/2292

It is built on contribution made by @M27afk in https://github.com/ghostfolio/ghostfolio/pull/2546.

I left the old code with dryRun but before merging in we should remove if my approach is ok.


.gitignore
```diff
@@ -40,3 +40,4 @@ yarn-error.log
 # System Files
 .DS_Store
 Thumbs.db
+.nx/cache
```
dtslvr, line 43: When you rebase this branch, you don't need this as it is already there.

apps/api/src/app/admin/admin.controller.ts
```diff
@@ -243,13 +247,30 @@ export class AdminController {
...
+
     return this.dataGatheringService.gatherSymbolForDate({
       dataSource,
       date,
+      isDryRun,
       symbol
     });
   }
 
+  @Post('test-scraper')
```
dtslvr, line 263: Instead of introducing a new endpoint, can you use the one specified in https://github.com/ghostfolio/ghostfolio/issues/2292.

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
```diff
@@ -240,6 +240,21 @@ export class AssetProfileDialog implements OnDestroy, OnInit {
       });
   }
 
+  public onTestScraper() {
+
+    this.adminService
+      .testScrapeConfig(this.assetProfileForm.controls['scraperConfiguration'].value)
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe((response) => {
+        if (response && response.price) {
```
dtslvr, line 249: `if (response.price) {`

dtslvr (changes requested):
Thanks for your suggestion. I have added a few comments.

apps/api/src/app/admin/admin.controller.ts
Hugo-Persson: How would I pass the scraper config? I think this is cleaner instead of bloating the other route. But do there exist an easy way to pass the configured scraper config to that route @dtslvr ?

dtslvr: Fair enough. I was not aware that the configuration may not have been saved yet.

Hugo-Persson: Ok I will clean up code and fix your comment tomorrow and the ping you

apps/client/src/app/services/admin.service.ts
```diff
@@ -175,16 +175,22 @@ export class AdminService {
...
   }: UniqueAsset & {
     date?: Date;
+  } & {
+    isDryRun?: boolean;
   }) {
     let url = `/api/v1/admin/gather/${dataSource}/${symbol}`;
 
     if (date) {
-      url = `${url}/${format(date, DATE_FORMAT)}`;
+      url = `${url}/${format(date, DATE_FORMAT)}?dryRun=true`;
```
dtslvr, line 184: Is this still needed?

.gitignore
dtslvr: Please remove this line (duplicate)

dtslvr (changes requested):
Can you please run `yarn format` that the build is running without errors?

apps/client/src/app/services/admin.service.ts
Hugo-Persson: Missed this, will remove

dtslvr:
Superseded by #2808.

PR #2789, opened 2023-12-28
Feature/add support to edit currency of asset profile with data source manual

dtslvr:


dtslvr:
Follow-up task: #2795

PR #2913, opened 2024-01-23
Adjusts Dockerfile to enable healthcheck introduced in 5fc84a0

### LINKED ISSUE #2912: [BUG] Healthcheck fails due to curl missing in container
**Bug Description**

The healthcheck introduced in `v2.43.0` depends on `curl` which is missing in the official container images. Thus, healthchecks are constantly failing and the container will never be considered `healthy` by Docker.

This is due to `curl` being installed in the `builder` step of the image's multi-stage build in [5fc84a0](https://github.com/ghostfolio/ghostfolio/commit/5fc84a06cc67f8b6f06291343b11d93f36ddd95c) instead of being installed in the final stage of the build. I will submit a PR with a fix.

**To Reproduce**

1. `docker compose -f docker/docker-compose.yml up -d `
2. `watch docker ps`
3. See the ghostfolio container never reaching `healthy` state
4. `docker exec <ghostfolio_container> curl`
5. See the execution of `curl` fail because it's not installed.

mocdaniel:
Fixes #2912

Testing the fix locally:

1. `docker build -t localtests/ghostfolio .`
2. Adjust the `image` in `docker/docker-compose.yml` to `localtests/ghostfolio`
3. `docker compose -f docker/docker-compose.yml --env-file .env.example up -d`
4. `watch docker ps`
5. Ghostfolio container should become healthy 10-20s after startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed the installation of `curl` in the `Dockerfile` so that the healthcheck introduced in `v2.43.0` works.
+
 ## 2.43.0 - 2024-01-23
 
 ### Added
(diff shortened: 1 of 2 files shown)


mocdaniel:
It'd be great if we could get a patch release once this is merged - right now, v2.43.0 is not usable in setups with e.g. reverse-proxies, as they most likely won't pick up on the unhealthy ghostfolio container.

dtslvr (approved):
Thank you for the quick fix @mocdaniel!

Issue #2901, opened 2024-01-21
Support filters in export endpoint


dtslvr:
Add support for filters (`filterByAccounts`, `filterByAssetClasses`, `filterByTags`) in `GET /api/v1/export` [endpoint](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/export/export.controller.ts#L19). The same functionality already exists for `getDetails()` (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/portfolio/portfolio.controller.ts#L82)).

Fullstack-Arshad:
Hi @dtslvr, I have worked on this issue, shall I raise PR?

dtslvr:
> shall I raise PR?

Sure 😎

Fullstack-Arshad:
PR Added, let me know if any there is any issue 😀

PR #2939, opened 2024-01-28
Added filter support in export endpoint

Fullstack-Arshad:
Feature  #2901: Support filters in export endpoint


apps/api/src/app/export/export.service.ts
dtslvr, line 1: You don't do anything with the `filters` yet 🤓 Can you add them here:

```
let { activities } = await this.orderService.getOrders(...)
```

dtslvr (changes requested):

dtslvr (approved):
Thank you for the contribution, @Fullstack-Arshad! 🙌🏻
Ready to merge now.

PR #3051, opened 2024-02-25
Fix importing by ISIN number

sahb1239:
When importing using CSV or JSON by ISIN number it the Yahoo finance provider will fail to get historical data due to the the ISIN number is not mapped to the symbol used by Yahoo Finance.

This sample CSV illustrates the problem:
```
Date,Account,Action,DataSource,Code,Quantity,unitPrice,Fee,Currency
2024-02-25,My Account,buy,YAHOO,DK0060442556,5,275.10,0,DKK
2024-02-25,My Account,buy,YAHOO,SPIC25KL.CO,5,275.10,0,DKK
```

Without the fix this will map to 2 different symbols as seen on the import screen here:
Screenshot 2024-02-25 154439

This leads to that the historical data gathering is failing and the same stock being shown twice:
image

This fix ensures that the correct symbol is returned from the yahoo-finance-service.ts. When trying to import the same sample file it now maps to the same symbol instead.
Screenshot 2024-02-25 154340

### DIFF
--- a/apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
+++ b/apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
@@ -196,7 +196,9 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
         shortName: assetProfile.price.shortName,
         symbol: assetProfile.price.symbol
       });
-      response.symbol = assetProfile.price.symbol;
+      response.symbol = this.convertFromYahooFinanceSymbol(
+        assetProfile.price.symbol
+      );
 
       if (assetSubClass === AssetSubClass.MUTUALFUND) {
         response.sectors = [];
@@ -229,10 +231,6 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
           ];
         }
       }
-
-      response.symbol = this.convertFromYahooFinanceSymbol(
-        assetProfile.price.symbol
-      );
 
       const url = assetProfile.summaryProfile?.website;
       if (url) {
--- a/apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
+++ b/apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
@@ -228,7 +228,16 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
             { name: assetProfile.summaryProfile?.sector, weight: 1 }
           ];
         }
+      } else if (assetSubClass == AssetSubClass.CRYPTOCURRENCY) {
+        response.symbol = response.symbol.replace(
+          new RegExp(`-${DEFAULT_CURRENCY}$`),
+          DEFAULT_CURRENCY
+        );
       }
+
+      /*
+
+      */
 
       const url = assetProfile.summaryProfile?.website;
       if (url) {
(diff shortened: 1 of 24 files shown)


dtslvr (changes requested):
The idea is great, but it breaks some other functionality: when you import `BTCUSD`, it shows as `BTC-USD` instead of `BTCUSD` (see `ok-btc.csv`).
I think this is because `yahooFinanceDataEnhancerService.getAssetProfile(symbol)`expects a symbol converted to _Yahoo_ format (via `convertToYahooFinanceSymbol()`).

sahb1239:
Nice catch - will take a look at this again later today / this week.

sahb1239:
> The idea is great, but it breaks some other functionality: when you import `BTCUSD`, it shows as `BTC-USD` instead of `BTCUSD` (see `ok-btc.csv`). I think this is because `yahooFinanceDataEnhancerService.getAssetProfile(symbol)`expects a symbol converted to _Yahoo_ format (via `convertToYahooFinanceSymbol()`).

Seems like the root cause is that yahoo-finance.service.ts -> search automatically removes '-' from the symbol. Using the import function it's possible to import with both symbol formats:
image

Both formats are currently valid due to the convertToYahooFinanceSymbol method and this also means that these symbols right now might be tracked multiple times and might require data migration if you want to combine these symbols.

I think the question now is which behavior you prefer - BTCUSD or BTC-USD which would mean either changing the search functionality or the import functionality. For the potential data migration I think that's best left in a separate issue where consequences / approaches can be discussed.

sahb1239:
I added the potential import fix where the format BTCUSD will be saved.

apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -228,6 +228,11 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
             { name: assetProfile.summaryProfile?.sector, weight: 1 }
           ];
         }
+      } else if (assetSubClass == AssetSubClass.CRYPTOCURRENCY) {
```
dtslvr, line 231: Can you remove this block and try to replace the following line above

```
response.symbol = assetProfile.price.symbol;
```

with

```
response.symbol = this.convertFromYahooFinanceSymbol(assetProfile.price.symbol);
```

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -38,17 +38,10 @@ export class YahooFinanceService implements DataProviderInterface {
   }: {
     symbol: string;
   }): Promise<Partial<SymbolProfile>> {
-    const { assetClass, assetSubClass, currency, name } =
+    const response =
```
dtslvr, line 41: This is simpler 🙂

```
return this.yahooFinanceDataEnhancerService.getAssetProfile(symbol);
```

dtslvr (changes requested):

dtslvr:
> I think the question now is which behavior you prefer - BTCUSD or BTC-USD which would mean either changing the search functionality or the import functionality.

`BTCUSD`

apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
sahb1239: Solved

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
sahb1239: Solved

sahb1239:
I updated the PR and retested with the sample files:
image

dtslvr (approved):
Thank you @sahb1239!

I have made some minor improvements (https://github.com/ghostfolio/ghostfolio/pull/3051/commits/8b08a1bf2ba22a573f8af1fbc3ee76020a49431e), ready to merge now.

PR #3052, opened 2024-02-25
Disable parallel execution of commands causing race condition

sahb1239:
Disable parallel execution of commands causing race condition between mkdir and cp


dtslvr (approved):
Nice catch! It seems to work 👍🏻

I was not aware that these commands are executed in parallel by default.

PR #3053, opened 2024-02-25
Clean up dist folders

dtslvr:


dtslvr:
FYI @sahb1239: Another small improvement

PR #3117, opened 2024-03-08
Updated historical data import

### LINKED ISSUE #3116: [BUG] Manually imported market data not showing correctly
**Bug Description**

Manually imported market data is not showing correctly in the overview and portfolio graphs:

Screen Shot 2024-03-08 at 16 17 33

After some debugging, I found that the source of the bug is related to *timezone* and *data decimation*. If I simply drop the `dateQuery` in the `marketDataService.getRange` function, I get the correct graph:

Screen Shot 2024-03-08 at 16 29 05

The first date in `dateQuery.in` is `"2017-05-17T00:00:00.000Z"` while the corresponding market data date is `"2017-05-16T22:00:00.000Z"`. I imagine all instruments are priced per day, regardless of being located in Switzerland or Japan, so is there are reason for keeping the timezone component?

If so, then the problem is the data decimation, and that we're querying the database for specific timestamps. In my opinion, it would make more sense to let the database handle the decimation, e.g. group by week and select first in each group to get weekly decimation.

Please let me know your thoughts and if you want me to take a look at a fix.

**To Reproduce**

1. Add NOK currency and set NOK as base currency.

2. Import activities from [ghostfolio-export-202403081157.json](https://github.com/ghostfolio/ghostfolio/files/14540598/ghostfolio-export-202403081157.json).
3. Import market data from [ghostfolio-pareto-market-data.csv](https://github.com/ghostfolio/ghostfolio/files/14540606/ghostfolio-pareto-market-data.csv)

**Environment**

- Ghostfolio Version 2.61.1
- Self-hosted
- Firefox
- Linux

helgehatt:
Send date strings as is and let server handle parsing

Fixes an issue where e.g. "2017-01-01" is sent as "2016-12-31T23:00:00.000Z"

Closes #3116


apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
```diff
@@ -195,15 +195,13 @@ export class AssetProfileDialog implements OnDestroy, OnInit {
           header: true,
           skipEmptyLines: true
         }
-      ).data;
+      ).data as { date: string; marketPrice: number }[];
```
dtslvr, line 198: `as UpdateMarketDataDto[]`

dtslvr (changes requested):
Thanks for your contribution, @helgehatt.

Can you do the same [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-market-data-detail/market-data-detail-dialog/market-data-detail-dialog.component.ts#L60) as well?

helgehatt:
@dtslvr Happy to help. In this case the date is a `Date` object, so we have to convert it to string in one way or another. Do you know if this method also causes issues?

dtslvr:
> @dtslvr Happy to help. In this case the date is a `Date` object, so we have to convert it to string in one way or another. Do you know if this method also causes issues?

Isn't it the same because `toISOString()` is also called (`this.data.date.toISOString()`)❓
The functionality is broken (same as import but used to edit a single entry via GUI).

helgehatt:
> > @dtslvr Happy to help. In this case the date is a `Date` object, so we have to convert it to string in one way or another. Do you know if this method also causes issues?
> 
> Isn't it the same because `toISOString()` is also called (`this.data.date.toISOString()`)❓ The functionality is broken (same as import but used to edit a single entry via GUI).

Converting from `Date` to string should be fine. The problem is when you convert from string to a `Date` object. That’s when timezone issues arise. So to achieve the same thing here, we would have to modify `date` in the interface to be of type string and avoid the conversion to `Date` in the first place.

https://github.com/ghostfolio/ghostfolio/blob/ba73f6de2e3019c1803d9e5fa45fca66d80bb618/apps/client/src/app/components/admin-market-data-detail/market-data-detail-dialog/interfaces/interfaces.ts#L5-L12

Alternatively, instead of avoiding conversion we can try to fix the conversion. Obviously the `parseDate` function used in the `AssetProfileDialog.onImportHistoricalData` method is giving unexpected results. Do you know how to debug the client side code? Much easier to bugfix when you can see the values of all variables.

dtslvr:
>  ...to achieve the same thing here...

I don't think there is any difference as they share the same endpoint which consumes `ISO8601` (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/admin/admin.controller.ts#L249)). In one case, the user inputs a date as `CSV` (parse date locally) and in the other case the date comes from a date picker (local date).

> Do you know how to debug the client side code?

You can either use the developer tools of your browser or `console.log()` statements.

helgehatt:
So the latest changes is how it looks like if we do the same in both cases and avoid conversion to `Date` altogether.

dtslvr (approved):
Thanks for your contribution, it looks pretty simple like this 🙂
Ready to merge now.

PR #3125, opened 2024-03-09
Fix average price calculation by only considering BUY transactions

gizmodus:
There was a problem with the calculation of the average price (which is used to calculate the performance from sells). Now, only BUY transactions are used to calculate the average price. Before, SELL transactions also had an effect on the average price which probably does not make sense. What do you think?

The small performance difference between the 1 SELL unit test and the 2 SELL unit test is because of the small, unknown time period between the two SELL transactions on the same day. This has an effect on the time weighted total investment.


dtslvr (approved):
Thank you for providing this fix, @gizmodus 🤩

PR #3145, opened 2024-03-13
Enable account balance update for fee and interest activities

rodriguestiago0:


dtslvr (approved):
Thank you for your contribution, @rodriguestiago0! 🙌🏻 

I had to slightly modify your change because the checkbox should only be enabled if an account has been selected.

In this move, I have implemented it for fees, as it works very similarly.

hulmgulm:
@dtslvr in version 2.155.0 I see the "Update Cash Balance" checkbox enabled for Dividends and Interrests and I can check it. But once I change the date, it gets disabled and unchecked. Is this the expected behaviour or is this a bug?

dtslvr:
> @dtslvr in version 2.155.0 I see the "Update Cash Balance" checkbox enabled for Dividends and Interrests and I can check it. But once I change the date, it gets disabled and unchecked. Is this the expected behaviour or is this a bug?

@hulmgulm Yes, that's the expected behavior. The _Update Cash Balance_ functionality is only available when the date selected is today's date.

PR #3167, opened 2024-03-18
Fix chart tooltip of benchmark comparator

helgehatt:
- X-axis marker is completely off (marks between 2019 and 2020, but cursor is in 2017).
- Labels doesn't always show for portfolio, and labels show twice for benchmark.
- Decimation is not working properly for benchmark data.
Screenshot from 2024-03-18 17-26-13

Screenshot from phone before changes:
FA360E9E-A91B-4148-9145-36E067F0ED29

After changes:
35381B66-AB3A-489E-A8F4-0FED66A1B750

Updates include changing tooltip mode from `"x"` to `"index"`. `"x"` mode just isn't working, it either displays too many labels or too few. To make `"index"` mode work, I've made changes so that benchmark data is fetched the exact same way as performance data. Additionally, I've modified the chart data by filtering out dates which are not present in both datasets. This is necessary to make sure that the indexing strategy works even though there is missing market data for either dataset.

### DIFF
--- a/apps/api/src/app/benchmark/benchmark.service.ts
+++ b/apps/api/src/app/benchmark/benchmark.service.ts
@@ -13,7 +13,6 @@ import {
 import {
   DATE_FORMAT,
   calculateBenchmarkTrend,
-  eachDayOfInterval,
   parseDate,
   resetHours
 } from '@ghostfolio/common/helper';
@@ -29,7 +28,13 @@ import { BenchmarkTrend } from '@ghostfolio/common/types';
 import { Injectable, Logger } from '@nestjs/common';
 import { SymbolProfile } from '@prisma/client';
 import { Big } from 'big.js';
-import { differenceInDays, format, isSameDay, subDays } from 'date-fns';
+import {
+  differenceInDays,
+  eachDayOfInterval,
+  format,
+  isSameDay,
+  subDays
+} from 'date-fns';
 import { isNumber, last, uniqBy } from 'lodash';
 import ms from 'ms';
 
@@ -210,23 +215,26 @@ export class BenchmarkService {
 
   public async getMarketDataBySymbol({
     dataSource,
-    startDate,
     endDate = new Date(),
+    startDate,
     symbol,
     userCurrency
   }: {
-    startDate: Date;
     endDate?: Date;
+    startDate: Date;
     userCurrency: string;
   } & UniqueAsset): Promise<BenchmarkMarketDataDetails> {
     const marketData: { date: string; value: number }[] = [];
 
     const days = differenceInDays(endDate, startDate) + 1;
-    const dates = eachDayOfInterval({
-      start: startDate,
-      end: endDate,
-      step: Math.round(days / Math.min(days, MAX_CHART_ITEMS)),
-      includeEnd: true
+    const dates = eachDayOfInterval(
+      {
+        start: startDate,
+        end: endDate
+      },
+      { step: Math.round(days / Math.min(days, MAX_CHART_ITEMS)) }
+    ).map((date) => {
+      return resetHours(date);
     });
 
     const [currentSymbolItem, marketDataItems] = await Promise.all([
@@ -301,9 +309,12 @@ export class BenchmarkService {
       });
     }
 
-    const includesToday = isSameDay(parseDate(last(marketData).date), endDate);
+    const includesEndDate = isSameDay(
+      parseDate(last(marketData).date),
+      endDate
+    );
 
-    if (currentSymbolItem?.marketPrice && !includesToday) {
+    if (currentSymbolItem?.marketPrice && !includesEndDate) {
       const exchangeRate =
         exchangeRates[`${currentSymbolItem.currency}${userCurrency}`]?.[
           format(endDate, DATE_FORMAT)
--- a/apps/api/src/app/benchmark/benchmark.service.ts
+++ b/apps/api/src/app/benchmark/benchmark.service.ts
@@ -13,6 +13,7 @@ import {
 import {
   DATE_FORMAT,
   calculateBenchmarkTrend,
+  eachDayOfInterval,
   parseDate,
   resetHours
 } from '@ghostfolio/common/helper';
@@ -28,13 +29,7 @@ import { BenchmarkTrend } from '@ghostfolio/common/types';
 import { Injectable, Logger } from '@nestjs/common';
 import { SymbolProfile } from '@prisma/client';
 import { Big } from 'big.js';
-import {
-  differenceInDays,
-  eachDayOfInterval,
-  format,
-  isSameDay,
-  subDays
-} from 'date-fns';
+import { differenceInDays, format, isSameDay, subDays } from 'date-fns';
 import { isNumber, last, uniqBy } from 'lodash';
 import ms from 'ms';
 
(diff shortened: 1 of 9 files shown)


apps/client/src/app/pages/portfolio/analysis/analysis-page.html
```diff
@@ -66,6 +66,7 @@ <h1 class="d-none d-sm-block h3 mb-3 text-center" i18n>Analysis</h1>
               />
             </div>
           </div>
+          <div><hr /></div>
```
dtslvr, line 69: The line stands for a sum and is intentionally not there 😏

apps/api/src/app/benchmark/benchmark.service.ts
```diff
@@ -217,6 +224,17 @@ export class BenchmarkService {
   } & UniqueAsset): Promise<BenchmarkMarketDataDetails> {
     const marketData: { date: string; value: number }[] = [];
 
+    const endDate = new Date();
```
dtslvr, line 227: Can you extend the function arguments by `endDate = new Date();` (`endDate?: Date;`) instead of this?

apps/api/src/app/benchmark/benchmark.service.ts
```diff
@@ -217,6 +224,17 @@ export class BenchmarkService {
...
+    const days = differenceInDays(endDate, startDate) + 1;
+    const step = Math.round(days / Math.min(days, MAX_CHART_ITEMS));
+    const dates = Array.from({ length: days / step }).map((_, i) =>
```
dtslvr, line 230: Could you use `forEach()` instead of `map()`?

apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -203,11 +203,8 @@ export class PortfolioCalculator {
     const dataGatheringItems: IDataGatheringItem[] = [];
     const firstIndex = transactionPointsBeforeEndDate.length;
 
-    let day = start;
-
-    while (isBefore(day, end)) {
+    for (let day = start; isBefore(day, end); day = addDays(day, step)) {
```
dtslvr, line 206: Can you use [eachDayOfInterval](https://date-fns.org/v2.29.3/docs/eachDayOfInterval)? I think that's even simpler.

dtslvr (changes requested):
Thank you very much for your contribution, @helgehatt. I have added a few comments.

apps/client/src/app/pages/portfolio/analysis/analysis-page.html
helgehatt: Aha, sorry, I just assumed it had been forgotten :sweat_smile:

apps/api/src/app/benchmark/benchmark.service.ts
helgehatt: Done :+1:

apps/api/src/app/benchmark/benchmark.service.ts
helgehatt: Replaced with `eachDayOfInterval` :+1:

apps/api/src/app/portfolio/portfolio-calculator.ts
helgehatt: I actually looked through the docs for exactly that function, but assumed it would be under "Day" functions. Have replaced it now :+1:

apps/api/src/app/benchmark/benchmark.service.ts
```diff
@@ -232,7 +252,7 @@ export class BenchmarkService {
           dataSource,
           symbol,
           date: {
-            gte: startDate
+            in: dates
```
dtslvr, line 255: This does not seem to work. `marketDataItems` is `[]` and thus this error message in the log: `No historical market data has been found for`...

apps/api/src/app/portfolio/portfolio-calculator.ts
```diff
@@ -199,16 +200,10 @@ export class PortfolioCalculator {
...
     const dataGatheringItems: IDataGatheringItem[] = [];
     const firstIndex = transactionPointsBeforeEndDate.length;
 
-    let day = start;
-
-    while (isBefore(day, end)) {
-      dates.push(resetHours(day));
-      day = addDays(day, step);
-    }
+    const dates = eachDayOfInterval({ start, end }, { step });
```
dtslvr, line 206: For `api/v2/portfolio/performance?range=1d` I just get a horizontal line (0). It seems to work as before if I do the following:

```
const dates = eachDayOfInterval({ start, end }, { step }).map((date) => {
  return resetHours(date);
});
```

dtslvr (changes requested):
Unfortunately, my manual tests were not successful.

karter16:
Just adding my thoughts here re Missing horizontal ruler between "Absolute Asset Performance" and "Absolute Currency Performance"

I wouldn't consider that this is missing a horizontal ruler. As I read it currently it makes sense that Absolute Asset Performance and Absolute Currency Performance don't have a ruler between them, as they are both components feeding into "Absolute Net Performance" I would therefore consider that a horizontal ruler is not necessary. Otherwise they are going to present as three separate categories of equal importance, when in fact the first two are components of the third (e.g. the total).

helgehatt:
Thank you for the comment @karter16. Thomas already said the line was intentionally not there and it has been removed 👍

dtslvr (changes requested):
Thank you for the changes. Data is now loaded, however there are some major flaws:

* Selecting a date range of 5+ years: line of portfolio is rendered as a green flat line
* Selecting a date range of week to date: line of portfolio is rendered, but not starting from 0
* Unselecting the benchmark: no line is rendered (should render line of portfolio)

helgehatt:
Yeah, I should've seen this coming. The problem is related to keeping the index of the performance data and the benchmark data identical. The previous code removes dates that are not in the intersection of performance data dates and benchmark data dates to avoid gaps in the charts. This obviously doesn't work well, because the benchmark data is missing altogether when not having a benchmark selected.

The new version computes the union of dates instead of intersection, and returns undefined values for dates where market data is missing. As a result, you get gaps in the benchmark data as shown here

Screenshot 2024-03-23 at 11-58-40 ${title}

but at the same time the available market data should always be shown, unlike what you experienced before. The strange thing is that I would expect there to be a gap in the performance data as well, since I've added a gap in the performance market data as well (the second data point is undefined and the fifth is set to zero)

Screenshot 2024-03-23 at 12-00-47 ${title}

It is also interesting to note that a market price of zero is treated as undefined in terms of portfolio performance.

dtslvr:
> The new version computes the union of dates instead of intersection, and returns undefined values for dates where market data is missing.

This is getting more and more complicated. With the knowledge that you have collected regarding the original problem, don't you think that the solution can be simplified? Instead of the union, maybe you can base it on the performance data?

helgehatt:
> This is getting more and more complicated. With the knowledge that you have collected regarding the original problem, don't you think that the solution can be simplified? Instead of the union, maybe you can base it on the performance data?

I agree. Last edit is based on the performance data, so this solution is quite simple.

dtslvr (approved):
Thanks for the adaptations, looks fine now 👍🏻
Ready to merge.

helgehatt:
I think it's a mistake to rely on `eachDayOfInterval` from `date-fns`. It returns timezone-aware dates:
```
> dateFns.eachDayOfInterval({ start: new Date("2021-01-01"), end: new Date("2021-01-03") }).map((x) => x.toISOString())
["2020-12-31T23:00:00.000Z", "2021-01-01T23:00:00.000Z", "2021-01-02T23:00:00.000Z"]
```

There are so many one-off date bugs in this app. Just to mention a few; the historical market data import, the first activity date and all the YTD/MTD/WTD ranges include the date in the previous year/month/week.

Issue #2879, opened 2024-01-15
[BUG] zombie processes after ghostfolio starts


nantostars:
**Bug Description**

I'm using ghostfolio 2.38.0 with docker compose using the template available in the repository (docker/docker-compose.yml).
Once started, the ghostfolio container leaves two zombie/defunct processes, even if the application is working correctly, there is something not managed correctly on how these processes exit.

I tried to understand which are the processes that fail, running ps command many times after the container start and this are the command that I've seen:

- ![image](https://github.com/ghostfolio/ghostfolio/assets/46993725/eee42417-3899-447a-8d54-1c442c981c55)

- ![image](https://github.com/ghostfolio/ghostfolio/assets/46993725/abbaf74c-48a9-4002-a5c7-1dbc5ec335a9)

- ![image](https://github.com/ghostfolio/ghostfolio/assets/46993725/9b14dcf9-18fc-48f7-9787-7809c590fc56)

**To Reproduce**

1. restart the ghostfolio container
2. check the presence of zombie processes on the docker host with command: ps aux | grep Zs 

**Expected behavior**

No zombie processes

**Screenshots**

This is the situation after processes become zombies:
![image](https://github.com/ghostfolio/ghostfolio/assets/46993725/8ec5a9b9-1c2f-4110-a5fb-a8b093234829)

**Logs**

**Environment**

- Ghostfolio Version 2.38.0
- Self-hosted
- Experimental Features disabled
- Browser Firefox
- OS Ubuntu 22.04 LTS running Docker version 24.0.7 managed with Portainer

Sonlis:
I'll have a look in the following days if that's fine by you @dtslvr

Sonlis:
It seems to be an issue with prisma: https://github.com/prisma/prisma/issues/17024
I am checking on their side if something can be done 👍

Sonlis:
Thanks to this article: https://maximorlov.com/process-signals-inside-docker-containers/ I understood why this is happening.
The CMD entrypoint starts the process as init process with PID 1. This process is responsible for handling signals. Yarn is not a process manager, so running `CMD [ "yarn", "start:production" ]` disables any signal handling in the container.

I am trying to think of the better way to handle it. An easy solution would be to use [tiny](https://github.com/krallin/tini), which introduces signal handling for very few modifications, by just adding `init: true` to the docker-compose file.
```yaml
  ghostfolio:
    image: ghostfolio/ghostfolio:latest
    init: true
```
```sh
[...]

PR #3168, opened 2024-03-19
Enable signal handling from host to container, and within container

Sonlis:
Solves https://github.com/ghostfolio/ghostfolio/issues/2879, as well as enabling signals between host and container.

- Do not use yarn as entrypoint, as it spawns multiple processes and does not clean them up within the container as long as `node main` is running.
- Instead, use a script that runs the db migrations and then start node, leaving only 2 processes: the script execution and `node`.
- Use [tini](https://github.com/krallin/tini) in docker-compose files, to handle signals from the host and terminate gracefully on SIGTERM. Before, using SIGTERM would terminate the container with exit code 1, which is an application error. With those changes, it exists with error 143, which is a graceful shutdown.

To see the changes' results:
1. spawn the containers with docker compose and exec into the ghostfolio container: `docker exec -it <container-id> /bin/sh`.
1. Add `ps`: `apt-get update && apt-get install procps -y`
1. run `ps -ef`:  
```sh
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 16:35 ?        00:00:00 /sbin/docker-init -- docker-entrypoint.sh ./entrypoint.sh
root        10     1  0 16:35 ?        00:00:00 /bin/sh ./entrypoint.sh
root       145    10 30 16:35 ?        00:00:05 node main 
```

4. Exit the container, and run `docker stop <container-id>`
5. Run `docker ps -a` to check the exit code:
```sh
CONTAINER ID   IMAGE                   COMMAND                  CREATED       STATUS                       PORTS      NAMES
63c52e56c4cb   ghostfolio_ghostfolio   "docker-entrypoint.s…"   3 hours ago   Exited (143) 2 seconds ago              ghostfolio-ghostfolio-1 
```
Code 143 means the container was terminated gracefully.


Dockerfile
```diff
@@ -58,5 +58,6 @@ RUN apt update && apt install -y \
 
 COPY --from=builder /ghostfolio/dist/apps /ghostfolio/apps
 WORKDIR /ghostfolio/apps/api
+COPY ./docker/entrypoint.sh ./entrypoint.sh
```
dtslvr, line 61: Why not use this earlier, where the other `COPY` instructions are?

docker/entrypoint.sh
dtslvr, line 1: Still `yarn`...?

dtslvr (changes requested):

dtslvr (commented):
> Use [tini](https://github.com/krallin/tini) in docker-compose files

Where is it? 🙂

Dockerfile
```diff
@@ -56,7 +56,14 @@ RUN apt update && apt install -y \
     openssl \
     && rm -rf /var/lib/apt/lists/*
 
+# Add tini, which is an init process that handles signaling within the container
```
dtslvr, line 59: What does [that](https://github.com/krallin/tini?tab=readme-ov-file#using-tini) mean?

> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just pass the --init flag to docker run.

Do we actually need to add _Tini_?

Sonlis: From my understanding, it is included in the docker Daemon, not every docker images itself. I think adding it to the image is better, so it's transparent for the end user. It would be possible to just add `init: true` to the docker compose files, and add to the documentation to add `--init true` to the `docker run` command. IMO it's not a huge difference as Tini is, well, tiny, so the docker image size is barely increased.

dtslvr:
Hey @Sonlis

The container does not start anymore:

```
Running database migrations
+ echo Running database migrations
+ npx prisma migrate deploy
Error: Could not find a schema.prisma file that is required for this command.
You can either provide it with --schema, set it as `prisma.schema` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
```

Sonlis:
Hmmm, I tested locally and it was working fine 🤔 I don't have access to my laptop until Tuesday, I'll check then

dtslvr:
I was not able to fix it, had to revert it.

PR #3193, opened 2024-03-25
Fix update cash balance

rodriguestiago0:


apps/api/src/app/account/account.service.ts
```diff
@@ -57,7 +57,7 @@ export class AccountService {
   > {
     const { include = {}, skip, take, cursor, where, orderBy } = params;
 
-    include.balances = { orderBy: { date: 'desc' }, take: 1 };
+    include.balances = { orderBy: { createdAt: 'desc' }, take: 1 };
```
dtslvr, line 60: This will not work if you have created a balance today for a previous day.

dtslvr (changes requested):

apps/api/src/app/account/account.service.ts
rodriguestiago0: What do you mean by creating a balance for today?

rodriguestiago0: In my opinion, cash balance table should only have one entry per account.
And use postgres capability on conflict to update the existing balance with the change.

dtslvr: > In my opinion, cash balance table should only have one entry per account.

No, we need multiple entries to show the progress of the account balance.

rodriguestiago0: So, when creating an account balance in the past, it needs to be inserted in order or be recalculated.

dtslvr: No, I don't think so.
However, your comment leads me to the conclusion that the current implementation for update cash balance only works if the activity (e.g. interest) is today. In the GUI, the checkbox must be disabled if the date differs from today.

rodriguestiago0: They need to be newer than the last account balance date.

rodriguestiago0: I think the best approach is to insert on this table the balance change and use the a view to calculate the accumulated values based on the date.

I will do this change and fix this

dtslvr: > I will do this change and fix this

I have some doubts that it will work and would suggest to solely disable the checkbox based on the date.

dtslvr: PR: #3229

studentism:
Sorry, I'm still lost on how this is marked as completed? The "update cash balance" checkbox is still available regardless of the date. It is also unclear how to manually recreate historical transactions like this.

PR #3194, opened 2024-03-26
Add user currency to export

christoph93:


apps/api/src/app/export/export.service.ts
```diff
@@ -95,7 +99,10 @@ export class ExportService {
                 : SymbolProfile.symbol
           };
         }
-      )
+      ),
+      user: {
+        settings: { currency: userWithSettings.Settings.settings.baseCurrency }
```
dtslvr, line 104: You can use `userCurrency`, then you don't even have to make a query 🙂

```
settings: { currency: userCurrency }
```

dtslvr (changes requested):
Thank you for the contribution @christoph93! 😎

I have found an easier way to access the currency of the user. Can you please revert everything that is not needed?

apps/api/src/app/export/export.service.ts
christoph93: Fixed!

dtslvr (approved):
Thank you! 👍🏻
Ready to merge now.

PR #3195, opened 2024-03-26
Better signal handling

Sonlis:
In this PR: https://github.com/ghostfolio/ghostfolio/pull/3168/files the workdir argument was removed, so the entrypoint script was run from `/`, and couldn't find the files required to run the migration and seeding. Set back the workdir so the script is instead evaluated from `/ghostfolio/apps/api` where the necessary files are located.


dtslvr:
Thanks @Sonlis. Have you tested it thoroughly?

Sonlis:
Yes, made sure the container was able to start properly 👍

dtslvr (approved):

PR #3199, opened 2024-03-27
Allow updating name of asset profile in Admin Control panel

### LINKED ISSUE #3151: Allow to edit name of asset profile in Admin Control panel
asset-profile-dialog

Currently, it is only possible to edit the name of an asset profile where data source is `MANUAL` (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html#L195)). Enable this functionality for every asset profile regardless of the data source.

If data source is `MANUAL`, the name can still be stored in the `SymbolProfile` database table. For all other data sources, an entry in `SymbolProfileOverrides` needs to be created or updated.

Fedron:
Closes #3151 

- Removes the restriction of only allowing an update to the name if the data source is manual in the client form
- For a data source that is manual, update the name on the symbol profile
- Create/update symbol profile override and set the name in the override if the data source is anything other than manual

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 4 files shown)


dtslvr (approved):
That looks fine! Thank you for this cool contribution, @Fedron 🙌🏻

PR #3206, opened 2024-03-29
Show market data for missing months in Market Data display

### LINKED ISSUE #2539: Daily data indicator control on the Market Data > Symbol screen does not display month if there are no data points for that month
**Bug Description**

On the Admin Control > Market Data > Symbol screen under the historical price graph is a control/section that list out past months and shows a red/green square for each day of that month depending on whether historical price data is populated for that day or not.

I am unsure whether it is by design or not, but if there are no data points for a given month, the entire month is missing from that control/list of past months. It makes sense that you wouldn't show months earlier than the first data point you have, but if the month/s that are missing are in the middle of your data set it's easy to miss that you have missing data for the month. 

**To Reproduce**

1. Create a manual source and populate at least one data point for a month some time ago. (or use an automated source with an entire missing month of data)
2. Populate a data point from today/yesterday.
3. Observe that intermediary months are missing entirely from the control

**Expected behavior**
My suggestion would be that the behaviour should be that any month with no data points, where the month is after the oldest data point you do have, should be displayed in the list (with every day marked in red).

**Screenshots**

Screenshot 2023-10-24 at 8 26 51 PM

**Logs**

**Environment**

- Self-hosted
- Ghostfolio Version 2.14.0
- Browser Safari Version 17.0 (19616.1.27.211.1)
- OS MacOS Sonoma

**Additional context**

Fedron:
Resolves #2539 

For months that have no market data defined, a month will be added with empty data (similar to how red boxes for days with no data are added) i.e
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Added missing dates to edit historical market data in the asset profile details dialog of the admin control panel
+
 ## 2.68.0 - 2024-03-29
 
 ### Added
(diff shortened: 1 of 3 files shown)


dtslvr (approved):
Thank you very much for your contribution, @Fedron! 😎
I have slightly modified the code to also support dates if you currently have no market data at all.

PR #3211, opened 2024-03-30
Add support for stock split orders

Fedron:
Add new order type "SPLIT" to enable tracking stock splits.
- Allow a negative number in the DTO for just the "SPLIT" order type

Uses the quantity field as the ratio for the stock split:
- A positive quantity represents a forward stock split (i.e. quantity = 2 = 2 for 1 split)
- A negative quantity represents a reverse stock split (i.e. quantity = -2 = 1 for 2 split)

The portfolio calculator now also takes in "SPLIT" orders and will update the total quantity of shares accordingly.

### DIFF
--- a/apps/api/src/app/order/create-order.dto.ts
+++ b/apps/api/src/app/order/create-order.dto.ts
@@ -15,9 +15,32 @@ import {
   IsNumber,
   IsOptional,
   IsString,
-  Min
+  Min,
+  ValidationArguments,
+  ValidationOptions,
+  registerDecorator
 } from 'class-validator';
 import { isString } from 'lodash';
+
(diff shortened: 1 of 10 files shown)


dtslvr:
Thank you for this suggestion, @Fedron. I'm pretty sure it is a useful feature. As the portfolio calculator is currently undergoing a major rebuild, I think it is not the ideal moment to add further complexity.

Hugo-Persson:
I think this looks like a great suggestion! @dl90 when do you think this can be merged in and released?

Grasume:
@dtslvr  does this need a new PR to get this added ? or can this be added still ?

PR #3229, opened 2024-04-01
Feature/disable option to update cash balance if date is not today

dtslvr:


davidojedalopez:
Would this mean that there's no way to re-create the historical changes of balances of an account? 

For example, if I just started using Ghostfolio, there will be no way to know how much my account has grown/shrunk over a period, even if I have all the transactions' details elsewhere?

Or is there an import feature that will allow this?

dtslvr:
> Would this mean that there's no way to re-create the historical changes of balances of an account?

No, the current implementation for update cash balance only works if the activity is today. To prevent unexpected results, the checkbox will be disabled.

studentism:
> To prevent unexpected results, the checkbox will be disabled.
The checkbox is still clickable. (I'm using the hosted premium version.) It doesn't do anything, and it is slightly confusing to have the option.

Just to be clear, is there no way to manually recreate historical data? So for accounts that I've had for over a decade, I'm just supposed to start tracking from today? And additionally, for dividends or interest, I have to enter it on the exact same day that I receive it in order to accurately track that information?

Sjohn21:
Is there any reason why the the current implementation for update cash balance only works if the activity is today?
I understand that updating cash balance between dates in the past would give issues with the current implementation, however an update on the date or after the date of the last cash balance point should work fine right?
If agreed I can see if I can create a PR.

PR #3230, opened 2024-04-01
Validate form DTO before closing & display validation errors

### LINKED ISSUE #3219: [BUG] API errors for forms not communicated to the user
**Bug Description**

When the client sends a request to the API and an error is returned such as 400 Bad Request when invalid form data is supplied there is no good indication of this. Currently the form just closes and that is it, you can only see the issue if you use Dev Tools and look at the console.

**To Reproduce**

1. Navigate to Portfolio > Activities
2. Create a new `BUY` activity
3. Set the quantity to a negative number

**Expected behavior**

The form should stay open and shouldn't close straight away. It would make sense the form would only close once it receives a successful status code from the API.

If the API returns a non-successful status code it would be useful to see what that was:
- A 500 error should probably just result in an error notification and closing(?) the dialog
- A 400 error (such as when providing a negative value for the quantity) would ideally keep the dialog open and highlight the field that caused the error and the message.

**Screenshots**

In a self-hosted environment all you see is this in the console:
image

**Environment**

- Ghostfolio Version 2.68.0
- Self-hosted
- Browser: Edge 123.0.2420.65
- OS: Unraid

Fedron:
Resolves #3219

Show validation errors for activity and account form dialog.

Create helper function `validateObjectForForm` that will take in an object and run the validators on that object. If any validation errors are encountered they are added to that form control's errors and displayed in the form i.e.

image

The form will only be closed if no validation errors are encountered.


apps/client/src/app/util/validation.util.ts
```diff
@@ -0,0 +1,27 @@
+import { FormGroup } from '@angular/forms';
+import { plainToInstance } from 'class-transformer';
+import { validate } from 'class-validator';
+
+export function validateObjectForForm<T>(
+  obj: T,
+  ctor: { new (): T },
```
dtslvr, line 7: `classDto` ❓

apps/client/src/app/util/validation.util.ts
```diff
@@ -0,0 +1,27 @@
...
+  form: FormGroup,
+  onSuccess: () => void
+): void {
+  const objInstance = plainToInstance(ctor, obj);
+
+  validate(objInstance as object).then((errors) => {
```
dtslvr, line 13: `const errors = await validate(objInstance as object);`

apps/client/src/app/util/validation.util.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (errors.length === 0) {
+      onSuccess();
+    } else {
+      errors.forEach((error) => {
```
dtslvr, line 17: `for (const error of errors) {` (to reduce nesting)

apps/client/src/app/util/validation.util.ts
```diff
@@ -0,0 +1,27 @@
...
+  obj: T,
+  ctor: { new (): T },
+  form: FormGroup,
+  onSuccess: () => void
+): void {
+  const objInstance = plainToInstance(ctor, obj);
+
+  validate(objInstance as object).then((errors) => {
+    if (errors.length === 0) {
+      onSuccess();
```
dtslvr, line 15: I'm not a fan of passing a callback. Is it possible to change the function to return a promise and `resolve()` here?

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
dtslvr, line 1: These error messages are rather technical. I would not display them at all, as they are also not localized.

apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
dtslvr, line 1: These error messages are rather technical. I would not display them at all, as they are also not localized.

dtslvr (changes requested):
This is a interesting approach! Thanks for your PR, @Fedron.

Is it possible to easily validate already on blur (when the value changes), instead of on submit?

apps/client/src/app/util/validation.util.ts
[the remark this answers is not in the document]
Fedron: Renamed

apps/client/src/app/util/validation.util.ts
Fedron: Renamed

apps/client/src/app/util/validation.util.ts
Fedron: Changed

apps/client/src/app/util/validation.util.ts
Fedron: Changed to return a promise that will resolve if there are no errors, otherwise reject

apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
Fedron: Removed

apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
Fedron: Removed

Fedron:
> This is a interesting approach! Thanks for your PR, @Fedron.
> 
> Is it possible to easily validate already on blur (when the value changes), instead of on submit?

I'm pretty sure this would as easy as adding a validator to the controls when the form group is being created. Currently only a required validator is ever used.

Reason I went for the approach of using the DTO though is that that same validation logic in the DTO doesn't need to be duplicated to the form group builder

dtslvr (approved):
Cool, thank you for the changes. I have made some final adaptions now.

Would you like to adopt this concept as well for the access, asset profile, tag and platform management, @Fedron?

Fedron:
> Cool, thank you for the changes. I have made some final adaptions now.
> 
> Would you like to adopt this concept as well for the access, asset profile, tag and platform management, @Fedron?

Thank you for the feedback!

Absolutely, I'll raise a new PR(s) extending this to the areas/forms you mentioned.

dtslvr:
> Absolutely, I'll raise a new PR(s) extending this to the areas/forms you mentioned.

I have created a follow-up ticket for it: #3294

PR #3232, opened 2024-04-01
Enable tini from docker compose instead of adding it to dockerfile

Sonlis:
This discussion: https://github.com/ghostfolio/ghostfolio/discussions/3216 raised an issue where `tini` would not be executable on platforms other than `linux/amd64`.

`Tini` supports different architectures, but it would require to modify the build steps and have an image per architecture, which is cumbersome and not user-friendly.

Add it instead to `docker-compose` so the docker daemon handles setting up tini, no matter the architecture.


dtslvr (approved):
Thank you for the fix, @Sonlis!

Issue #3282, opened 2024-04-16
[BUG] Pre-commit fails by default because the script is not executable by default


Sonlis:
**Bug Description**

By default, pre-commit has a permission of `744`, which means users can only read the file, not execute it. When committing, this warning is shown:
```
hint: The 'git-hooks//pre-commit' hook was ignored because it's not set as executable.
```

**To Reproduce**

1. Clone the repository
2. Run `git config core.hooksPath ./git-hooks/`
3. Commit an empty commit.

**Expected behavior**

By default, the script should be executable.

**Environment**

- Ghostfolio Version 2.72.0
- Self-hosted
- MacOs

dtslvr:
Closed via https://github.com/ghostfolio/ghostfolio/pull/3283.

PR #3283, opened 2024-04-16
Make pre-commit executable by default

Sonlis:
Solves this issue: https://github.com/ghostfolio/ghostfolio/issues/3282, by setting the default permissions on `pre-commit` to 755 from 644, effectively making the script executable by default.


git-hooks/pre-commit
dtslvr, line 1: Why is there no change? I only see this:

> Empty file.

Sonlis: The changes are at the top, next to the file name: `100644 -> 100755`. This is the default permission of the file when you clone the repository.

dtslvr (approved):

PR #3368, opened 2024-05-04
Feature/move holdings table to holdings tab of home page

dtslvr:


karter16:
Just wanted to say I love this change! Removes duplication and simplifies!

PR #3393, opened 2024-05-10
Feature/rework portfolio calculator

gizmodus:
### High Priority
- [x] Overview chart not the same as in holding table (difference seems to be the fees): Import `test/import/ok-vti-buy-on-ibkr.csv` in a new user account to reproduce
- [x] `netPerformancePercentWithCurrencyEffect` is not the same in `GET api/v1/portfolio/holdings?range=max` (_Holdings_ table) and `GET api/v1/portfolio/position/:dataSource/:symbol` (_Holding_ details dialog)
- [x] `WTD` / `MTD` chart starts one day too early

### Low Priority
- [ ] Liabilities are missing sometimes in _Summary_
- [ ] Portfolio page logs `Computed portfolio snapshot in XXX seconds` three times


apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -130,17 +138,27 @@ export abstract class PortfolioCalculator {
       });
 
     this.redisCacheService = redisCacheService;
-    this.useCache = useCache;
+    this.useCache = false; // useCache;
     this.userId = userId;
 
-    const { endDate, startDate } = getInterval(dateRange);
```
dtslvr, line 136: `dateRange` is unused after this change leading to max performance regardless of the selected date range 🤔

dtslvr (approved):
Ready to merge.

PR #3413, opened 2024-05-15
Feature/set preselect account when there is only one

### LINKED ISSUE #3347: Preselect account in CreateOrUpdateActivityDialog
If the user has exactly one account, it should be preselected in the `CreateOrUpdateActivityDialog`.

create-or-update-activity-preselect-account

SirZemar:
Closes  #3347


dtslvr (approved):
Thank you very much for your contribution, @SirZemar!
I have rearranged the condition, however the logic is still the same. Ready to merge now.

Issue #3453, opened 2024-06-02
[BUG] Exception when disabling unconfigured Biometric Authentication


dittodhole:
**Bug Description**

Just wanted to try out Biometric Authentication, did not configure anything, and then wanted to disable it again.
Still experiencing #3439

**To Reproduce**

1. Open ghostfolio `http://ghostfolio:3333` and login as admin
2. Goto "My Ghostfolio" at `/en/account`
2. Enable "Biometric Authentication"
3. Disable "Biometric Authentication"
4. Confirm "Do you really want to remove this sign in method?"
5. Error "Oops! Something went wrong. Please try again later."

https://github.com/ghostfolio/ghostfolio/assets/429216/06dec3b9-2cb9-4b33-b2aa-cc7f801ae5b7

**Expected behavior**

No failure

**Logs**

```
[Nest] 115  - 06/02/2024, 2:55:44 PM   ERROR [ExceptionsHandler] 
Invalid `prisma.authDevice.delete()` invocation:
An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
PrismaClientKnownRequestError: 
Invalid `prisma.authDevice.delete()` invocation:
An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
    at In.handleRequestError (/ghostfolio/apps/api/node_modules/@prisma/client/runtime/library.js:122:6877)
    at In.handleAndLogRequestError (/ghostfolio/apps/api/node_modules/@prisma/client/runtime/library.js:122:6211)
    at In.request (/ghostfolio/apps/api/node_modules/@prisma/client/runtime/library.js:122:5919)
    at async l (/ghostfolio/apps/api/node_modules/@prisma/client/runtime/library.js:127:11167)
    at async AuthDeviceController.deleteAuthDevice (/ghostfolio/apps/api/main.js:1:522555)
    at async /ghostfolio/apps/api/node_modules/@nestjs/core/router/router-execution-context.js:46:28
    at async /ghostfolio/apps/api/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

**Environment**

- Ghostfolio Version 2.84.0
- Self-hosted
- Experimental Features enabled and disabled, tried both
- Brave 1.66.115
- Windows 11 Pro

juanabascal:
[...]

PR #3496, opened 2024-06-14
Feature/improve error handling in biometric authentication registration

juanabascal:
Fixing #3453.

My guess is that somehow authn support is not enabled in the user's browser. I manually disabled authn in Firefox and I was able to get the same experience as the user.

To disable Authn:
- Open Firefox and type about:config in the address bar.
- Search for security.webauth.webauthn and set it to false by double-clicking it. This will disable the WebAuthn API.

I have done a two-way solution:
1. Disable fingerprint login if Authn is not supported by the browser
2. In case of an error during fingerprint login set up, uncheck the toggle and show a snack message.


apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -69,12 +71,14 @@ export class UserAccountSettingsComponent implements OnDestroy, OnInit {
     private settingsStorageService: SettingsStorageService,
     private tokenStorageService: TokenStorageService,
     private userService: UserService,
+    private snackBar: MatSnackBar,
```
dtslvr, line 74: Please move up to keep the alphabetical order ⬆️

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -42,6 +43,7 @@ export class UserAccountSettingsComponent implements OnDestroy, OnInit {
   public hasPermissionToUpdateViewMode: boolean;
   public hasPermissionToUpdateUserSettings: boolean;
   public isAccessTokenHidden = true;
+  public isFingerprintSupported: boolean;
```
dtslvr, line 46: What about

`public isFingerprintSupported = this.doesBrowserSupportAuthn();`

and remove the initialization in the constructor?

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -277,13 +281,30 @@ export class UserAccountSettingsComponent implements OnDestroy, OnInit {
       });
   }
 
-  private registerDevice() {
+  private registerDevice(aEvent: MatSlideToggleChange) {
```
dtslvr, line 284: I would prefer to change this function to return a promise (and throwing an error) instead of wiring it with `MatSlideToggleChange`.

dtslvr (changes requested):
Thanks for your contribution, @juanabascal 🙌🏻
I have added a few comments.

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
juanabascal: Yup, that'd be another option.

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
juanabascal: I thought about that. I like it.

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
juanabascal: Didn't notice. Will update.

juanabascal:
@dtslvr addressed the comments. Let me know what you think.

dtslvr (approved):
Thank you for the adaptions! Looks fine now 😎
Ready to merge.

PR #3507, opened 2024-06-22
Faster autocomplete

Usiel:
See: https://github.com/ghostfolio/ghostfolio/discussions/3421

This implements a variant of the linked discussed idea. 

Initially, I tried making this work with the existing `/api/v1/portfolio/holdings` endpoint, but the latency was still too high for a smooth user experience. With the separate endpoint the feature is more like I imagined it to be. This is especially true when Ghostfolio is hosted locally and the difference between a strictly local request vs waiting for a 3rd party is more pronounced.

Screen Capture on 2024-06-22 at 11-48-25

The separate endpoint still hooks into the `OrderService` which requires a minor change to implement a `SEARCH_QUERY` filter in `getOrders(...)`.

There's still behavior I might want to optimize: If an item is highlighted and focused (keypress arrow down) and the remaining suggestions come in after that, the focus is lost.

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.controller.ts
+++ b/apps/api/src/app/portfolio/portfolio.controller.ts
@@ -1,5 +1,6 @@
 import { AccessService } from '@ghostfolio/api/app/access/access.service';
 import { OrderService } from '@ghostfolio/api/app/order/order.service';
+import { LookupItem } from '@ghostfolio/api/app/symbol/interfaces/lookup-item.interface';
 import { UserService } from '@ghostfolio/api/app/user/user.service';
 import { HasPermissionGuard } from '@ghostfolio/api/guards/has-permission.guard';
 import {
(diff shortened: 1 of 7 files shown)


apps/api/src/app/portfolio/portfolio.controller.ts
```diff
@@ -319,6 +320,28 @@ export class PortfolioController {
     return { holdings: Object.values(holdings) };
   }
 
+  @Get('lookup')
+  @UseGuards(AuthGuard('jwt'), HasPermissionGuard)
+  @UseInterceptors(TransformDataSourceInResponseInterceptor)
+  public async lookupSymbol(
+    @Headers(HEADER_KEY_IMPERSONATION.toLowerCase()) impersonationId: string,
+    @Query('accounts') filterByAccounts?: string,
```
Usiel, line 328: Not implemented on the client yet, but this allows us to filter by account if the user already selected an account.

dtslvr (commented):
Thanks a lot for creating this pull request, @Usiel! It indeed seems to be an improvement in terms of usability, but at the price of additional complexity. I am currently undecided whether this solution is optimal for this project.

Usiel:
I understand, so far nobody else seemed to ask for this (or a similar) feature. I'm happy to keep this patch for my own deployment only :)

Issue #3247, opened 2024-04-06
Allow user to activate / deactivate X-ray rule


dtslvr:
An X-ray rule could be irrelevant for the user and they would prefer to deactivate it.

<img width="1181" alt="rules-with-menu" src="https://github.com/ghostfolio/ghostfolio/assets/4159106/efcf98b1-7778-4eab-8086-bb8b764e137a">

### Tasks

- [ ] Add a menu to toggle (_Activate_ / _Deactivate_) a rule
- [ ] Make a request `PUT api/v1/user/setting` with the payload:
`{ rule.[key]: { isActive: true / false } }`
e.g. `{ "rule.EmergencyFundSetup": { "isActive": true } }`
- [ ] Set `isActive` dynamically in rules based on user settings (see [AccountClusterRiskCurrentInvestment](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/models/rules/account-cluster-risk/current-investment.ts#L81))

Sonlis:
I'll have a look at it @dtslvr

PR #3537, opened 2024-07-01
Add option to disable x-ray rules

Sonlis:
Closes https://github.com/ghostfolio/ghostfolio/issues/3247

Give the user the ability to disable x-ray-rules. Store in the user config an object for each rule, following the format:
```
ruleName: { isActive: boolean }
```
Frontend does a PUT request to `api/v1/user/setting`. The endpoint on the backend had to be modified to do a deepmerge instead of a shallow merge, as the rules are stored in a nested array. Doing a shallow merge would replace all the rules by the only one being modified.


apps/client/src/app/components/rule/rule.component.ts
```diff
@@ -16,8 +21,39 @@ import {
...
+    };
+    this.dataService
+      .putUserSetting(settings)
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe(() => {
+        this.dataService
+          .fetchPortfolioReport()
+          .pipe(takeUntil(this.unsubscribeSubject))
+          .subscribe((report) => {
+            for (const ruleGroup in report.rules) {
```
Sonlis, line 47: I wonder if it would make more sense to have the data update in the `rules` files instead of here, so here just emit an event an `rules` fetches back the portfolio report, so it updates all rules instead of going through this for loop.

Sonlis: Also, querying again the `report` endpoint might not be a good idea, as for some reason that I need to uncover, the response time varies between 100ms and 1 minute. Storing the `isActive` parameter in the server reponse's JSON could be a solution, so we don't need to query the backend when toggling the rule, but it means every rule is evaluated, active or not, when the report is requested.

dtslvr: > I wonder if it would make more sense to have the data update in the `rules` files instead of here, so here just emit an event.

Yes, that would be much better. `RuleComponent` (as well as `RulesComponent`) should be "dumb" and just use `@Input()` / `@Output`.

→ `FirePageComponent` should refetch the portfolio report (while showing the loading state)

apps/client/src/app/components/rule/rule.component.html
```diff
@@ -44,7 +50,34 @@
     } @else {
       <div class="flex-grow-1">
         <div class="h6 my-1">{{ rule?.name }}</div>
-        <div class="evaluation">{{ rule?.evaluation }}</div>
+        <div class="evaluation">
+          @if (rule?.evaluation) {
+            {{ rule?.evaluation }}
+          } @else {
+            Rule is disabled
```
dtslvr, line 57: Please remove the `@else` part completely

dtslvr (changes requested):
Thanks for the initial setup, @Sonlis. The logic works pretty well! There is still a bit of potential for improvement in terms of general implementation. Please check my comments.

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -149,6 +150,29 @@ export class FirePageComponent implements OnDestroy, OnInit {
...
+    this.dataService
+      .putUserSetting(event)
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe(() => {
+        this.dataService
+          .fetchPortfolioReport()
+          .pipe(takeUntil(this.unsubscribeSubject))
+          .subscribe((portfolioReport) => {
+            this.accountClusterRiskRules =
+              portfolioReport.rules['accountClusterRisk'] || null;
```
Sonlis, line 164: This warning comes out when activating or deactivating a rule https://angular.dev/errors/NG0956. This is because when we update the rule, we fetch and update the DOM for all rules. However the `firePageComponent` doesn't have any clues of which rules belongs to which category, so reloading a single category or rule is impossible. I am trying to find a way to do so.

dtslvr: What happens if you change `track rule` to `track rule.evaluation` [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/rules/rules.component.html#L16)?

If that works, it should finally be `track rule.key`, but this seems to be missing on `PortfolioReportRule`.

Sonlis: Indeed, I wasn't aware of this `track` directive. I will work on adding the `key` so we reload only the correct key.

Sonlis: Actually, `key` is present in `PortfolioReportRule`, so we can reload only the modified rule.

apps/api/src/app/portfolio/rules.service.ts
```diff
@@ -12,21 +12,26 @@ export class RulesService {
...
           rule.getSettings(aUserSettings)
         );
 
         return {
           evaluation,
-          value,
+          isActive: true,
+          key: rule.getKey(),
+          name: rule.getName(),
+          value
```
dtslvr, line 26: Please move up where it was before⬆️ (short form has priority 😉)

apps/api/src/app/user/user.controller.ts
```diff
@@ -24,7 +24,7 @@ import { JwtService } from '@nestjs/jwt';
 import { AuthGuard } from '@nestjs/passport';
 import { User as UserModel } from '@prisma/client';
 import { StatusCodes, getReasonPhrase } from 'http-status-codes';
-import { size } from 'lodash';
+import { size, merge } from 'lodash';
```
dtslvr, line 27: `{ merge, size }`

apps/api/src/app/user/user.service.ts
```diff
@@ -197,6 +197,18 @@ export class UserService {
       (user.Settings.settings as UserSettings).viewMode = 'DEFAULT';
     }
 
+    // Set default values for x-ray rules
```
dtslvr, line 200: `X-ray`

apps/client/src/app/components/rule/rule.component.html
```diff
@@ -44,7 +47,30 @@
...
+          class="mx-1 no-min-width px-2"
+          mat-button
+          [matMenuTriggerFor]="accountMenu"
+          (click)="$event.stopPropagation()"
+        >
+          <ion-icon name="ellipsis-horizontal" />
+        </button>
+        <mat-menu #accountMenu="matMenu" xPosition="before">
+          <button mat-menu-item (click)="onUpdateRule(rule)">
+            @if (rule?.evaluation) {
```
dtslvr, line 67: Could you check for `rule?.isActive`?

apps/client/src/app/components/rule/rule.component.ts
```diff
@@ -16,8 +19,18 @@ import {
 export class RuleComponent implements OnInit {
   @Input() isLoading: boolean;
   @Input() rule: PortfolioReportRule;
+  @Output() ruleUpdated = new EventEmitter<UpdateUserSettingDto>();
```
dtslvr, line 22: Please add a line break between `@Input()` and `@Output`

apps/client/src/app/components/rules/rules.component.ts
```diff
@@ -11,6 +18,11 @@ import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
 export class RulesComponent {
   @Input() hasPermissionToCreateOrder: boolean;
   @Input() rules: PortfolioReportRule[];
+  @Output() rulesUpdated = new EventEmitter<UpdateUserSettingDto>();
```
dtslvr, line 21: Please add a line break between `@Input()` and `@Output`

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -149,6 +150,29 @@ export class FirePageComponent implements OnDestroy, OnInit {
       });
   }
 
+  public onRulesUpdated(event: UpdateUserSettingDto) {
+    this.isLoading = true;
+    this.dataService
+      .putUserSetting(event)
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe(() => {
+        this.dataService
```
dtslvr, line 159: Could you extract this in a separate function `initializePortfolioReport()` and reuse it as well in `ngOnInit()`? You could use a dedicated variable `isLoadingPortfolioReport`...

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -163,6 +166,7 @@ <h4 class="align-items-center d-flex m-0">
         <gf-rules
           [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
           [rules]="feeRules"
+          (rulesUpdated)="onRulesUpdated($event)"
         />
       </div>
     </div>
```
dtslvr, line 172: Could you add a new section here showing all inactive rules: `<span i18n>Inactive</span>`

dtslvr (changes requested):
Thanks @Sonlis! I have added some more feedback, let me know if you have any question.

apps/client/src/app/components/rule/rule.component.html
Sonlis: Oopsie, forgot that one

apps/client/src/app/pages/portfolio/fire/fire-page.html
Sonlis: So not showing them in the report, and instead having them at the bottom in the Inactive section?

dtslvr: Exactly. This is the starting point for further personalization options in the future.

Sonlis: How would the Inactive section look like? Having each rules that are inactive listed one by one, or separated by their categories ?

dtslvr: One by one to keep it more compact.

Sonlis:
One problem remaining, these 3 lines:
```
      @if (rules?.length === 0) {
        <gf-rule [isLoading]="true" />
      }
```
here: https://github.com/ghostfolio/ghostfolio/blob/40de0cead5abc58c57bcf3c4549c03999db81e9c/apps/client/src/app/components/rules/rules.component.html#L12 shows a constant loading for rules categories that have no rules due to the inactive filtering here: https://github.com/ghostfolio/ghostfolio/pull/3537/files#diff-142ba9596be0882e7f457d6d48d2bbe8ab423202926e1cd4accd56973e7856c3R114. I am still trying to figure out what would be the best way to tackle this

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -160,10 +169,28 @@ <h4 class="align-items-center d-flex m-0">
...
-        />
+        @if (!isLoadingPortfolioReport) {
+          <gf-rules
+            [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
+            [rules]="feeRules"
+            (rulesUpdated)="onRulesUpdated($event)"
+          />
+        }
+      </div>
+      <div>
```
dtslvr, line 180: Would wrapping this `div` solve the issue described [here](https://github.com/ghostfolio/ghostfolio/pull/3537#issuecomment-2278197002)?

```
@if (inactiveRules?.length > 0) {
  <div>
    ...
  </div>
}
```

As it is the last block, I think it is not a big problem if it is suddenly visible (without skeleton).

apps/client/src/app/components/rule/rule.component.html
```diff
@@ -44,7 +47,30 @@
     } @else {
       <div class="flex-grow-1">
         <div class="h6 my-1">{{ rule?.name }}</div>
-        <div class="evaluation">{{ rule?.evaluation }}</div>
```
dtslvr, line 47: I would revert to this line as it is easier to read

apps/client/src/app/components/rule/rule.component.html
```diff
@@ -44,7 +47,30 @@
...
+          @if (rule?.evaluation) {
+            {{ rule?.evaluation }}
+          }
+        </div>
+      </div>
+      <div>
+        <button
+          class="mx-1 no-min-width px-2"
+          mat-button
+          [matMenuTriggerFor]="accountMenu"
```
dtslvr, line 60: `[matMenuTriggerFor]="rulesMenu"`

apps/client/src/app/components/rule/rule.component.html
```diff
@@ -44,7 +47,30 @@
...
+          (click)="$event.stopPropagation()"
+        >
+          <ion-icon name="ellipsis-horizontal" />
+        </button>
+        <mat-menu #accountMenu="matMenu" xPosition="before">
```
dtslvr, line 65: `#rulesMenu="matMenu"`

apps/client/src/app/components/rules/rules.component.html
```diff
@@ -12,9 +12,10 @@
       @if (rules?.length === 0) {
         <gf-rule [isLoading]="true" />
       }
+
       @if (rules !== null && rules !== undefined) {
-        @for (rule of rules; track rule) {
-          <gf-rule [rule]="rule" />
+        @for (rule of rules; track rule.key) {
+          <gf-rule [rule]="rule" (ruleUpdated)="onRulesUpdated($event)" />
```
dtslvr, line 18: Rename to `onRuleUpdated($event)` ❓

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -110,6 +103,32 @@ export class FirePageComponent implements OnDestroy, OnInit {
...
+            (rule) => rule.isActive
+          ) || null;
+        this.emergencyFundRules =
+          portfolioReport.rules['emergencyFund'].filter(
+            (rule) => rule.isActive
+          ) || null;
+        this.feeRules =
+          portfolioReport.rules['fees'].filter((rule) => rule.isActive) || null;
+        this.isLoadingPortfolioReport = false;
+        this.changeDetectorRef.markForCheck();
```
dtslvr, line 128: Add line break

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -110,6 +103,32 @@ export class FirePageComponent implements OnDestroy, OnInit {
       });
   }
 
+  public initializePortfolioReport() {
+    this.isLoadingPortfolioReport = true;
```
dtslvr, line 107: Add line break

apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
```diff
@@ -149,6 +168,16 @@ export class FirePageComponent implements OnDestroy, OnInit {
       });
   }
 
+  public onRulesUpdated(event: UpdateUserSettingDto) {
+    this.isLoading = true;
```
dtslvr, line 172: Add line break

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -124,10 +124,13 @@ <h4 class="align-items-center d-flex m-0">
             <gf-premium-indicator class="ml-1" />
           }
         </h4>
-        <gf-rules
-          [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
-          [rules]="emergencyFundRules"
-        />
+        @if (!isLoadingPortfolioReport) {
```
dtslvr, line 127: Is this really needed? It won't show the skeleton loader with this condition 🤔

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -136,10 +139,13 @@ <h4 class="align-items-center d-flex m-0">
             <gf-premium-indicator class="ml-1" />
           }
         </h4>
-        <gf-rules
-          [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
-          [rules]="currencyClusterRiskRules"
-        />
+        @if (!isLoadingPortfolioReport) {
```
dtslvr, line 142: Is this really needed? It won't show the skeleton loader with this condition 🤔

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -148,10 +154,13 @@ <h4 class="align-items-center d-flex m-0">
             <gf-premium-indicator class="ml-1" />
           }
         </h4>
-        <gf-rules
-          [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
-          [rules]="accountClusterRiskRules"
-        />
+        @if (!isLoadingPortfolioReport) {
```
dtslvr, line 157: Is this really needed? It won't show the skeleton loader with this condition 🤔

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -160,10 +169,28 @@ <h4 class="align-items-center d-flex m-0">
             <gf-premium-indicator class="ml-1" />
           }
         </h4>
-        <gf-rules
-          [hasPermissionToCreateOrder]="hasPermissionToCreateOrder"
-          [rules]="feeRules"
-        />
+        @if (!isLoadingPortfolioReport) {
```
dtslvr, line 172: Is this really needed? It won't show the skeleton loader with this condition 🤔

apps/client/src/app/pages/portfolio/fire/fire-page.html
```diff
@@ -160,10 +169,28 @@ <h4 class="align-items-center d-flex m-0">
...
+        }
+      </div>
+      <div>
+        <h4 class="align-items-center d-flex m-0">
+          <span i18n>Inactive</span>
+          @if (user?.subscription?.type === 'Basic') {
+            <gf-premium-indicator class="ml-1" />
+          }
+        </h4>
+        @if (!isLoadingPortfolioReport) {
```
dtslvr, line 187: Is this really needed?

dtslvr (changes requested):
Welcome back and thanks for the changes @Sonlis. It's very close now! 😎

apps/client/src/app/pages/portfolio/fire/fire-page.html
Sonlis: It wouldn't fix it, I was talking about the other categories. For example `Emergency Fund` is in constant loading if none of its rules are enabled, because of this: https://github.com/ghostfolio/ghostfolio/blob/2893d71377073ad6c53acd0230e5b0f9a2cad9d4/apps/client/src/app/components/rules/rules.component.html#L12.
I think a solution would be to check if the object containing the rules is of length 0 instead. The endpoint anyways return always the result for all rules in 1 response, so checking the individual length of rules to see if they are still loading is useless.

dtslvr: I see. Maybe the rules per group could be initialized as `rules: string[] = undefined;`

Then check like this:

```
@if (!rules) {
  <gf-rule [isLoading]="true" />
} @else if (rules.length > 0) {
  @for (rule of rules; track rule) {
    <gf-rule [rule]="rule" />
  }
}
```

What do you think?

Sonlis: I still think it is useless to have a check per rule set, because the backend returns all of them at the same time anyways. I'll push a commit that do it your way, but I think the way is to set this `isLoading` in the fire page, for all rules at once.

Sonlis: I actually pushed a version that seems optimal, let me know what you think. The `initialiazePortfolioReport` lets the children component know when the report has finished loading.
Last thing is that if the `inactive` section does not have the same top padding compared to the other rules section, trying to figure out why.

dtslvr (approved):
Thank you for successfully completing this contribution, @Sonlis! 🙌🏻
Ready to merge now.

PR #3539, opened 2024-07-03
Fix JSON scraper content-type detection

Summon528:
Header like `content-type: application/json; charset=utf-8` is a valid content-type header but will fall into the non-json codepath currently. Update the checking logic to accommodate headers like this


dtslvr (approved):
Thank you for this improvement, @Summon528! 👍🏻

PR #3614, opened 2024-07-27
ci: harden security following OWASP best practices

rare-magma:
This PR hardens the security of the container by
- changing the default user in the container to a non root one
- dropping all capabilities and adding only the ones that are required
- setting the filesystem as read only

For more info see https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-3-limit-capabilities-grant-only-specific-capabilities-needed-by-a-container and https://cheatsheetseries.owasp.org/cheatsheets/NodeJS_Docker_Cheat_Sheet.html


dtslvr (changes requested):
Hi @rare-magma,

Thank you for taking the time to create this pull request!

First of all, I think it would be best to handle the security hardening and the health check changes independently. I'm not entirely sure about the health check at the moment as it seems to be a breaking change.

Could you please focus on the security hardening first and remove the health check changes from this PR? Thank you.

rare-magma:
hi @dtslvr, that makes sense, I've split the healthcheck change into a separate PR: https://github.com/ghostfolio/ghostfolio/pull/3621

docker/entrypoint.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 set -ex
-
+export NODE_ENV=production
```
dtslvr, line 4: Why do you need to set it in `Dockerfile` and here in `entrypoint.sh`?

Dockerfile
```diff
@@ -11,13 +11,13 @@ COPY ./package.json package.json
...
 COPY ./prisma/schema.prisma prisma/schema.prisma
 
-RUN apt update && apt install -y \
-  g++ \
-  git \
-  make \
-  openssl \
-  python3 \
-  && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y --no-install-suggests \
```
dtslvr, line 14: Why changing `apt` to `apt-get`?

Dockerfile
```diff
@@ -11,13 +11,13 @@ COPY ./package.json package.json
...
+    g++ \
```
dtslvr, line 15: Can you keep the indentation?

Dockerfile
```diff
@@ -50,16 +50,17 @@ RUN npm run database:generate-typings
...
-
 LABEL org.opencontainers.image.source="https://github.com/ghostfolio/ghostfolio"
+ENV NODE_ENV=production
 
-RUN apt update && apt install -y \
-  curl \
-  openssl \
-  && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y --no-install-suggests \
+    curl \
```
dtslvr, line 57: Can you keep the indentation?

docker/docker-compose.yml
```diff
@@ -1,7 +1,19 @@
...
   ghostfolio:
-    image: ghostfolio/ghostfolio:latest
+    image: docker.io/ghostfolio/ghostfolio/ghostfolio:latest
     init: true
+    read_only: true
+    cap_drop:
+      - ALL
+    security_opt:
+      - no-new-privileges:true
+    # uncomment to set resource usage limits
```
dtslvr, line 10: Can you remove the resource usage limits? Anyone who needs it can configure it themselves.

docker/docker-compose.yml
```diff
@@ -22,7 +34,24 @@ services:
...
+      - ALL
+    cap_add:
+      - CHOWN
+      - DAC_READ_SEARCH
+      - FOWNER
+      - SETGID
+      - SETUID
+    security_opt:
+      - no-new-privileges:true
+    # uncomment to set resource usage limits
```
dtslvr, line 48: Can you remove the resource usage limits? Anyone who needs it can configure it themselves.

docker/docker-compose.yml
```diff
@@ -32,8 +61,22 @@ services:
...
   redis:
-    image: redis:alpine
+    image: docker.io/library/redis:alpine
+    cap_drop:
+      - ALL
+    cap_add:
+      - DAC_OVERRIDE
+    security_opt:
+      - no-new-privileges:true
+    # uncomment to set resource usage limits
```
dtslvr, line 73: Can you remove the resource usage limits? Anyone who needs it can configure it themselves.

dtslvr (changes requested):
Thank you for splitting the issue. I have added some comments / questions...

docker/entrypoint.sh
rare-magma: looks like it's redundant, I've removed it in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/482c99144e3916e707417ab4957d130f057f7a91

Dockerfile
rare-magma: @dtslvr I've replaced `apt` with `apt-get` due to the following lint warning: https://github.com/hadolint/hadolint/wiki/DL3027

docker/docker-compose.yml
rare-magma: I've added some examples on resource usage limits because it's recommended to set them in order to prevent DoS attacks (see https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts), these can be difficult to adjust because of different usage levels of the services so I left them commented out. Happy to remove them if you think it doesn't add value

Dockerfile
dtslvr: Okay! 👍🏻

docker/docker-compose.yml
dtslvr: Yes, I would leave it out for the sake of simplicity. Anyone can tweak this file as they wish.

docker/docker-compose.yml
```diff
@@ -22,7 +34,24 @@ services:
       timeout: 5s
       retries: 5
   postgres:
```
dtslvr, line 36: Add line break

docker/docker-compose.yml
```diff
@@ -32,8 +61,22 @@ services:
       retries: 5
     volumes:
       - postgres:/var/lib/postgresql/data
+
   redis:
-    image: redis:alpine
+    image: docker.io/library/redis:alpine
+    # cap_drop:
```
dtslvr, line 67: Why with comments? Either add it or leave it out 🙂

docker/docker-compose.yml
```diff
@@ -22,7 +34,24 @@ services:
...
+    cap_drop:
+      - ALL
+    cap_add:
+      - CHOWN
+      - DAC_READ_SEARCH
+      - FOWNER
+      - SETGID
+      - SETUID
+    security_opt:
+      - no-new-privileges:true
```
dtslvr, line 47: `- no-new-privileges: true` (with space)

docker/docker-compose.yml
```diff
@@ -32,8 +61,22 @@ services:
...
+
   redis:
-    image: redis:alpine
+    image: docker.io/library/redis:alpine
+    # cap_drop:
+    #   - ALL
+    # cap_add:
+    #   - DAC_OVERRIDE
+    security_opt:
+      - no-new-privileges:true
```
dtslvr, line 72: `- no-new-privileges: true` (with space)

dtslvr (changes requested):
Is `NODE_ENV: production` still needed?

* https://github.com/ghostfolio/ghostfolio/blob/main/docker/docker-compose.build.yml#L9
* https://github.com/ghostfolio/ghostfolio/blob/main/docker/docker-compose.yml#L9

Dockerfile
rare-magma: fixed in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/fd3823a70dcdb1c32e43478b955bbb2fc66ef6b3

Dockerfile
rare-magma: fixed in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/fd3823a70dcdb1c32e43478b955bbb2fc66ef6b3

docker/docker-compose.yml
rare-magma: @dtslvr unfortunately this particular field is not compatible with that style, similar to the volumes field: `- postgres:/var/lib/postgresql/data`

docker/docker-compose.yml
rare-magma: added in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/75a0cb301e886645838ebac32c9ebafa4c58b6a1

docker/docker-compose.yml
rare-magma: Removed in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/75a0cb301e886645838ebac32c9ebafa4c58b6a1

docker/docker-compose.yml
rare-magma: see comment above

rare-magma:
@dtslvr indeed `NODE_ENV` in docker compose is redundant now, I've removed it in https://github.com/ghostfolio/ghostfolio/pull/3614/commits/75a0cb301e886645838ebac32c9ebafa4c58b6a1

dtslvr (approved):
Thank you for the contribution @rare-magma! 😎
Ready to merge now.

PR #3622, opened 2024-07-30
Fixed broken command

KernelTurtle:
i have edited the `docker compose` to `docker-compose` as the prior doesn't work


dtslvr:
Hi @KernelTurtle

Thank you for your pull request. The recommended command-line syntax is `docker compose` as per the official Docker Docs: [What are the differences between Compose V1 and Compose V2?](https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2)

Please upgrade to _Compose V2_. Thanks for your understanding!

PR #3642, opened 2024-08-04
fix: image size regression

rare-magma:
In https://github.com/ghostfolio/ghostfolio/pull/3614/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R63 a new layer was added to the image unintentionally. This increased the docker image size from ~700MB to ~1.2GB. I've removed this redundant layer.
I've also noticed that prisma is including a deprecated engine for this docker image, see: https://www.prisma.io/docs/orm/reference/prisma-schema-reference#linux-debian-x86_64


rare-magma:
I've also noticed that the `node_modules` folder in the prod build of the api includes a bunch of dev dependencies because the `nx`'s generated `package.json` includes the `@nx/webpack` package which seems unnecessary. I'm not sure how to remove it since I'm not familiar with `nx`

rare-magma:
If the @nx/webpack dependency is removed the image size is further reduced to ~525MB (vs ~700MB) and no dev dependency is included in it

Dockerfile
```diff
@@ -3,6 +3,14 @@ FROM --platform=$BUILDPLATFORM node:20-slim AS builder
 # Build application and add additional files
 WORKDIR /ghostfolio
 
+RUN apt-get update && apt-get install -y --no-install-suggests \
```
rare-magma, line 6: I've moved the package installation step because it's less likely to change so the cache will be invalidated less frequently

Dockerfile
```diff
@@ -36,6 +37,8 @@ RUN npm run build:production
 
 # Prepare the dist image with additional node_modules
 WORKDIR /ghostfolio/dist/apps/api
+# Exclude unnecessary dev deps
+RUN sed -i '/@nx/d' package.json
```
dtslvr, line 41: Could we use `npm install --omit=dev` to prevent the installation of `devDependencies`?
Source: https://stackoverflow.com/a/9276112/744012

dtslvr (changes requested):

Dockerfile
rare-magma: @dtslvr in this case the generated package.json includes `@nx/webpack` in the production dependencies so I think the `--omit=dev` won't prevent it from being installed

rare-magma: @dtslvr I excluded the webpack file in the nx project options in https://github.com/ghostfolio/ghostfolio/pull/3642/commits/12e4eb5b9ea64e0e93a7c11ea15be63c5ebfa63e and it seemed to fix the problem

dtslvr (approved):
Thank you for the changes. I will try it.

dtslvr:
I think it looks fine @rare-magma.
https://hub.docker.com/layers/ghostfolio/ghostfolio/2.103.0-alpha/images/sha256-8a79da488f1beb06a18c44364a3a421fc610051940565e009b8084cd4b70e4fb

PR #3661, opened 2024-08-11
Feature/migrate from Angular Material design 2 to 3

dtslvr:


dtslvr:
@KenTandrian I have started to collect sub tasks.

KenTandrian:
@dtslvr for task 2 -- in the documentation, the FAB color in **dark mode** looks like this:

FAB

with darker background and light-colored icon. Do you think we should follow that or restore to primary color (light background with dark icon)?

KenTandrian:
Task 03 can be done by adding this to `.theme-dark` in `theme.scss`:
```scss
@include mat.table-overrides(
  (
    row-item-outline-color: rgba(255, 255, 255, 0.12)
  )
);
```

dtslvr:
> @dtslvr for task 2 -- in the documentation, the FAB color in **dark mode** looks like this:
> 
> FAB
>
> with darker background and light-colored icon. Do you think we should follow that or restore to primary color (light background with dark icon)?

My personal preference is to stick with a more monochrome color palette, using our primary brand color and a consistent range of gray tones. For that reason, I would restore the FAB to a primary color with a dark-colored icon (in dark mode).

What are your thoughts on this?

KenTandrian:
I don't have a preference. FAB with primary color looks good.

apps/client/src/app/components/header/header.component.html
```diff
@@ -15,9 +15,9 @@
     <ul class="alig-items-center d-flex list-inline m-0 px-2">
```
KenTandrian: @dtslvr typo on `alig-items-center`?

dtslvr: Good catch! 🧐

I have prepared a fix on the `main` branch: https://github.com/ghostfolio/ghostfolio/pull/5317

dtslvr:
I have moved the sub tasks to https://github.com/ghostfolio/ghostfolio/issues/5316.

PR #3708, opened 2024-08-29
Extract tags selector to @ghostfolio/ui

dinirichard:


libs/ui/src/lib/tags-selector/tags-selector.component.ts
```diff
@@ -0,0 +1,106 @@
...
+    this.tagsSignal()
+      ? this.tagsSignal.update((tags) => [...tags, newTag])
+      : this.tagsSignal.update(() => [newTag]);
+    this.currentTag.set('');
+    event.option.deselect();
+  }
+
+  public onRemoveTag(aTag: Tag) {
+    this.tagsSignal.update((tagsSignal) => {
+      const index = tagsSignal.indexOf(aTag);
```
dtslvr, line 97: What about this?

```
return tagsSignal.filter((tag) => { return tag !== aTag });
```

dtslvr (changes requested):
Thanks for your contribution, @dinirichard 🙂
I have added a few comments.

dtslvr:
Following up on the code review. Let me know if you need anything, @dinirichard!

PR #3756, opened 2024-09-12
Bugfix/fix destructuring in activities filters

dtslvr:


dtslvr:
Thanks for the report, @x1c0.

PR #3827, opened 2024-09-27
Feature/improve performance of portfolio snapshot computation

ceroma:
After doing some perf analysis on the portfolio snapshot computation, this line turned out to be the biggest offender on my machine (Raspberry Pi 4).

My current DB consists of 5189 activities and 478 different symbols. The majority (about 400) of the symbols represent single `INTEREST` data points.

This line runs for each `SymbolProfile`, and deep-cloning is very expensive. Specially if most of the time (in my case) it ends up filtering out just one element out of the entire clone. On my machine, this line averaged 280ms per invocation, amounting to over 2 minutes during a snapshot computation.

After this diff, this line averaged 1.5ms per symbol, with a peak of 56ms for my symbol with the highest number of activities (1500). The total time spent filtering and deep-cloning is now less than a second (~750ms).

Per my understanding of the code, this change should preserve the same behavior as before, but I'm not too familiar with TypeScript so please feel free to be very pedantic in the review.

Test Plan:
```
```
```
```
4. Apply this patch
5. Repeat steps 1-3
```
2c2
<   "expiration": 1727423725104, 
---
>   "expiration": 1727424454187,
```


dtslvr (approved):
Great finding, @ceroma! 🤩
Thanks for your contribution and putting together the in-depth analysis.

PR #3829, opened 2024-09-27
Feature/improve performance of portfolio snapshot computation - 2

ceroma:
Continuing my perf analysis of the portfolio snapshot computation, this was another big time sink on my machine, specially when generating each day for the `'5y'` and `'max'` date ranges.

For reference, before this diff, the `dateRange` loop was taking on average ~100ms per symbol for me (Raspberry Pi 4, 478 symbols in the DB). After this diff, it takes ~10ms per symbol, shaving off 40s of computation time.

Again, I'm not too familiar with TypeScript so please feel free to be very pedantic in the review.

Test Plan:
```
```
```
```
4. Apply this patch
5. Repeat steps 1-3
```
2c2
<   "expiration": 1727449402720,
---
>   "expiration": 1727449639794,
```

### DIFF
--- a/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
+++ b/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
@@ -16,13 +16,14 @@ import {
   addDays,
   addMilliseconds,
   differenceInDays,
-  eachDayOfInterval,
   format,
   isBefore
 } from 'date-fns';
 import { cloneDeep, first, last, sortBy } from 'lodash';
 
 export class TWRPortfolioCalculator extends PortfolioCalculator {
+  private chartDatesDescending: string[];
+
   protected calculateOverallPerformance(
     positions: TimelinePosition[]
   ): PortfolioSnapshot {
@@ -820,31 +821,35 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
         startDate = start;
       }
 
+      const startDateString = format(startDate, DATE_FORMAT);
+      const endDateString = format(endDate, DATE_FORMAT);
+
       const currentValuesAtDateRangeStartWithCurrencyEffect =
-        currentValuesWithCurrencyEffect[format(startDate, DATE_FORMAT)] ??
-        new Big(0);
+        currentValuesWithCurrencyEffect[startDateString] ?? new Big(0);
 
       const investmentValuesAccumulatedAtStartDateWithCurrencyEffect =
-        investmentValuesAccumulatedWithCurrencyEffect[
-          format(startDate, DATE_FORMAT)
-        ] ?? new Big(0);
+        investmentValuesAccumulatedWithCurrencyEffect[startDateString] ??
+        new Big(0);
 
       const grossPerformanceAtDateRangeStartWithCurrencyEffect =
         currentValuesAtDateRangeStartWithCurrencyEffect.minus(
           investmentValuesAccumulatedAtStartDateWithCurrencyEffect
         );
 
-      const dates = eachDayOfInterval({
-        end: endDate,
-        start: startDate
-      }).map((date) => {
-        return format(date, DATE_FORMAT);
-      });
-
       let average = new Big(0);
       let dayCount = 0;
 
-      for (const date of dates) {
+      if (!this.chartDatesDescending) {
+        this.chartDatesDescending = Object.keys(chartDateMap).sort().reverse();
+      }
+
+      for (const date of this.chartDatesDescending) {
+        if (date > endDateString) {
+          continue;
+        } else if (date < startDateString) {
+          break;
+        }
+
         if (
           investmentValuesAccumulatedWithCurrencyEffect[date] instanceof Big &&
           investmentValuesAccumulatedWithCurrencyEffect[date].gt(0)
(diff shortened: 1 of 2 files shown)


apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -23,6 +31,47 @@ import {
 import { cloneDeep, first, last, sortBy } from 'lodash';
 
 export class TWRPortfolioCalculator extends PortfolioCalculator {
+  private cachedEachDayOfInterval: { [range: string]: string[] };
```
dtslvr, line 34: When you initialize here, you could get rid of the whole constructor:
```
private cachedEachDayOfInterval: { [range: string]: string[] } = {};
```

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -834,12 +883,16 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
...
+          end: endDate,
+          start: startDate
+        }).map((date) => {
+          return format(date, DATE_FORMAT);
+        });
+        this.cachedEachDayOfInterval[dateRange] = dates;
+      }
 
       let average = new Big(0);
       let dayCount = 0;
```
dtslvr, line 898: That's an interesting finding! 👍🏻

There might even be a superior optimization by just replacing

```
for (const date of dates) {
```
with
```
for (const date of Object.keys(chartDateMap)) {
```
below (and remove `const dates = ...` completely).

dtslvr (changes requested):
Many thanks for this in-depth analysis, @ceroma! 👍🏻

I have added a few comments. Can you please try the impact of the suggestion (change loop from `dates` to `Object.keys(chartDateMap)` in `getSymbolMetrics()`)?

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
ceroma: I took a quick look at doing that, and I think I have some questions. I don't fully grasp everything that the portfolio calculator is doing, so I might be missing some bigger picture context here, but wouldn't that proposed change result in different values in the end?

For instance, when the outer loop is at `dateRage = '1d'`, the size of `dates` will be 2, right? But if we use `for (const date of Object.keys(chartDateMap) { ... }`, then it will go through way more days. Wouldn't this significantly affect the resulting values?

One other option here that I just thought of is to do something like the pseudocode below instead of creating/caching the dates:
```
currentDate = startDate;
while (currentDate < endDate) {
  ...
  currentDate += 1;
}
```

dtslvr: Which outer loop do you mean? I don't think the date range matters here. The portfolio calculator always calculates the snapshot for `max` while `chartDateMap` includes all relevant dates.

ceroma: I meant this loop:
```
for (const dateRange of ['1d', '1y', '5y', 'max', ...]) {
  ...
  const dates = eachDayOfInterval(...);
  ...
  for (const date of dates) {
    ...
  }
}
```


> I would expect that the helper function eachDayOfInterval is already doing something very similar. No?

Yeah, but I think the slowdown, at least in my case, is due to `eachDayOfInterval` allocating big arrays for the `'5y'` and `'max'` (11 years for me) ranges. Manually incrementing the days would avoid that overhead.

dtslvr: > Wouldn't this significantly affect the resulting values?

I agree! So your caching approach is better. However, we could additionally filter the result of `eachDayOfInterval()` by checking if the date is relevant (part of `chartDateMap`) to further optimize. What do you think?

ceroma: Here's an example to show what I think I'm failing to understand. Let's say we have these dates and values in `investmentValuesAccumulatedWithCurrencyEffect`:
2024-01-01: 0
2024-01-02: 0
2024-01-03: 0
2024-01-04: 1
2024-01-05: 1

The `for (const date of dates)` loop computes the average for these days, which would be `0.4` considering all dates.

Now if we filter by the dates present in `chartDateMap`, and let's say `chartDateMap` only has 2 of those dates, the first and the last. Then the average would result in `0.5`.

Would that be ok?

dtslvr: > Now if we filter by the dates present in chartDateMap, and let's say chartDateMap only has 2 of those dates, the first and the last. Then the average would result in 0.5.

`chartDateMap` includes all dates where you have an activity. So you will get the dates where the investment value actually changes. Not quite sure if it makes a difference, though.

ceroma: Ah! Since `getChartDateMap` uses `MAX_CHART_ITEMS` I mistakenly assumed it would contain at most `MAX_CHART_ITEMS`. I'll try using `chartDateMap` here and report back.

ceroma: Using `chartDateMap`, this actually reduced the time spent per symbol on this `dateRange` loop by an extra 5ms. So now, on my machine, it's at 9.7ms instead of 100ms.

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -820,31 +821,35 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
...
-        start: startDate
-      }).map((date) => {
-        return format(date, DATE_FORMAT);
-      });
-
       let average = new Big(0);
       let dayCount = 0;
 
-      for (const date of dates) {
+      if (this.chartDatesDescending === undefined) {
```
dtslvr, line 842: `if (!this.chartDatesDescending) {`

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -820,31 +821,35 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
...
+        this.chartDatesDescending = Object.keys(chartDateMap).sort().reverse();
+      }
+      for (const date of this.chartDatesDescending) {
```
dtslvr, line 845: Please add a line break

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -820,31 +821,35 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
...
+        if (date > endDateString) {
+          continue;
+        }
+        if (date < startDateString) {
```
dtslvr, line 849: `} else if (date < startDateString) {`

dtslvr (changes requested):
I have reviewed the code again and added a few suggestions.

dtslvr (approved):
Thanks again for this optimization, @ceroma! 🙌🏻

Ready to merge.

PR #3854, opened 2024-10-01
feat: display message if no search results were found in symbol-autocomplete

### LINKED ISSUE #3779: [Feature] Show message if no results have been found in activity dialog
### Steps to reproduce

* Go to _Portfolio_ > _Activities_
* Click the + button
* Search for `asdlkfj` in the _Name, symbol or ISIN_ input field

Add a synthetic (disabled) option (`mat-option`) to the [symbol autocomplete component](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/symbol-autocomplete) if no results have been found with the following label:

> Oops! Could not find any assets.

dw-0:
Fixes #3779 

## Before:
image

## After:
24-10-01_19-40-38_brave

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added a message if no results have been found in the create or update activity dialog
+- Added a message to the search asset component if no results have been found in the create or update activity dialog
 - Added support to customize the rule thresholds in the _X-ray_ section (experimental)
 
 ### Changed


apps/client/src/locales/messages.de.xlf
```diff
@@ -7258,6 +7258,14 @@
           <context context-type="linenumber">280</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="b225488f8b209e9704760dc9f5d99845a5d07bf6" datatype="html">
+        <source>Oops! Could not find any assets.</source>
+        <target state="translated">Hoppla! Es konnten keine Assets gefunden werden.</target>
```
dtslvr, line 7263: In the whole app we are using _Ups!_, can you change for consistency reasons? 😃

libs/ui/src/lib/symbol-autocomplete/symbol-autocomplete.component.html
```diff
@@ -12,6 +12,11 @@
   (optionSelected)="onUpdateSymbol($event)"
 >
   @if (!isLoading) {
+    @if (!empty && !filteredLookupItems.length) {
```
dtslvr, line 15: Technically this works perfectly. However, I would not show the message for a single character. As the search only provides results for a [query length > 1](https://github.com/ghostfolio/ghostfolio/blob/main/libs/ui/src/lib/symbol-autocomplete/symbol-autocomplete.component.ts#L123), it is a bit annoying.

dtslvr (changes requested):
Thank you for the contribution, @dw-0! It already works, while testing I have found potential improvements.

apps/client/src/locales/messages.de.xlf
dw-0: Oh, yeah of course! Sorry for that oversight.

libs/ui/src/lib/symbol-autocomplete/symbol-autocomplete.component.html
dw-0: You are right.
So for fixing this, i introduced a new variable. The reason behind that is, that checking for the input being larger than 1 character results in the message popping up for the debounce time of 400ms, that looked wrong to me. So now a flag is set after the loading is done, inside the subscribe, which then checks for the result length. That way it works quite nicely.

dtslvr (approved):
I have refactored the code to get rid of the additional variable.

Ready to merge now.

dw-0:
Oh, i haven't had @empty on my radar while writing the code. That's actually quite nice. Will keep that in mind for the future :)

dtslvr:
Thanks again for your contribution, @dw-0! 🖤

PR #3855, opened 2024-10-02
Feature/improve performance of portfolio snapshot computation - 3

ceroma:
This is similar to the previous diff. Here we can speed things up by using the dates in `chatDatesMap` instead of computing the whole interval every time.

For reference, before this diff, the `while (isBefore(...))` loop was taking on average ~88ms per symbol for me (Raspberry Pi 4, 481 symbols in the DB). After this diff, it takes ~12ms per symbol, shaving off almost 40s of total computation time.

Test Plan:
```
```
```
```
4. Apply this patch
5. Repeat steps 1-3
```
2c2
<   "expiration": 1727834561292,
---
>   "expiration": 1727836180162,
```

### DIFF
--- a/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
+++ b/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
@@ -22,7 +22,7 @@ import {
 import { cloneDeep, first, last, sortBy } from 'lodash';
 
 export class TWRPortfolioCalculator extends PortfolioCalculator {
-  private chartDatesDescending: string[];
+  private chartDates: string[];
 
   protected calculateOverallPerformance(
     positions: TimelinePosition[]
@@ -228,11 +228,11 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
 
     const dateOfFirstTransaction = new Date(first(orders).date);
 
-    const unitPriceAtStartDate =
-      marketSymbolMap[format(start, DATE_FORMAT)]?.[symbol];
+    const endDateString = format(end, DATE_FORMAT);
+    const startDateString = format(start, DATE_FORMAT);
 
-    const unitPriceAtEndDate =
-      marketSymbolMap[format(end, DATE_FORMAT)]?.[symbol];
+    const unitPriceAtStartDate = marketSymbolMap[startDateString]?.[symbol];
+    const unitPriceAtEndDate = marketSymbolMap[endDateString]?.[symbol];
 
     if (
       !unitPriceAtEndDate ||
@@ -278,7 +278,7 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
 
     // Add a synthetic order at the start and the end date
     orders.push({
-      date: format(start, DATE_FORMAT),
+      date: startDateString,
       fee: new Big(0),
       feeInBaseCurrency: new Big(0),
       itemType: 'start',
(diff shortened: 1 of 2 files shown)


ceroma:
I'll rebase and update this PR once #3829 merges.

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -227,11 +228,11 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
 
     const dateOfFirstTransaction = new Date(first(orders).date);
 
-    const unitPriceAtStartDate =
-      marketSymbolMap[format(start, DATE_FORMAT)]?.[symbol];
+    const endString = format(end, DATE_FORMAT);
+    const startString = format(start, DATE_FORMAT);
```
ceroma, line 232: I think this would probably be better as `endDateString`/`startDateString`, but those were already used in the other dates-loop at the end of the method. What do you think of renaming those at the end to `range[Start|End]DateString`, and these here at the top to `[end|start]DateString`?

ceroma:
I'll also check the test failure.

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
dtslvr, line 1: Please remove unused `addDays`

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
```diff
@@ -839,14 +844,12 @@ export class TWRPortfolioCalculator extends PortfolioCalculator {
       let average = new Big(0);
       let dayCount = 0;
 
-      if (!this.chartDatesDescending) {
-        this.chartDatesDescending = Object.keys(chartDateMap).sort().reverse();
-      }
+      for (let i = this.chartDates.length - 1; i >= 0; i -= 1) {
```
dtslvr, line 847: What speaks against this?

```
for (const date of this.chartDates.reverse()) {
```

I think it is easier to understand. Drawback: `reverse()` is destructive

dtslvr (changes requested):
Thank you for the further analysis, @ceroma! 🤓
I have added a few comments.

apps/api/src/app/portfolio/calculator/twr/portfolio-calculator.ts
ceroma: This method (`getSymbolMetrics`) is called multiple times during the snapshot computation. So if we reverse the array here, we'd need to reverse it back at the top of the method where we traverse `this.chartDates` in chronological order.

Alternatively, we could cache both `this.chartDates` and `this.chartDatesDescending` (or `this.chartDatesReverse`). The extra memory shouldn't be too bad. What do you think?

dtslvr: Thanks for the explanation. I think another variable does not make sense just for that 🤓 Maybe change to `i--` and add a comment:

```
for (let i = this.chartDates.length - 1; i >= 0; i--) {
  // Loop through chart dates in reverse order
  const date = this.chartDates[i];
  ...
}
```

dtslvr (approved):
Thanks again @ceroma! 👍🏻

Ready to merge now.

Issue #3233, opened 2024-04-02
[BUG] Dividend fees affect performance, but dividends  don't


sander85:
**Bug Description**

When dividend is added with its fees (taxes) the performance of the holding goes down and fees are calculated into change. But dividend itself isn't. This way the holding looks a lot worse than it actually is.

**To Reproduce**

1. Add buying activity for stock
2. Check performance
3. Add dividend activity
4. With each new dividend the performance goes worse.

**Expected behavior**

Dividend fees should not affect performance if dividend itself doesn't.

**Environment**

- Ghostfolio Version 2.67.0
- Self-hosted
- Experimental Features disabled

dtslvr:
Hi @sander85

Thank you for the report. Can you please attach a minimal export file to reproduce this scenario?

sander85:
The change before adding dividends were over 690€ and performance of the profile/holding over 200%. It fell to 196,42%.

PR #3857, opened 2024-10-02
Add realized dividends to gross performance

dandevaud:
Hi @dtslvr 

I saw there was an issue raised with dividends lowering the performance (#3233)
I added logic to handle dividends like sell-order for performance calculation.

I would say from economical perspective this makes most sense, as when dividends are paid the stock price will lower by the dividend amount.

Moreover, I extracted the sellOrder and dividend handling in a separate method for readability

### DIFF
--- a/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator-msft-buy-with-dividend.spec.ts
+++ b/apps/api/src/app/portfolio/calculator/twr/portfolio-calculator-msft-buy-with-dividend.spec.ts
@@ -155,25 +155,25 @@ describe('PortfolioCalculator', () => {
             dividendInBaseCurrency: new Big('0.62'),
             fee: new Big('19'),
             firstBuyDate: '2021-09-16',
-            grossPerformance: new Big('33.25'),
-            grossPerformancePercentage: new Big('0.11136043941322258691'),
+            grossPerformance: new Big('33.87'),
+            grossPerformancePercentage: new Big('0.11343693482483756447'),
             grossPerformancePercentageWithCurrencyEffect: new Big(
-              '0.11136043941322258691'
+              '0.11343693482483756447'
             ),
-            grossPerformanceWithCurrencyEffect: new Big('33.25'),
+            grossPerformanceWithCurrencyEffect: new Big('33.87'),
             investment: new Big('298.58'),
             investmentWithCurrencyEffect: new Big('298.58'),
             marketPrice: 331.83,
             marketPriceInBaseCurrency: 331.83,
-            netPerformance: new Big('14.25'),
-            netPerformancePercentage: new Big('0.04772590260566682296'),
+            netPerformance: new Big('14.87'),
+            netPerformancePercentage: new Big('0.04980239801728180052'),
             netPerformancePercentageWithCurrencyEffectMap: {
-              max: new Big('0.04772590260566682296')
+              max: new Big('0.04980239801728180052')
             },
             netPerformanceWithCurrencyEffectMap: {
               '1d': new Big('-5.39'),
-              '5y': new Big('14.25'),
-              max: new Big('14.25'),
+              '5y': new Big('14.87'),
+              max: new Big('14.87'),
               wtd: new Big('-5.39')
             },
             quantity: new Big('1'),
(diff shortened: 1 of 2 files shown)


dtslvr (commented):
Thank you very much for your contribution, @dandevaud! 🙂

While reviewing, I was surprised that no test had to be adjusted with your PR. I have therefore extended the only dividend test (https://github.com/ghostfolio/ghostfolio/pull/3891). Can you please rebase here?

Ideally, the dividend performance should be listed here separately and also included in the overall performance:

portfolio-analysis

What do you think?

dandevaud:
Thanks @dtslvr 
 
 I updated the teste accordingly.
 
 We can list the dividend separately, but I would already expect it in the "Absolute Asset Performance". 
 But your input had me thinking of something else, the holding detail:
 
image

There the "Change" and "Change with currency effect" should be change to something like "Total Return" and "Total Return with currency effect". 
Change is rather related to price changes and total return is accepted as word for all returns relating to the asset (incl. sell and dividends)

Let me know what you think and if I should adapt the overal performance as well.

dtslvr:
> portfolio-analysis

> We can list the dividend separately, but I would already expect it in the "Absolute Asset Performance".

The _Asset Performance_ label is indeed not an ideal label. What about:

* Capital Gains
  * %
* Dividend Income
  * %
* Currency Exchange Effect
  * %
------
* Total Return
  * %

> There the "Change" and "Change with currency effect" should be change to something like "Total Return" and "Total Return with currency effect".

What about "Total Return" / "Total Return with currency effect" and "Total Return (%)" / "Total Return with currency effect (%)" in the holding detail dialog?

dandevaud:
Hi @dtslvr 

Please excuse my late reply.

Sounds better, but instead of dividend income we should use something like Realized Profit/Loss, which includes sells as well.
But then it gets kind of complicated for the % calculation. 
Do you have an idea?

dandevaud:
Hi @dtslvr 

Again please excuse my late reply, had some busy weeks.

I revisted the theoretical performance calculation, meaning I have not yet checked the code on how to implement it as I wanted your inputs first.

The idea of splitting up the absolut net asset performance good, I would nonetheless go with realized and unrealized gains/losses.
Realized include all gains/losses from sold assets and dividends and unrealized includes the performance of the yet hold assets compared to their purchase price.

I currently face a problem with the relative performance, respectively the cost basis to use.
Assuming the following scenario:
Buying 4 stocks @ 25, it raises to 40 you sell 1, it raises to 50
| | **t1**| **t2** | **t3** |
|--------|--------|--------|--------|
| **Quantity**| 4| 3 | 3 |
| **Price**| 25 | 40 | 50 |
| **PF-Value** | 100 | 120 | 150 |
| **Sold** | - | 40 | - |


Calculating the P/Ls would lead to:

- Unrealized 75 ( current holding * (current price - initial price) --> 3 * (50-25))
- Realized 15 (sold holding * (sell price - initial price) --> 1 * (40-25))
- Total 90 (sum of the above)

For the relative performance we need to have a cost basis to calculate against. 
The simplest would be the total investment, as it bests reflects the total return on ones investment. 
| | **Absolut performance**| **Cost basis**| **relative performance**|
|--------|--------|--------|--------|
| **Unrealized P/L** | 75| 100| 75%|
| **Realized P/L**| 15| 100| 15%|
| **Total**| 90| 100| 90%|


The problem is the unrealized P/L does not correctly refelct the price change of the underlying asset. Whilst the asset price doubled, the unrealized P/L only has a performance of 75%. 

I checked yahoo finance how they handle it and the use different cost basis for the different P/L calculation, but do not show a total performance value. 
For unrealized P/L they use the cost basis of the shares left, in the case above this would be 75 (shares left * initial price --> 3*75). 
| | **Absolut performance**| **Cost basis**| **relative performance**|
|--------|--------|--------|--------|
| **Unrealized P/L** | 75| 75| 100%|
| **Realized P/L**| 15| 100| 15%|
| **Total**| -| -| -|


I would go with the first approach not correctly reflecting the price change, as it accounts for all performance of a portfolio.

Whats your take on this? Or would you even choose a totally different approach?

Many thanks for your feedback and kind regards

dtslvr:
Hi @dandevaud

Thanks for the detailed thoughts!

I don’t have a strong opinion on this yet, but I’m cautious about making changes that might affect how users currently rely on the calculations. One potential approach could be to support interchangeable calculation methods, allowing flexibility while maintaining the current behavior.

Do you think there’s a simpler way to address this, especially regarding dividends?

dpjrodrigues:
Leaving here an example of how getquin handles the dividends in the performace screen. It is included in the total return but have a performance breakdown that shows what is price gain, dividends or realized losses.
image

balgerion:
Any chance for this to be merged ? I would love to migrate from getquin to ghostofolio but this for me is just must have feature

s0129:
I would also love this to be integrated. Maybe add a setting somewhere to enable/disable dividends counting towards performance?

PR #3858, opened 2024-10-02
Fix exchange rate gathering job

dandevaud:
Hi @dtslvr 

I had an issues with one datagathering job as a value was missing.

In the combination GBp <-> GBP and the root currency handling one value was missing ([08.10.2013 for GBP was missing](https://finance.yahoo.com/quote/GBPUSD%3DX/history/?period1=1381017600&period2=1381363200))
This lead to the whole update job to fail, now it checks if a value is present before trying to multiply it with the sub currency factor.

I

### DIFF
--- a/apps/api/src/services/data-provider/data-provider.service.ts
+++ b/apps/api/src/services/data-provider/data-provider.service.ts
@@ -666,12 +666,13 @@ export class DataProviderService {
     } = {};
 
     for (const date in rootData) {
-      if (rootData[date].marketPrice === null) {
-        continue;
+      if (isNumber(rootData[date].marketPrice)) {
+        data[date] = {
+          marketPrice: new Big(factor)
+            .mul(rootData[date].marketPrice)
+            .toNumber()
+        };
       }
-      data[date] = {
-        marketPrice: rootData[date].marketPrice
-      };
     }
 
     return data;
--- a/apps/api/src/services/data-provider/data-provider.service.ts
+++ b/apps/api/src/services/data-provider/data-provider.service.ts
@@ -667,7 +667,9 @@ export class DataProviderService {
 
     for (const date in rootData) {
       data[date] = {
-        marketPrice: new Big(factor).mul(rootData[date].marketPrice).toNumber()
+        marketPrice: rootData[date].marketPrice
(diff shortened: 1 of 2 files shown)


apps/api/src/services/data-provider/data-provider.service.ts
```diff
@@ -667,7 +667,9 @@ export class DataProviderService {
 
     for (const date in rootData) {
       data[date] = {
```
dtslvr, line 669: Would it also work to make the check here (and skip that date)?

apps/api/src/services/data-provider/data-provider.service.ts
```diff
@@ -667,7 +667,9 @@ export class DataProviderService {
...
-        marketPrice: new Big(factor).mul(rootData[date].marketPrice).toNumber()
+        marketPrice: rootData[date].marketPrice
```
dtslvr, line 670: Should be check for `rootData[date].marketPrice >= 0`? Currently, `0` would result in `null`.

apps/api/src/services/data-provider/data-provider.service.ts
dandevaud: You're right, looks even better.
Adapted it accordingly

apps/api/src/services/data-provider/data-provider.service.ts
dandevaud: I assumed if the marketPrice is 0 there might be a problem anyway.
But you are right, I even went a step further allowing negative marketPrices an strictly checking for null values.

dtslvr: Looks like we lost the factor now. Can you check my change please?

dandevaud (commented):
Please excuse I must have deleted too much.
Looks good now

dtslvr (approved):
Thanks for your contribution, @dandevaud! 😎

Ready to merge now.

b-reich:
@dtslvr when can we expect a new release?

dtslvr:
> @dtslvr when can we expect a new release?

@b-reich https://github.com/ghostfolio/ghostfolio/releases/tag/2.113.0

PR #3859, opened 2024-10-02
feat: tags handling improvements

matejgerek:
All components were tested on data created by me.

-----

The handling of tags should be improved by the following changes

* [x]  Remove tags from info endpoint (`GET api/v1/info`)
* [x]  Change tags in user endpoint (`GET api/v1/user`)
   * [x]  Include all available tags (`WHERE userId is NULL or id of current user`)
   * [x]  Add a new flag `isUsed: boolean` to each tag
* [x]  Change `tagsAvailable` to tags of user endpoint in `GfHoldingDetailDialogComponent`
* [x]  Change `tagsAvailable` to tags of user endpoint in `CreateOrUpdateActivityDialog`
* [x]  Filter tags by `isUsed` in `GfAssistantComponent`

### DIFF
--- a/apps/api/src/services/tag/tag.service.ts
+++ b/apps/api/src/services/tag/tag.service.ts
@@ -23,7 +23,11 @@ export class TagService {
         name: 'asc'
       },
       include: {
-        orders: true
+        _count: {
+          select: {
+            orders: true
+          }
+        }
       },
       where: {
         OR: [
@@ -39,8 +43,8 @@ export class TagService {
 
     return tags.map((tag) => ({
       ...tag,
-      isUsed: tag.orders.length > 0,
-      orders: undefined
+      isUsed: tag._count.orders > 0,
+      _count: undefined
     }));
   }
 }
--- a/apps/api/src/services/tag/tag.service.ts
+++ b/apps/api/src/services/tag/tag.service.ts
@@ -6,22 +6,8 @@ import { Injectable } from '@nestjs/common';
 export class TagService {
   public constructor(private readonly prismaService: PrismaService) {}
 
-  public async getPublic() {
-    return this.prismaService.tag.findMany({
-      orderBy: {
-        name: 'asc'
-      },
-      where: {
-        userId: null
-      }
-    });
-  }
-
   public async getTagsForUser(userId: string) {
     const tags = await this.prismaService.tag.findMany({
-      orderBy: {
-        name: 'asc'
-      },
       include: {
         _count: {
           select: {
@@ -32,6 +18,9 @@ export class TagService {
             }
           }
         }
+      },
+      orderBy: {
+        name: 'asc'
       },
       where: {
         OR: [
--- a/apps/api/src/services/tag/tag.service.ts
+++ b/apps/api/src/services/tag/tag.service.ts
@@ -42,9 +42,10 @@ export class TagService {
       }
     });
 
-    return tags.map((tag) => ({
-      ...omit(tag, '_count'),
-      isUsed: tag._count.orders > 0
+    return tags.map(({ _count, id, name }) => ({
+      id,
+      name,
+      isUsed: _count.orders > 0
     }));
   }
 }
--- a/apps/api/src/services/tag/tag.service.ts
+++ b/apps/api/src/services/tag/tag.service.ts
@@ -22,29 +22,18 @@ export class TagService {
       orderBy: {
         name: 'asc'
       },
+      include: {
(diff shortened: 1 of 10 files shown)


matejgerek:
@dtslvr Hi Thomas, the PR is ready for review. I am not able to select reviewers, so I am tagging you here.

apps/api/src/services/tag/tag.service.ts
```diff
@@ -17,18 +18,33 @@ export class TagService {
...
-    return this.prismaService.tag.findMany({
+  public async getTagsForUser(userId: string) {
+    const tags = await this.prismaService.tag.findMany({
       orderBy: {
         name: 'asc'
       },
+      include: {
+        _count: {
+          select: {
+            orders: true
```
dtslvr, line 29: The specification was too unspecific. It needs the flag `isUsed` related to the current user:

```
orders: {
  where: {
     userId
  }
}
```

apps/api/src/services/tag/tag.service.ts
```diff
@@ -17,18 +18,33 @@ export class TagService {
...
+          },
+          {
+            userId: null
           }
-        }
+        ]
       }
     });
+
+    return tags.map((tag) => ({
```
dtslvr, line 45: With this, you don't need `omit()`:

```
return tags.map(({ _count, id, name }) => ({
  id,
  name,
  isUsed: _count.orders > 0
}));
```

libs/ui/src/lib/assistant/assistant.component.ts
```diff
@@ -165,13 +165,15 @@ export class GfAssistantComponent implements OnChanges, OnDestroy, OnInit {
...
     });
-    this.tags = this.user?.tags.map(({ id, name }) => {
-      return {
-        id,
-        label: translate(name),
-        type: 'TAG'
-      };
-    });
+    this.tags = this.user?.tags
+      .filter((tag) => tag.isUsed)
```
dtslvr, line 169: Please use the explicit form here:

```
.filter(({ isUsed }) => { return isUsed; })
```

libs/ui/src/lib/assistant/assistant.component.ts
```diff
@@ -270,6 +272,16 @@ export class GfAssistantComponent implements OnChanges, OnDestroy, OnInit {
       this.filterForm.enable({ emitEvent: false });
     }
 
+    this.tags = this.user?.tags
+      .filter((tag) => tag.isUsed)
```
dtslvr, line 276: Please use the explicit form here:

```
.filter(({ isUsed }) => { return isUsed; })
```

dtslvr (changes requested):
Thank you for the contribution, @matejgerek! 😎
I have reviewed the code and added a few comments.

libs/ui/src/lib/assistant/assistant.component.ts
matejgerek: 289c693

libs/ui/src/lib/assistant/assistant.component.ts
matejgerek: 08de88d

apps/api/src/services/tag/tag.service.ts
matejgerek: 7f80738

apps/api/src/services/tag/tag.service.ts
matejgerek: 3ed6027 - hmm, that makes sense, my bad 🤝

apps/api/src/services/tag/tag.service.ts
matejgerek: removing unnecessary `omit` import - e0f7357
missing userId in return - 877fd87

matejgerek: I initially chose omit because it's not explicit, meaning that the tag model can have other properties that don't need to be specified in the return. 
For example the user interface defines tags like this:
```ts
tags: (Tag & { isUsed: boolean })[];
```
But the tag model probably won't be changed as often so your suggestion makes more sense I guess.

dtslvr (approved):
I have made some final adjustments (see [here](https://github.com/ghostfolio/ghostfolio/pull/3859/commits/270b94dd1558d18bdaa62adc8d3d9d0bdc17fa45) if you are interested). Thanks again for this contribution, @matejgerek! 🙌🏻

Ready to merge now.

PR #3867, opened 2024-10-04
feat: set up husky and pre commit hook for linting and format check

### LINKED ISSUE #3866: [FEATURE] Set up husky and pre-commit hook for linting and format check
Set up [husky](https://github.com/typicode/husky) and a pre-commit hook for linting and format check (prettier).

_Originally posted by @dw-0 in https://github.com/ghostfolio/ghostfolio/issues/2782#issuecomment-2392182145_

dw-0:
Fixes #3866 

I noticed that npm run lint will only run linting on the "api" project.
So i modified the command to make sure linting runs on all project.
The format check will only run on uncommitted files, this flag is sadly not available for linting. But due to nx caching i don't see a huge impact here, the first commit might take a bit longer in case nx has nothing cached so far.
The linting command has a `--quiet` flag which is required, as warnings will make the command exit with status code 1 too.
We can remove that flag once everything is tidy again.

## How to test:
Fail on format issues:
- Disable prettier in your IDE settings
- Modify a file in a way that formatting is definitely incorrect
- Commit that file and it should fail
- 
Fail on lint issues:
- Set any rule to "error" that is currently set to "warn", e.g. `@typescript-eslint/member-ordering`
- Commit that change and it should fail


package.json
```diff
@@ -188,6 +188,7 @@
     "eslint-plugin-cypress": "2.15.1",
     "eslint-plugin-import": "2.29.1",
     "eslint-plugin-storybook": "0.6.15",
+    "husky": "^9.1.6",
```
dtslvr, line 191: Can you use the exact version (without `^`)?

dtslvr (changes requested):
Thank you for your contribution, @dw-0 🙂

Do we still need `./git-hooks/` and the instructions [here](https://github.com/ghostfolio/ghostfolio/blob/main/DEVELOPMENT.md#setup)?

dw-0:
> Do we still need `./git-hooks/` and the instructions [here](https://github.com/ghostfolio/ghostfolio/blob/main/DEVELOPMENT.md?rgh-link-date=2024-10-04T08%3A58%3A27Z#setup)?

I knew i forgot something :)
Fixed it now. And we now pinned the version to 9.1.6.

.husky/pre-commit
```diff
@@ -0,0 +1,7 @@
+# Run lint and abort if there are any lint errors
+# --quiet will supress warnings which would cause this check to fail too
+# this is temporary until all warnings are fixed
+npm run lint --quiet || exit 1
+
+# Run format:check on modified and uncommitted files and abort if there are any formatting errors
+npm run format:check --uncommitted || exit 1
```
dtslvr, line 7: Can we use `format:write` to auto fix potential errors?

dw-0: Yes i think we could do that. Will push an update in a second.

dtslvr:
> Modify a file in a way that formatting is definitely incorrect

I can commit a file in _GitHub Desktop_ without any problems which fails with a manual format check. Do I have to consider anything else? I have installed the new npm dependencies.

dw-0:
Interesting. The only gui integration i use for git is the one from Intellij IDEA, so im not familiar with Github Desktop. However, i found this issue here: https://github.com/desktop/desktop/issues/17385#issuecomment-1718170235
It is kind of odd if that is actually required to be set up prior so that the commit hooks will work. But currently i don't have another idea.

dtslvr (approved):
The `"prepare": "husky"` script was missing in the `package.json` file 😮 Now, it's working as expected. Linting is a bit slow, though.

I have switched back to `npm run format:check` as I think the developer experience is better (I got a weird state where it committed the auto formatted code but I still had an unstaged change).

Ready to merge now.

dw-0:
> The `"prepare": "husky"` script was missing in the `package.json` file 😮 Now, it's working as expected. Linting is a bit slow, though.

Hm, i thought that is only needed if the .husky folder with the pre-commit hook does not exist yet, as that command will just generate some template files. Well, lesson learned. Did you execute npm prepare then yourself? Or did Github Desktop do this??

Yes linting is slow the first time. But NX will generate the cache for each project. The workspace has 5 projects so if you change a file in only one project, the other 4 projects will be skipped next time you commit a file. We may be able to further improve it by adding another dependency to the project. Sadly nx does not support linting only affected files right now, but with https://github.com/lint-staged/lint-staged that could be achieved i think. Should i take a look at it before you merge this PR?


> I have switched back to `npm run format:check` as I think the developer experience is better (I got a weird state where it committed the auto formatted code but I still had an unstaged change).
> 
👍 im fine with it. I think every developer should have prettier run on saving a file anyways, and those who don't, need to look in the git console to find out why the commit fails.

dtslvr:
> Did you execute npm prepare then yourself?

It has generated some files when I ran `npm install` (every new developer needs to do it anyway).

PR #3876, opened 2024-10-05
feat: set up recommendedTypeChecked rule in eslint configuration

### LINKED ISSUE #3874: [FEATURE] Set up recommendedTypeChecked rule in eslint configuration
Setup `recommendedTypeChecked`.

> We could also think about adding the recommended sets of typescript-eslint (and overwrite every rule to "warn" until its fixed) before we start switching the previous rules back to error again. There might be some overlapping rules which may clean up the projects eslintrc.json once all rules are correctly in place again: 
> * https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended-type-checked.ts

_Originally posted by @dw-0 in https://github.com/ghostfolio/ghostfolio/issues/2782#issuecomment-2392182145_

dw-0:
Fixes #3874 

`"@typescript-eslint/no-misused-new": "error"` was removed, because it is already part of `@typescript-eslint/recommended-type-checked`
Other rules which are part of `@typescript-eslint/recommended-type-checked` were moved further down the file, separated by a comment, so they are easier to distinguish from the rest of the rules.
New rules which got introduced with `@typescript-eslint/recommended-type-checked` and would throw errors were set to "warn" so they can be solved later.

Next step could be, to also add `@typescript-eslint/stylistic-type-checked` to the rules (https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/stylistic-type-checked.ts) via the same process: add rules -> lint -> overwrite broken rules from "error" to "warn".
Alternative: consider adding https://github.com/eslint-stylistic/eslint-stylistic to the project

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -93,26 +93,28 @@
         "no-prototype-builtins": "warn",
         "no-async-promise-executor": "warn",
         "no-constant-condition": "warn",
-        // the following rules are part of @typescript-eslint/recommended-type-checked, remove once solved
-        "@typescript-eslint/no-unused-expressions": "warn",
-        "@typescript-eslint/ban-ts-comment": "warn",
-        "@typescript-eslint/no-unsafe-assignment": "warn",
-        "@typescript-eslint/no-unsafe-member-access": "warn",
-        "@typescript-eslint/no-unsafe-return": "warn",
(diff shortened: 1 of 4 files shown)


dtslvr (approved):
👍🏻

PR #3878, opened 2024-10-05
feat: set up stylisticTypeChecked rule in eslint configuration

### LINKED ISSUE #3877: [FEATURE] Set up stylisticTypeChecked rule in eslint configuration
Set up `stylisticTypeChecked`.

> Add @typescript-eslint/stylistic-type-checked to the rules (https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/stylistic-type-checked.ts) via the same process: add rules -> lint -> overwrite broken rules from "error" to "warn".

_Originally posted by @dw-0 in https://github.com/ghostfolio/ghostfolio/pull/3876#issue-2568064758_

dw-0:
Fixes #3877 

Draft, once this is clearified:
>I have a question to this rule: https://typescript-eslint.io/rules/naming-convention/
>It is currently completely disabled. Is this intended? Should it stay disabled?

_Originally posted by @dw-0 in https://github.com/ghostfolio/ghostfolio/issues/3877#issuecomment-2395105096_

Within the scope of this PR, i also added some more type and style checks to the tsconfig files. I think those make sense as well, i hope that was okay and does not require another PR. But since i made it its own commit, it wouldn't be a big issue to extract it into a new PR.

What i noticed when i worked on this PR is, that for some reason the linting can indeed take quite a long time even though the cache was already built. Especially when squashing commits, it seems like the cache is ignored. Maybe we really want to look into linting only staged files, this should speed up the overall linting quite a lot. The build action could then run the full lint check on new PRs (would require a new ticket for this).

### DIFF
--- a/apps/client/tsconfig.json
+++ b/apps/client/tsconfig.json
@@ -15,7 +15,8 @@
   ],
   "angularCompilerOptions": {
     "strictInjectionParameters": true,
-    "strictTemplates": false
+    "strictInputAccessModifiers": false, // todo: enable strict input access modifiers
+    "strictTemplates": false // todo: enable strict templates
   },
   "compilerOptions": {
     "target": "es2020"
--- a/apps/client/tsconfig.json
+++ b/apps/client/tsconfig.json
@@ -15,8 +15,9 @@
   ],
   "angularCompilerOptions": {
     "strictInjectionParameters": true,
-    "strictInputAccessModifiers": false, // todo: enable strict input access modifiers
-    "strictTemplates": false // todo: enable strict templates
+    // TODO: Enable stricter rules for this project
+    "strictInputAccessModifiers": false,
+    "strictTemplates": false
   },
   "compilerOptions": {
     "target": "es2020"
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -23,7 +23,6 @@
     },
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true,
-    // set to true to enable stricter type checking for this project
     "strict": false,
     "strictNullChecks": false,
     "strictPropertyInitialization": false,
@@ -34,7 +33,7 @@
     "noPropertyAccessFromIndexSignature": false,
     "noUnusedLocals": false,
     "noUnusedParameters": false,
-    "allowUnreachableCode": true // set to false for stricter type checking
+    "allowUnreachableCode": true
   },
   "exclude": ["node_modules", "tmp"]
 }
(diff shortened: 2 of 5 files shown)


dw-0:
@dtslvr I reenabled the `@typescript-eslint/naming-convention` rule and set it to warn. I think i defined a feasable naming convention for this project. But please doublecheck and see if there is any nogo present. Set the rule to "error" and run a lint. Then you will see all the things that do not match the defined naming convention. It is quite a lot, so you decide if we apply this rule as is, tweak the naming convention or keep it disabled.

dtslvr:
> I reenabled the `@typescript-eslint/naming-convention` rule and set it to warn.

Hmm, there are a lot of them. Not sure if we can ever fix them (e.g. the included resources from _Prisma_ are all in pascal case and considered a problem). Probably better setting it to `off` at the moment. Currently it is quite distracting to see all the warnings in the code.

tsconfig.base.json
```diff
@@ -20,7 +20,21 @@
       "@ghostfolio/client/*": ["apps/client/src/app/*"],
       "@ghostfolio/common/*": ["libs/common/src/lib/*"],
       "@ghostfolio/ui/*": ["libs/ui/src/lib/*"]
-    }
+    },
+    "noFallthroughCasesInSwitch": true,
+    "forceConsistentCasingInFileNames": true,
+    // set to true to enable stricter type checking for this project
```
dtslvr, line 26: `"strict": false, // TODO: Enable stricter type checking for this project`

apps/client/tsconfig.json
```diff
@@ -15,7 +15,8 @@
   ],
   "angularCompilerOptions": {
     "strictInjectionParameters": true,
-    "strictTemplates": false
+    "strictInputAccessModifiers": false, // todo: enable strict input access modifiers
+    "strictTemplates": false // todo: enable strict templates
```
dtslvr, line 19: `"strictTemplates": false // TODO: Enable strict templates`

dtslvr (changes requested):

dw-0:
> Hmm, there are a lot of them. Not sure if we can ever fix them (e.g. the included resources from _Prisma_ are all in pascal case and considered a problem). Probably better setting it to `off` at the moment. Currently it is quite distracting to see all the warnings in the code.

I agree, we could keep the definition i created but simply turn it off as you suggest. In the end, the naming convention doesn't really do anything, it's not like it makes the code more robust or safer, its purely aesthetic.

dtslvr (approved):

PR #3887, opened 2024-10-06
Feature/switch @typescript-eslint/no-unused-vars to error

dtslvr:


dtslvr:
cc @dw-0

dw-0:
Do you want to stick to only that eslint rule or do you want to enable `noUnusedLocals` and `noUnusedParameters` from tsconfig.base.json as well? As this will also result in removing unused code.

dtslvr:
* `noUnusedLocals`❓
* `noUnusedParameters` - I think I fixed parameters also in this PR, no?

> As this will also result in removing unused code.
Would you like to tackle this issue on your own or should i also push commits onto this PR?

Feel free 🙂

dw-0:
> * `noUnusedLocals`❓
> * `noUnusedParameters` - I think I fixed parameters also in this PR, no?

https://github.com/dtslvr/ghostfolio/blob/feature/switch-typescript-eslint-no-unused-vars-rule-to-error/tsconfig.base.json#L34-L35
Those two need to be enabled as well. I just saw that i must have f'ed up in the last PR and did not include the TODO above `strict: false`, like we have commented all the other things that are going to be tackled. 😅

> Feel free 🙂

👍

PR #3888, opened 2024-10-06
style: set prefer-const to error

dw-0:
cc @dtslvr 

I would start and work on this rule in parallel if you are okay with it :)

### DIFF
--- a/apps/api/src/app/admin/admin.service.ts
+++ b/apps/api/src/app/admin/admin.service.ts
@@ -233,7 +233,7 @@ export class AdminService {
     const extendedPrismaClient = this.getExtendedPrismaClient();
 
     try {
-      let [assetProfiles, count] = await Promise.all([
+      const symbolProfileResult = await Promise.all([
         extendedPrismaClient.symbolProfile.findMany({
           orderBy,
           skip,
@@ -264,6 +264,8 @@ export class AdminService {
         }),
         this.prismaService.symbolProfile.count({ where })
       ]);
+      const assetProfiles = symbolProfileResult[0];
+      let count = symbolProfileResult[1];
 
       const lastMarketPrices = await this.prismaService.marketData.findMany({
         distinct: ['dataSource', 'symbol'],
--- a/apps/api/src/app/import/import.service.ts
+++ b/apps/api/src/app/import/import.service.ts
@@ -272,10 +272,10 @@ export class ImportService {
       const currency = activity.currency;
       const date = activity.date;
       const error = activity.error;
+      let fee = activity.fee;
       const quantity = activity.quantity;
       const SymbolProfile = activity.SymbolProfile;
       const type = activity.type;
-      let fee = activity.fee;
       let unitPrice = activity.unitPrice;
 
       const assetProfile = assetProfiles[
--- a/apps/api/src/app/import/import.service.ts
+++ b/apps/api/src/app/import/import.service.ts
@@ -266,21 +266,18 @@ export class ImportService {
 
     const activities: Activity[] = [];
 
-    for (let [
-      index,
-      {
-        accountId,
-        comment,
-        currency,
-        date,
-        error,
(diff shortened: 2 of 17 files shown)


apps/api/src/app/admin/admin.service.ts
```diff
@@ -264,6 +264,8 @@ export class AdminService {
         }),
         this.prismaService.symbolProfile.count({ where })
       ]);
+      const assetProfiles = symbolProfileResult[0];
+      let count = symbolProfileResult[1];
```
dtslvr, line 268: I've always wondered whether it can be solved more elegantly 😄

apps/api/src/app/import/import.service.ts
```diff
@@ -266,21 +266,18 @@ export class ImportService {
...
+    for (const [index, activity] of activitiesExtendedWithErrors.entries()) {
+      const accountId = activity.accountId;
+      const comment = activity.comment;
+      const currency = activity.currency;
+      const date = activity.date;
+      const error = activity.error;
+      const quantity = activity.quantity;
+      const SymbolProfile = activity.SymbolProfile;
+      const type = activity.type;
+      let fee = activity.fee;
```
dtslvr, line 278: I would move up even if it is with `let`

apps/api/src/app/import/import.service.ts
```diff
@@ -266,21 +266,18 @@ export class ImportService {
...
+      let unitPrice = activity.unitPrice;
```
dtslvr, line 279: I would move up even if it is with `let`

dtslvr (changes requested):
Thank you for the contribution, @dw-0 👍🏻
I have added a few comments.

apps/api/src/app/import/import.service.ts
dw-0: You mean declare `fee` and `unitPrice` before the const declarations? so sort let before const?

apps/api/src/app/admin/admin.service.ts
dw-0:
```ts
const [assetProfiles, totalAssetCount] = await Promise.all([
	... , 
	...
]);
let count = totalAssetCount;
```
Im unsure if this would be more elegant than the other solution. Maybe a little bit more self explanatory than accessing the values only be their index.

Alternatively we could do something like this:
```ts
public async getMarketData(...): Promise<AdminMarketData>{
	const [assetProfiles, count] = await Promise.all([
		...,
		...
	]);
	adminMarketData.count = count;
	...
	adminMarketData.marketData = await Promise.all(...);

	if (presetId) { // this is btw not so elegant as well, we could get rid of one indentation here
    	if (presetId === 'ETF_WITHOUT_COUNTRIES') {
        	adminMarketData.marketData = adminMarketData.marketData.filter(...);
        } else if (presetId === 'ETF_WITHOUT_SECTORS') {
          adminMarketData.marketData = adminMarketData.marketData.filter(...);
        }

        adminMarketData.count = adminMarketData.marketData.length;
    }

	return adminMarketData;
}
```
But again, im not sure if this really makes it more elegant and if such a simple change should involve this refactoring.

dtslvr: No, let's keep your initial solution

apps/api/src/app/import/import.service.ts
dtslvr:
```
const accountId = activity.accountId;
const comment = activity.comment;
const currency = activity.currency;
const date = activity.date;
const error = activity.error;
let fee = activity.fee;
const SymbolProfile = activity.SymbolProfile;
const type = activity.type;
let unitPrice = activity.unitPrice;
```

dw-0: I see, now i understand :)

dtslvr (approved):

Issue #3755, opened 2024-09-12
[BUG]


VibeNL:
**Important Notice**

The issue tracker is **ONLY** used for reporting bugs. New features should be discussed in our [Slack](https://join.slack.com/t/ghostfolio/shared_invite/zt-vsaan64h-F_I0fEo5M0P88lP9ibCxFg) community or in [Discussions](https://github.com/ghostfolio/ghostfolio/discussions).

Incomplete or non-reproducible issues may be closed, but we are here to help! If you encounter difficulties reproducing the bug or need assistance, please reach out to our community channels mentioned above.

Thank you for your understanding and cooperation!

**Bug Description**

Lots of errors in the logs:
Nest] 103  - 09/12/2024, 5:31:13 PM   ERROR [ExceptionsHandler] job stalled more than allowable limit
Error: job stalled more than allowable limit
    at Job.finished (/ghostfolio/apps/api/node_modules/bull/lib/job.js:494:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TWRPortfolioCalculator.initialize (/ghostfolio/apps/api/main.js:1:428569)
    at async TWRPortfolioCalculator.getSnapshot (/ghostfolio/apps/api/main.js:1:425138)
    at async PortfolioService.getDetails (/ghostfolio/apps/api/main.js:1:485385)
    at async Promise.all (index 1)
    at async PortfolioService.getAccounts (/ghostfolio/apps/api/main.js:1:482711)
    at async PortfolioService.getAccountsWithAggregations (/ghostfolio/apps/api/main.js:1:483404)

**To Reproduce**

Have it running.

**Expected behavior**

A working application

**Screenshots**

**Logs**

See above

**Environment**

- Ghostfolio Version 2.107.0
-  Self-hosted
- Experimental Features disabled
- Browser
- OS

**Additional context**

flowcool:
Same here and have to reboot container and flush data :(

flowcool:
@dtslvr : Does it ring a bell, any plan for a fix ?

ceroma:
Just in case it helps someone - I currently have the following locally to get around this issue:
```
BullModule.registerQueue({
  name: PORTFOLIO_SNAPSHOT_QUEUE,
  settings: {
    lockDuration: 600000,
    stalledInterval: 0
  }
}),
```
in `apps/api/src/services/queues/portfolio-snapshot/portfolio-snapshot.module.ts`

[...]

PR #3894, opened 2024-10-08
Feature/expose portfolio snapshot calculation timeout as environment variable

ceroma:
Expose `PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` env var to allow people to configure the `lockDuration` option for the Portfolio Snapshot queue. `@nodejs/bull`'s default is 30s, so if a job holds the JS event loop longer than that it gets marked as stalled ([reference](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#advanced-settings)). The end-user experience right now for when that happens is that the ghostfolio container crashes.

Test Plan:
1. Before this diff, my container always crashes since my portfolio snapshot computation takes about 2 minutes
2. Apply this diff, set `PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` to 5 minutes (`300000`)
3. Open Accounts page to trigger portfolio snapshot computation
4. Check it no longer crashes


ceroma:
This should solve #3755

libs/common/src/lib/config.ts
```diff
@@ -51,6 +51,7 @@ export const DEFAULT_PAGE_SIZE = 50;
 export const DEFAULT_PROCESSOR_CONCURRENCY_GATHER_ASSET_PROFILE = 1;
 export const DEFAULT_PROCESSOR_CONCURRENCY_GATHER_HISTORICAL_MARKET_DATA = 1;
 export const DEFAULT_PROCESSOR_CONCURRENCY_PORTFOLIO_SNAPSHOT = 1;
+export const DEFAULT_QUEUE_LOCK_DURATION_PORTFOLIO_SNAPSHOT = 30000;
```
ceroma, line 54: Maybe we should default this to something higher to unbreak the folks experiencing the crash without requiring them to set an extra variable? Normally I'd prefer using the library's default, but since the behavior prior to using Job queues was to keeping computing the portfolio snapshot until done without crashing, it might make sense to set this higher here.

dtslvr: I am not yet sure whether this solution is appropriate, as `QUEUE_LOCK_DURATION` exposes a lot of implementation details.

Ideas I have in mind
* Expose something like a snapshot computation timeout
* Set the default to `0` or `30000` (as the library default)

ceroma: > I am not yet sure whether this solution is appropriate, as QUEUE_LOCK_DURATION exposes a lot of implementation details.

Good point, I agree.

> Expose something like a snapshot computation timeout

Sounds reasonable

> Set the default to `0` or `30000` (as the library default)

Do you know if `0` works? The [bull reference](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#advanced-settings) doesn't mention it being a valid value. I tried to find where it's used, and I think it gets translated to the `NX` param to a Redis `SET` call [here](https://github.com/OptimalBits/bull/blob/c2f37ee194f0d16d78329e2fa4ea335db37350a6/lib/commands/takeLock-1.lua#L13), which seems to expect a positive value.

dtslvr: > Do you know if `0` works? The [bull reference](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#advanced-settings) doesn't mention it being a valid value.

Sorry, I mixed it up with [stalledInterval](https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#queue):

```
interface AdvancedSettings {
  ...
  stalledInterval: number = 30000; // How often check for stalled jobs (use 0 for never checking).
  ...
}
```

My favorite at the moment: Name the variable `PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` with a default value of `30000`. Users with issues can increase that number.

apps/api/src/services/configuration/configuration.service.ts
```diff
@@ -59,6 +60,9 @@ export class ConfigurationService {
       PROCESSOR_CONCURRENCY_PORTFOLIO_SNAPSHOT: num({
         default: DEFAULT_PROCESSOR_CONCURRENCY_PORTFOLIO_SNAPSHOT
       }),
+      PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT: num({
+        default: DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT
+      }),
```
ceroma, line 65: I'm wondering if this should be `PROCESSOR_TIMEOUT_PORTFOLIO_SNAPSHOT` to be consistent with the other variables above?
```
PROCESSOR_CONCURRENCY_GATHER_ASSET_PROFILE
PROCESSOR_CONCURRENCY_GATHER_HISTORICAL_MARKET_DATA
PROCESSOR_CONCURRENCY_PORTFOLIO_SNAPSHOT
PROCESSOR_TIMEOUT_PORTFOLIO_SNAPSHOT
```

Thoughts?

dtslvr: See here: https://github.com/ghostfolio/ghostfolio/pull/3913

PR #3895, opened 2024-10-08
refactor: enable noUnusedLocals and noUnusedParameters

dw-0:
This PR enables noUnusedLocals and noUnusedParameters in `tsconfig.base.json`.
This way, issues are caught during compile time.
usage of one `// @ts-expect-error:` comment was required, as nothing i tried satisfied the compiler and it kept throwing that false positive, that the `configurationService` is not used, even though it is used directly afterwards inside of the super() call.
It would be possible to remove the `// @ts-expect-error:` and do the following:
```diff
-- private readonly configurationService: ConfigurationService
++ protected readonly configurationService: ConfigurationService
```
But this also does not feel completely correct, as `private` seems to be the correct access modifier in this context.

If the priority would be to rather get rid of the ts-expect-error comment, im fine with setting private to protected, yet im not sure if that would be correct and i don't actually understand why this satisfies the compiler.


apps/api/src/app/auth/google.strategy.ts
```diff
@@ -11,6 +11,7 @@ import { AuthService } from './auth.service';
 export class GoogleStrategy extends PassportStrategy(Strategy, 'google') {
   public constructor(
     private readonly authService: AuthService,
+    // @ts-expect-error: Property is being used in the constructor
     private readonly configurationService: ConfigurationService
```
dtslvr, line 15: It could work without `@ts-expect-error` when you avoid declaring it as a class property:
```
configurationService: ConfigurationService
```

apps/api/src/app/auth/google.strategy.ts
```diff
@@ -24,13 +25,7 @@ export class GoogleStrategy extends PassportStrategy(Strategy, 'google') {
...
   }
 
-  public async validate(
-    request: any,
-    token: string,
-    refreshToken: string,
-    profile: Profile,
-    done: Function
-  ) {
+  public async validate(profile: Profile, done: Function) {
```
dtslvr, line 28: I don't think this signature is still the same when these params are missing 🤔

apps/api/src/interceptors/transform-data-source-in-response/transform-data-source-in-response.interceptor.ts
```diff
@@ -21,7 +21,7 @@ export class TransformDataSourceInResponseInterceptor<T>
   ) {}
 
   public intercept(
-    context: ExecutionContext,
+    _context: ExecutionContext,
```
dtslvr, line 24: Why?

apps/api/src/app/auth/google.strategy.ts
dtslvr: Also use `_` prefix?

libs/ui/src/lib/line-chart/line-chart.component.ts
```diff
@@ -85,7 +86,7 @@ export class GfLineChartComponent
       Tooltip
     );
 
-    Tooltip.positioners['top'] = (elements, position) =>
+    Tooltip.positioners['top'] = (position: TooltipPosition) =>
```
dtslvr, line 89: I don't think this signature is still the same when these params are missing 🤔

libs/common/src/lib/chart-helper.ts
```diff
@@ -103,7 +103,7 @@ export function getVerticalHoverLinePlugin(
   colorScheme?: ColorScheme
 ) {
   return {
-    afterDatasetsDraw: (chart, x, options) => {
+    afterDatasetsDraw: (chart, options) => {
```
dtslvr, line 106: I don't think this signature is still the same when these params are missing 🤔

apps/client/src/app/components/investment-chart/investment-chart.component.ts
```diff
@@ -79,7 +80,7 @@ export class InvestmentChartComponent implements OnChanges, OnDestroy {
       Tooltip
     );
 
-    Tooltip.positioners['top'] = (elements, position) =>
+    Tooltip.positioners['top'] = (position: TooltipPosition) =>
```
dtslvr, line 83: I don't think this signature is still the same when these params are missing 🤔

apps/api/src/app/account/create-account.dto.ts
```diff
@@ -36,6 +36,6 @@ export class CreateAccountDto {
   name: string;
 
   @IsString()
-  @ValidateIf((object, value) => value !== null)
+  @ValidateIf((value) => value !== null)
```
dtslvr, line 39: Still the same meaning?

apps/api/src/app/account/update-account.dto.ts
```diff
@@ -35,6 +35,6 @@ export class UpdateAccountDto {
   name: string;
 
   @IsString()
-  @ValidateIf((object, value) => value !== null)
+  @ValidateIf((value) => value !== null)
```
dtslvr, line 38: Still the same meaning?

dtslvr (changes requested):
Thank you for the contribution, @dw-0. I have added a few comments.

apps/api/src/app/auth/google.strategy.ts
dw-0: Actually this method is not used at all, at least my IDE marks this method as unused, but i didn't want to remove the code completely in this PR and instead focus on the typescript compile errors.

Is this method called in a very specific place which expects the method to have that exact signature it had before my refactoring?

apps/api/src/app/auth/google.strategy.ts
dw-0: Something like this also worked :)
```
public constructor() {
    const configurationService = inject(ConfigurationService);
    super(...)
}
```

apps/api/src/interceptors/transform-data-source-in-response/transform-data-source-in-response.interceptor.ts
dw-0: The `_` typically signalizes that we have to declare a variable we don't intend to use. That's also something the typescript compiler understands and the reason why it works.
Another way would be to keep the parameter as is, and then inside the method, assign it to `_`. Something like this:
```ts
public intercept(context, next) {
	_ = context;
	<remaining code>
}
```
I don't think that is cleaner.

Do you think the interceptor will not work properly anymore if the first parameter is not explicitly called `context`?

dw-0:
@dtslvr Hi there, regarding your questions to the method signatures. Maybe i don't understand the codebase and context of it enough to see any issue with the method signatures. Do those method signatures you are referring to need to match that specific signature they had before? Maybe it would be better then to keep those parameters there, but prefix all intentionally unused ones with an `_` as well. Maybe a safer refactoring.

apps/api/src/app/auth/google.strategy.ts
dtslvr: This method is indirectly called by extending the `PassportStrategy` (_Google Sign_ in in the cloud).

apps/api/src/interceptors/transform-data-source-in-response/transform-data-source-in-response.interceptor.ts
dtslvr: Okay, I didn't know. I would suggest to prefix all unused arguments (where the order matters) with `_` then

dtslvr (commented):
Thank you for the adaptions.

When I try to start the backend on this branch via `npm run watch:server` (and start in VS Code), I get this error:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/user/ghostfolio/node_modules/@angular/core/fesm2022/core.mjs not supported.
Instead change the require of /Users/user/ghostfolio/node_modules/@angular/core/fesm2022/core.mjs to a dynamic import() which is available in all CommonJS modules.
    at /Users/user/ghostfolio/dist/apps/api/main.js:15316:47
    at Object.<anonymous> (/Users/user/ghostfolio/dist/apps/api/main.js:18593:12)
    at Object..js (/Users/user/ghostfolio/node_modules/ts-node/dist/index.js:851:20) {
  code: 'ERR_REQUIRE_ESM'
```

I don't see in the changes what is the problem. Any ideas @dw-0?

dw-0:
> I don't see in the changes what is the problem. Any ideas @dw-0?

Yeah i see, the issue is using `inject()` in `google.strategy.ts` for injecting the two services. I'll try and see what alternatives work.

dtslvr:
> I'll try and see what alternatives work.

I think I have tried this and it worked:
```
public constructor(configurationService: ConfigurationService) {
  ...
}
```

dw-0:
The only two options i see is:
1. set the access modifier to `protected` instead of `private`.
2. add the `// @ts-expect-error` comment to ignore this odd compiler behavior.

The actual reason behind the compiler complaining is that, when the service is injected in the constructor, it expects the service to be used in another place of the class. So in a method for example where the property is actually accessed with `this`. This does and cannot not happen in the constructor. Accessing `this` before a super call is forbidden and super must always be called before any access to `this` is done.

The reason why 1 works:
It tells the compiler that a subclass may access this property, so it shuts up. So on the one hand the error is gone, on the other hand this is technically not correct because there are no subclasses of GoogleStrategy making use of that property.

You had an idea here: https://github.com/ghostfolio/ghostfolio/pull/3895#discussion_r1795918878
But to be honest i did not really understand what you mean by that. Can you elaborate?

dw-0:
> > I'll try and see what alternatives work.
> 
> I think I have tried this and it worked:
> 
> ```
> public constructor(configurationService: ConfigurationService) {
>   ...
> }
> ```

Ah i see, now i understand what what you mean with https://github.com/ghostfolio/ghostfolio/pull/3895#discussion_r1795918878

The reason this works is almost the same as declaring the access modifier as protected. In your code, the access modifier now is implicitly `public`. Technically also not correct. But im fine with doing it that way if you are fine with it.

dtslvr (approved):
Thanks again, @dw-0! 🙌🏻
Ready to merge now.

Issue #3381, opened 2024-05-05
[BUG] Each interest entry creates a new "symbol"


lildude:
**Important Notice**

The issue tracker is **ONLY** used for reporting bugs. New features should be discussed in our [Slack](https://join.slack.com/t/ghostfolio/shared_invite/zt-vsaan64h-F_I0fEo5M0P88lP9ibCxFg) community or in [Discussions](https://github.com/ghostfolio/ghostfolio/discussions).

Incomplete or non-reproducible issues may be closed, but we are here to help! If you encounter difficulties reproducing the bug or need assistance, please reach out to our community channels mentioned above.

Thank you for your understanding and cooperation!

**Bug Description**

Each interest entry for an account creates a new "symbol" which results in a very full and cluttered "Market data" page.

**To Reproduce**

1. Add two interest activities to an account
2. Browse to Admin control -> Market Data
3. Note there are two entries. Click each and note the different "symbols".

**Expected behavior**

Interest really shouldn't get a symbol but if it must because of design decisions, the same "symbol" should be used per account, as is the case for a normal stock.

Maybe interest shouldn't be on the market data page at all?

**Screenshots**

![Image](https://github.com/ghostfolio/ghostfolio/assets/627280/de345a06-fc28-48a8-a3e1-18dade4cd28b)

![Image](https://github.com/ghostfolio/ghostfolio/assets/627280/117167ab-a79f-474d-a0b0-7af396e975b2)

![Image](https://github.com/ghostfolio/ghostfolio/assets/627280/e59dd4f6-562d-4939-afae-1bb30a6cadb0)

**Logs**

N/A

**Environment**

- Ghostfolio Version 2.78
- Self-hosted
- Experimental Features enabled or disabled - none
- Firefox
- macos

**Additional context**

dtslvr:
> Interest really shouldn't get a symbol but if it must because of design decisions, the same "symbol" should be used per account, as is the case for a normal stock.

Yes, each activity (_Order_) is associated to an asset profile (_SymbolProfile_), therefore reusing the asset profile per account could be cleaner. Would you like to work on that, @lildude?

dickwolff:
@dtslvr I may have a repro on this, or at least an issue that is related. 

When importing multiple interests, randomly it generates a new symbol for interest.

I have this input (taken from the import file:
```json
{
    "accountId": "8996c276-de46-4fed-8f2d-9270681dbd49",
    "comment": "Interest on cash",
[...]

PR #3899, opened 2024-10-09
RFC/one Symbol Profile per interest name

ceroma:
Currently, each Interest activity creates a new Symbol Profile. This may leave the DB cluttered with multiple single-activity Symbols (see #3381), causing a slowness in portfolio snapshot computation and some difficulty in parsing all symbols in Admin Control > Market Data page.

This is a proposal to keep a single Symbol Profile for each unique name used for an Interest-typed activity. What do you think?

Test Plan:
1. Create an Interest activity named "AAAA"
2. Create another Interest activity named "AAAA"
3. Verify in Market Data that only one Symbol Profile exists for both interest activities
4. Delete the first "AAAA" Interest activity
5. Verify the other activity still exists
6. Verify the "AAAA" Symbol Profile still exists
7. Delete the second "AAAA" Interest activity
8. Verify the "AAAA" Symbol Profile got deleted
9. Create an Interest activity named "BBBB"
10. Create another Interest activity named "BBBB"
11. Update one of the activities to be named "CCCC"
12. Verify one "BBBB" and one "CCCC" activities exist
12. Verify in Market Data that there are two Symbol Profiles, one for "BBBB" and one for "CCCC"
13. Rename Symbol Profile "CCCC" to "DDDD" from the Market Data page
14. Verifify the activity in Activities page shows up as "DDDD" now

### DIFF
--- a/apps/api/src/app/order/order.service.ts
+++ b/apps/api/src/app/order/order.service.ts
@@ -130,11 +130,28 @@ export class OrderService {
       data.SymbolProfile.connectOrCreate.create.assetSubClass = assetSubClass;
       data.SymbolProfile.connectOrCreate.create.dataSource = dataSource;
       data.SymbolProfile.connectOrCreate.create.name = name;
-      data.SymbolProfile.connectOrCreate.create.symbol = id;
       data.SymbolProfile.connectOrCreate.create.userId = userId;
+
+      let symbolProfileSymbol = id;
+
+      if (['FEE', 'INTEREST'].includes(data.type)) {
+        const symbolProfile = await this.prismaService.symbolProfile.findFirst({
+          where: {
+            dataSource: dataSource,
+            name: name,
+            userId: userId
+          }
+        });
+
+        if (symbolProfile) {
+          symbolProfileSymbol = symbolProfile.symbol;
+        }
+      }
+
+      data.SymbolProfile.connectOrCreate.create.symbol = symbolProfileSymbol;
       data.SymbolProfile.connectOrCreate.where.dataSource_symbol = {
         dataSource,
-        symbol: id
+        symbol: symbolProfileSymbol
       };
     }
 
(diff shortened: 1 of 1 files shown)


apps/api/src/app/order/order.service.ts
```diff
@@ -123,10 +123,29 @@ export class OrderService {
       data.SymbolProfile.connectOrCreate.create.assetSubClass = assetSubClass;
       data.SymbolProfile.connectOrCreate.create.dataSource = dataSource;
       data.SymbolProfile.connectOrCreate.create.name = name;
-      data.SymbolProfile.connectOrCreate.create.symbol = id;
+
+      let symbolProfileSymbol = id;
+
+      if (data.type === 'INTEREST') {
```
dtslvr, line 129: Shouldn't it be the same behavior for all types according to https://github.com/ghostfolio/ghostfolio/issues/3381#issuecomment-2171894093?

apps/api/src/app/order/order.service.ts
```diff
@@ -123,10 +123,29 @@ export class OrderService {
...
+        const symbolProfile = await this.prismaService.symbolProfile.findFirst({
+          orderBy: {
```
dtslvr, line 131: Probably not necessary for `findFirst`

apps/api/src/app/order/order.service.ts
```diff
@@ -123,10 +123,29 @@ export class OrderService {
...
+            createdAt: 'desc'
+          },
+          where: {
```
dtslvr, line 134: There will be collisions in the cloud setup. A `userId` (including migration path) is needed to avoid them.

apps/api/src/app/order/order.service.ts
```diff
@@ -620,12 +639,52 @@ export class OrderService {
...
+      const symbolProfileDataSource =
+        data.SymbolProfile.connect.dataSource_symbol.dataSource;
+      const symbolProfileName =
+        data.SymbolProfile.connect.dataSource_symbol.symbol;
+
+      delete data.SymbolProfile.update;
+      delete data.SymbolProfile.connect;
+
+      const symbolProfile = await this.prismaService.symbolProfile.findFirst({
+        orderBy: {
```
dtslvr, line 662: Probably not necessary for `findFirst`

apps/api/src/app/order/order.service.ts
dtslvr: Add `currency` as below?

dtslvr (changes requested):
Thank you for the contribution, @ceroma.
I have reviewed your RFC and added a few comments.

apps/api/src/app/order/order.service.ts
ceroma: > There will be collisions in the cloud setup. A `userId` (including migration path) is needed to avoid them.

You mean collisions like users A and B having the same INTEREST named "Bank Interest"? Shouldn't it be the fine, in the same way that it's fine for users A and B to have different activities pertaining the same AAPL symbol, for instance?

apps/api/src/app/order/order.service.ts
ceroma: I wasn't sure if those comments meant "yeah, they go through the same code path when being created" or "yeah, we should reuse profiles for all those types". Given that I'm not too familiar with how non-INTEREST types are used across the codebase, I felt it was safer to start with just INTERESTs. Should we apply the same logic to all of them then?

apps/api/src/app/order/order.service.ts
dtslvr: Maybe it would work for `INTEREST`, but I see an issue for `ITEM`. If you and me both own a _Beach House_, we could not share the same asset profile because of different historical market prices.

apps/api/src/app/order/order.service.ts
dtslvr: If `where` is unique (see my comment on collision below), I think `orderBy` is not needed

apps/api/src/app/order/order.service.ts
ceroma: Ah, good point! But in that same train of thought, I could see a user having multiple `ITEM` items with the same name to represent different physical items, right? Should we keep `ITEM`s as is so as to not break that expected behavior for historical market prices?

dtslvr: Yes, we could keep `ITEM` (and `LIABILITY`) as is and see them as unique.
As a result there will be one Symbol Profile per `INTEREST` / `FEE`?

ceroma: Sounds good to me, I'll update accordingly.
Just to confirm though - do you still think we should have `INTEREST`/`FEE` unique per `userId` or it's ok for them to be globally unique?

dtslvr: I would still do it unique per `userId`, otherwise if I rename it would also be renamed for you.

ceroma: So, in the current implementation that would not happen. See `updateOrder` below. When you rename a `INTEREST` order, it will either connect it to an existing `SymbolProfile` with the new name or create a new `SymbolProfile` if one doesn't exist with that name.

ceroma:
Updated the PR to consider `FEE`s as well, and querying by `userId`.

ceroma:
Hey @dtslvr, can you take a look at the new revision when you have a chance?

gabrielgradinaru:
Hi @dtslvr, could you pls review this PR?

PR #3916, opened 2024-10-12
Feature/add functionality to gather missing data only

dandevaud:
@dtslvr 

I had an issue locally with the currency data gathering job. 
I regularly run into errors as prisma runs out of available connections while trying to update all data.
Even within the defined 12 retries of the job I was unable to update all market data values.

Hence I've added a functionality to update the missing values only. It might be helpful for test environments as well, which are not running 24/7 and hence do not perform a daily market data update job.

Let me know what you think


asiananimal:
@dtslvr I would love to have this feature as my portfolio has a lot of different equity and on most days my portfolio doesn't get fully updated unless I manually go in and pull the missing data.

dandevaud:
I added the functionality to the market data tab as well to gather it for all Assets, previously one could only trigger it for one asset.

asiananimal:
@dtslvr What are your thoughts on whether this could be merged into the main branch?

PR #3973, opened 2024-10-22
refactor: run lint only on affected projects in pre-commit hook

dw-0:
@dtslvr As discussed by mail, here a reworked pre-commit hook that only runs linting on affected projects.
Reference: https://nx.dev/nx-api/nx/documents/affected

An decrease in committing time due to faster linting was observed. E.g. changes in project `client` may affect projects `client-e2e` and `ui`. Therefor those 3 projects get linted.
image


dtslvr (approved):
Thanks for the adaption, @dw-0! 👌🏻

PR #3979, opened 2024-10-24
Fix/@typescript-eslint/no-empty-function

dw-0:


apps/api/src/app/health/health.controller.ts
```diff
@@ -16,9 +16,6 @@ import { HealthService } from './health.service';
 export class HealthController {
   public constructor(private readonly healthService: HealthService) {}
 
-  @Get()
```
dtslvr, line 19: ⚠️ This will destroy the [health check endpoint](https://github.com/ghostfolio/ghostfolio#health-check-experimental)!

What about this?

```
@Get()
@HttpCode(HttpStatus.OK)
public async getHealth() {
  return { status: getReasonPhrase(HttpStatus.OK) };
}
```

dtslvr (changes requested):
Thanks for your contribution, @dw-0!
I have added a comment regarding health check endpoint. Can you also add an entry to the changelog similar to [this one](https://github.com/ghostfolio/ghostfolio/blob/main/CHANGELOG.md?plain=1#L111) please?

apps/api/src/app/health/health.controller.ts
dw-0: I see. I was wondering where it is actually used for. I send a request to it and no response came back so i thought it is not used and only the other health check endpoints in that file are used anymore as they actually return a response.
In that case we should actually make it really return a 200 status code like described in the link you provided.

dtslvr (approved):
Thanks for this contribution, @dw-0! 👍🏻
Ready to merge.

apps/api/src/app/health/health.controller.ts
dtslvr: I have extended the documentation by `status: "OK"`

PR #3980, opened 2024-10-24
Fix/@typescript-eslint/consistent type definitions

dw-0:
depends on #3979


libs/common/src/lib/types/toggle-option.type.ts
dtslvr, line 1: Can you rename and move the file to `libs/common/src/lib/interfaces/toggle-option.interface.ts` please?

libs/common/src/lib/types/x-ray-rules-settings.type.ts
dtslvr, line 1: Can you rename and move the file to `libs/common/src/lib/interfaces/x-ray-rules-settings.interface.ts` please?

dtslvr (changes requested):
Thanks for the contribution, @dw-0! 😎
I have reviewed the changes and added a few comments.

dtslvr (approved):
Thanks @dw-0! 🙌🏻
Ready to merge now.

PR #3985, opened 2024-10-24
Fix/@typescript-eslint/consistent-generic-constructors

dw-0:
depends on #3983

PR #3986, opened 2024-10-24
Fix/@typescript-eslint/consistent-indexed-object-style

dw-0:
depends on #3985 

If you think with this PR there is too much information going to get missing, we have two options i guess:
1. create custom types that better describe what each record holds
2. disable the rule and do not enforce the use of records


dtslvr:
Yes, I think some of the information is lost. But introducing custom types seems like overkill 🤔

dw-0:
I agree that custom types would be overkill but i don't know how else the "value information" could be preserved. 
The only solution i found so far to keep information from this:
```ts
exchangeRates: { [dateString: string]: number };
```
or this:
```ts
marketSymbolMap: { [date: string]: { [symbol: string]: Big } };
```
is this:
```ts
type DateString = string;
type SymbolString = string;

exchangeRates: Record<DateString, number>;
marketSymbolMap: Record<DateString, Record<SymbolString, Big>>;
```
On the one hand, now it is very clear what the Record is about, on the other hand it feels tedious to go through the project and define dozens of custom types which are only aliases for primitives.

I have no clear opinion on how tragic the lost information is, so that is something you have to decide. 😅

dtslvr:
> On the one hand, now it is very clear what the Record is about, on the other hand it feels tedious to go through the project and define dozens of custom types which are only aliases for primitives.
> 
> I have no clear opinion on how tragic the lost information is, so that is something you have to decide. 😅

Same here. Maybe leave it as it is for now, I don't see the advantage of using `Record`.

dw-0:
Thats fine to me. Any dev who wants to use Record can use it anyways.
Should i create a new PR that switches the rule off? Or would you like to have to shown as a warning still?

dtslvr:
> Should i create a new PR that switches the rule off? Or would you like to have to shown as a warning still?

I’d prefer switching it off. Feel free to go ahead with a PR!

PR #3992, opened 2024-10-27
Proposal: Support for PUID/PGID

asiananimal:


ymage:
It seems you removed the `USER node` line in Dockerfile, so the container starts as root before switching to less-privilege user.

PR #4000, opened 2024-10-29
cicd: add stackname/container names to docker compose files

dw-0:
By adding (more or less unique) names to each stack and container, you allow to have multiple stacks at the same time and you can prevent container naming issues that e.g. a container with the name "redis" or "postgres" already exists:
image

### DIFF
--- a/docker/docker-compose.build.yml
+++ b/docker/docker-compose.build.yml
@@ -1,6 +1,8 @@
+name: ghostfolio_build
 services:
   ghostfolio:
     build: ../
+    container_name: gf-application-build
     init: true
     env_file:
       - ../.env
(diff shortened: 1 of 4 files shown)


docker/docker-compose.yml
```diff
@@ -1,6 +1,8 @@
+name: ghostfolio
 services:
   ghostfolio:
     image: docker.io/ghostfolio/ghostfolio:latest
+    container_name: gf-application
```
dtslvr, line 5: Suggestion: `container_name: ghostfolio`

docker/docker-compose.build.yml
```diff
@@ -1,6 +1,8 @@
+name: ghostfolio_build
 services:
   ghostfolio:
     build: ../
+    container_name: gf-application-build
```
dtslvr, line 5: Suggestion: `container_name: ghostfolio-build`

dtslvr (approved):
Thanks @dw-0! 🤓

Issue #3995, opened 2024-10-27
[BUG] Add buy / sell / dividend activity dialog allows for negative quantity input


dw-0:
**Bug Description**

When adding a new buy activity, the quantity field allows negative input. Although you cannot save the activity with a negative quantity, i think it would improve UI/UX by simply disallowing negative input overall. Alternatively, the save button should be disabled as soon as invalid input is detected and the input field should be marked accordingly.

**To Reproduce**

1. Add new buy, sell or dividend activity of any kind, e.g. for symbol `VUAA.L`
2. Type in a negative number in the quantity input field or press the down input spinner to decrease amount until negative

**Expected behavior**

Quantity field should not go below 0

**Screenshots**

![image](https://github.com/user-attachments/assets/b2042ecd-1d94-4a14-8d19-887ac194eb1a)

**Environment**

- Ghostfolio Version 2.119.0
- Self-hosted / Dev
- Experimental Features disabled
- Browser: Brave
- OS: Windows

dw-0:
@dtslvr 😄

dtslvr:
For a generic solution, [validateObjectForForm()](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/util/form.util.ts#L5) should be adapted so that `@Min` of [CreateOrderDto](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/order/create-order.dto.ts) is taken into account.

mohanbyte:
Hi @dtslvr , @dw-0  can i work on this issue?

mohanbyte:
![image](https://github.com/user-attachments/assets/d49cadaf-a6b3-4c37-95ff-1f19ee4b8892)
Updated as per expected behavior. Let me know if any modifications needed.

johnson-jnr:
hi @dtslvr 
Can this issue still be picked up?

PR #4035, opened 2024-11-08
Bugfix : minimum value constraint to quantity input in activity dialog

mohanbyte:
This PR fixes ghostfolio/ghostfolio #3995

### DIFF
--- a/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
+++ b/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.html
@@ -186,7 +186,7 @@
     >
       <mat-form-field appearance="outline" class="w-100">
         <mat-label i18n>Quantity</mat-label>
-        <input formControlName="quantity" matInput type="number" />
+        <input formControlName="quantity" matInput type="number" [min]="0" />
       </mat-form-field>
     </div>
     <div


dtslvr (commented):
Thanks for your contribution, @mohanbyte!

The proposed solution is straightforward, and I appreciate the approach taken here! One limitation, though, is that it duplicates the validation logic in the template which could lead to maintenance challenges. A more generic solution might work better in this case. Could you take a look at [my comment](https://github.com/ghostfolio/ghostfolio/issues/3995#issuecomment-2442330080) for further details and see that approach would be feasible?

mohanbyte:
Hi @dtslvr based on your feedback do you want me utilize validateObjectForForm? ( this i have observed to be utilized on submit on other components). which i could use on value changes to validate the form and including other fields in 'ignoreFields' property.
Or does moving the validation to class suffice like this   
```
this.activityForm.get('quantity').setValidators(Validators.min(0));
    this.activityForm.get('quantity').updateValueAndValidity();
```

dtslvr:
> based on your feedback do you want me utilize validateObjectForForm? ( this i have observed to be utilized on submit on other components). which i could use on value changes to validate the form and including other fields in 'ignoreFields' property.

Yes, exactly. Using `validateObjectForForm` not only on submit, but on value changes / blur to increase the responsiveness of the validation.

mohanbyte:
Hi @dtslvr , have implemented accordingly. Let me know if any modifications needed.

dtslvr (changes requested):
Thanks for the update, @mohanbyte!
It indeed seems to work now. However, I don't like the boilerplate code it needs to validate a single field. Would it be possible to hook this code inside of
```
this.activityForm.valueChanges
```
and ditch the ignored fields?

mohanbyte:
Hi @dtslvr , does this look good? Let me know if any changes required.

dtslvr:
@mohanbyte I have refactored the code. It works as expected for a negative `quantity`. Unfortunately, it does not invalidate the field for a negative `unitPrice` nor `fee` 🤔

Do you have an idea why the borders of _Unit Price_ / _Fee_ do not get red?

mohanbyte:
@dtslvr the error states for both are not visible because formcontrol with those name are hidden and the fields which are visible are set with a different formControlName i.e 

> unitPriceInCustomCurrency  & feeInCustomCurrency

`  <div class="d-none">
      <mat-form-field appearance="outline" class="w-100">
    ...
        <input formControlName="unitPrice" matInput type="number" />
        <span class="ml-2" matTextSuffix>{{
          activityForm.get('currency').value
        }}</span>
      </mat-form-field>
    </div>`

`    <div class="d-none">
      <mat-form-field appearance="outline" class="w-100">
        <mat-label i18n>Fee</mat-label>
        <input formControlName="fee" matInput type="number" />
        <span class="ml-2" matTextSuffix>{{
          activityForm.get('currency').value
        }}</span>
      </mat-form-field>
    </div>`

dtslvr:
Good point, @mohanbyte!

When you enter a negative fee, it does not work immediately (as you have explained), but `onSubmit` the border still gets red. Can we adopt the same behavior for `valueChanges`?

mohanbyte:
@dtslvr  have updated as per expected behavior , let me know if any more modifications needed.

dtslvr:
> @dtslvr have updated as per expected behavior , let me know if any more modifications needed.

Hi @mohanbyte, I have tested your recent update. When I enter a fee of `-1`, the input field gets invalid (red) which is fine. When I adjust to the value to `1`, the input field is still invalid (red). The user is unable to add an activity in this state.

PR #4073, opened 2024-11-25
coingecko nft support

devinaconley:
this PR extends the coingecko provider to support NFT pricing

I _almost_ put this in a separate provider class, but the `/search` endpoint already returns both sets of results, which felt like a good case to keep them unified 

currently this uses the `nft:` prefix on the symbol to differentiate token type and which endpoints to use. it would probably be cleaner to use the asset type as a switch, but that felt like a major change to the provider interface


devinaconley:
just updated this PR with the latest upstream @dtslvr

PR #4153, opened 2024-12-29
Feature/refactor got calls to use AbortSignal.timeout

slagiewka:
Instead of manually creating AbortController and controlling the abort with setTimeout.

Feature available since node v16.14.0 and v17.3.0[^1] and is built to replace the exact scenario that all these requests have.

[^1]:https://nodejs.org/docs/latest-v22.x/api/globals.html#static-method-abortsignaltimeoutdelay


slagiewka:
BTW. I have no idea why this works, since `got` is in v11 here, and the support for that was added in [v12.2](https://github.com/sindresorhus/got/releases/tag/v12.2.0). 
However, I did a quick test with a 1ms timeout and calls are indeed failing then.

dtslvr (approved):
Thanks for simplifying this, @slagiewka! 👍🏻

PR #4154, opened 2024-12-29
Feature/remove got in favour of WHATWG Fetch

slagiewka:
The built-in `fetch` is based on `undici`. It became stable in node v21[^1] with imported undici@v5.25.4.

While node 20 does not not have it unflagged as experimental, it is not hidden behind a flag and it's pinned to undici v6.20.0[^2].

[^1]:https://nodejs.org/en/blog/release/v21.0.0
[^2]:https://nodejs.org/en/blog/release/v20.18.1


slagiewka:
The difference that I can see is that `got` has some retries built-in and these were not explicitly disabled in any of the use cases here. 

If this is a requirement, just replacing it with `fetch` will not suffice.

--- 

The problems with current `got`:
1. Currently locked to v11 ([from 2 years ago](https://github.com/sindresorhus/got/releases/tag/v11.8.6)) 
2. Newer versions are ESM only
3. This project is still based on node 20, therefore it can't use non-experimental `require(esm)` (v22 is required or ghostfolio should switch to ESM).
4. It's not as portable as WHATWG Fetch is.
5. It's another dependency. If retries are not required - leaner `package.json` is always nice to see.
6. v11 has no `signal` officially added to the API (though it somehow works)

---

Let me know what you think. I don't mind if the project stays on `got`. However, I'm quite curious what are your views on:
1. Updating node
2. Moving to ESM
3. Reducing unnecessary dependencies in favour of the standard library

apps/api/src/app/logo/logo.service.ts
```diff
@@ -50,13 +49,12 @@ export class LogoService {
...
-    return got(
+    return fetch(
       `https://t0.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=${aUrl}&size=64`,
       {
         headers: { 'User-Agent': 'request' },
-        // @ts-ignore
         signal: abortController.signal
       }
-    ).buffer();
+    ).then((res) => res.arrayBuffer());
```
slagiewka, line 58: Seems like it breaks the platform logos for accounts. I'm trying to figure out how to replace it properly.

slagiewka: OK, fixed. 

BTW. If `response.blob()` would be used, we could use both `blob.arrayBuffer()` as well as `blob.type`. `type` contains the original media type. 
The issue is that `image/png` is hardcoded: https://github.com/ghostfolio/ghostfolio/blob/46cbbd3bc4af9ed395ef81476190eb37640a83de/apps/api/src/app/logo/logo.controller.ts#L49

I found a logo that had `image/jpeg` originally. And even though it was sent with `image/png` instead, it still shows itself OK.

dtslvr:
It seems that the merge of https://github.com/ghostfolio/ghostfolio/pull/4153 has caused issues with this PR.

> However, I'm quite curious what are your views on:
> 
> 1. Updating node
> 2. Moving to ESM
> 3. Reducing unnecessary dependencies in favour of the standard library

Thanks for your thoughts!

**Node.js**: Yes, we should keep up with new versions, but sticking with LTS versions is important.
**Moving to ESM**: Okay, as long as it’s done in small, manageable iterations.
**Reducing unnecessary dependencies**: Absolutely, using the standard library is preferable where possible.

slagiewka:
> It seems that the merge of #4153 has caused issues with this PR.

Yeah, I kind of expected that :sweat_smile: Never mind, already rebased :wink:

apps/api/src/services/data-provider/ghostfolio/ghostfolio.service.ts
```diff
@@ -140,7 +138,7 @@ export class GhostfolioService implements DataProviderInterface {
           // @ts-ignore
```
dtslvr, line 138: Remove

apps/api/src/app/info/info.service.ts
```diff
@@ -199,16 +196,15 @@ export class InfoService {
 
   private async countGitHubStargazers(): Promise<number> {
     try {
-      const { stargazers_count } = await got(
+      const { stargazers_count } = await fetch(
```
dtslvr, line 199: Better❓
```
const { stargazers_count }: { stargazers_count: number } = await fetch(
```

apps/api/src/app/info/info.service.ts
```diff
@@ -155,16 +154,15 @@ export class InfoService {
 
   private async countDockerHubPulls(): Promise<number> {
     try {
-      const { pull_count } = await got(
+      const { pull_count } = await fetch(
```
dtslvr, line 157: Better❓
```
const { pull_count }: { pull_count: number } = await fetch(
```

apps/api/src/services/data-provider/data-enhancer/openfigi/openfigi.service.ts
```diff
@@ -43,14 +42,20 @@ export class OpenFigiDataEnhancerService implements DataEnhancerInterface {
...
-        // @ts-ignore
+    const mappings = (await fetch(
+      `${OpenFigiDataEnhancerService.baseUrl}/v3/mapping`,
+      {
+        headers: {
+          'Content-Type': 'application/json',
+          ...headers
+        },
+        method: 'POST',
+        body: JSON.stringify([
```
dtslvr, line 53: Can you move up to keep the alphabetical order?

dtslvr (changes requested):
Thanks for your contribution, @slagiewka! 😎

apps/api/src/app/info/info.service.ts
slagiewka: I added the cast after the call to be consistent with other places. Also adding types on const is less convicing for some LSPs.

dtslvr (approved):
Thanks for your contribution including the adaptions, @slagiewka! 🙌🏻
Ready to merge now.

apps/api/src/app/logo/logo.service.ts
dtslvr: Removing `response.contentType('image/png');` is a great idea. I have created a follow-up issue (https://github.com/ghostfolio/ghostfolio/issues/4172). Would you like to work on it, @slagiewka?

PR #4155, opened 2024-12-29
Feature/refactor to use body-parser.json supplied from Nest

slagiewka:
It's not required to handle that dependency manually as Nest (platform-express) supports that via this helper method.

The version in use will be controlled by `@nestjs/platform-express` and is currently the same as the removed one.

One thing less to worry about.

Tested manually (no idea if this is automated in any way) by trying to upload the `test/import/ok-500-activies.json` file. It still works with this change and doesn't work if I disable it temporarily :slightly_smiling_face:


dtslvr (approved):
> One thing less to worry about.

Thank you for the contribution, @slagiewka! 😎

PR #4170, opened 2024-12-31
Feature/refactor to use less `lodash`

slagiewka:
In API space only. As the project targets ES2022, some built-ins can be used.

When ES2024 is available (node 22), stuff like Ojbect.groupBy will be available.

I'm also tempted to reduce the use of `_.sortBy` as it can be replaced by the standard library. I have no idea why lodash is used.


slagiewka:
> As the project targets ES2022

Actually, I think this assumption was wrong. I thought I saw that in the API space. But it's [set to ES2021](https://github.com/ghostfolio/ghostfolio/blob/b9917e72b2ae6c9dfa9c3cdcd9bc556c8141eb34/apps/api/tsconfig.app.json#L7). 

Now I don't know why it works then. I also don't see `lib` being specified anywhere in tsconfigs.

I did a quick test and used `Object.groupBy` in one place. My LSP reports a `ts` problem.
image

However, running `npm run lint` shows no issues. Possibly there's something misconfigured and errors are not showing up.

dtslvr (approved):
Thanks for your contribution, @slagiewka! 🤓

PR #4173, opened 2025-01-01
Feature/get logo contentType from blob

### LINKED ISSUE #4172: [FEATURE] Remove hard-coded content type from logo controller
If `response.blob()` would be used, we could use both `blob.arrayBuffer()` as well as `blob.type`. `type` contains the original media type. 
The issue is that `image/png` is hardcoded: https://github.com/ghostfolio/ghostfolio/blob/46cbbd3bc4af9ed395ef81476190eb37640a83de/apps/api/src/app/logo/logo.controller.ts#L49

I found a logo that had `image/jpeg` originally. And even though it was sent with `image/png` instead, it still shows itself OK.

_Originally posted by @slagiewka in https://github.com/ghostfolio/ghostfolio/pull/4154#discussion_r1899192436_

slagiewka:
The property `type` contains the MIME type[^1] which can be forwarded to the caller.

[^1]:https://developer.mozilla.org/en-US/docs/Web/API/Blob/type

Fixes #4172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Eliminated `got` in favor of using `fetch`
+- Send original MIME type with the forwarded logo
 - Changed the `REDIS_HOST` from `localhost` to `redis` in `.env.example`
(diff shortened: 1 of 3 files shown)


dtslvr (approved):
Thanks for this improvement, @slagiewka! 😎
The code looks fine, ready to merge.

Issue #4178, opened 2025-01-02
Cloning fee the NAME is not saved correctly


Lipown:
**Bug Description**

When cloning fee, Name is change to some sort of ID. No matter if I edit the name afterwards. 

![image](https://github.com/user-attachments/assets/94e08dcf-bc3e-4281-8483-419964369297)

**To Reproduce**

1. Got Portfolio > Activites > Choose fee to clone
2. Enter Name
3. Save
4. The name in activities is changed to some sort of ID

**Expected behavior**

Name is saved as entered

**Environment**

- Ghostfolio Version 2.132.0 - 2024-12-30
- Self-hosted 
- Experimental Features enabled
- Browser - Firefox Nightly
- OS - W11

Schwenksen:
It's also the same by cloning Interest/Zinsen. It can be edited but the ID shows up once again.

Schwenksen:
I found out that when I edit in marketdata the ID to the proper name then it's displayed correct after a refresh.

PR #4183, opened 2025-01-09
bugfix: update name of the item type

fcrizul:
Issue:
While editing orders of type ITEM, it was discovered that the name field was being replaced with the order ID instead of retaining the updated name. This caused the name field to display incorrect data after editing.

Fix:
Adjusted the logic to ensure that the name field is updated correctly during the editing process. Now, when editing an order of type ITEM, the name will reflect the new value provided instead of being overwritten by the order ID.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


dtslvr:
Is this fix related to https://github.com/ghostfolio/ghostfolio/issues/4178?

fcrizul:
> Is this fix related to #4178?

@dtslvr no, but I've added a commit for that case

dtslvr (approved):
Thanks for your contribution, @fcrizul! 😎
I have added a commit to handle activities of type `INTEREST` and `LIABILITY` as well.
Ready to merge now.

PR #4187, opened 2025-01-10
import: fix import of FEE record with unitPrice=0

ikruglov:
The current code gives a misleading error: `activities.0 historical exchange rate at 2025-01-10 is not available from "EUR" to "USD"` when importing a record with unitPrice=0.

The reason is that the code treats `unitPrice=0` as an undefined value, giving the error.

This PR fixes it.


dtslvr (approved):
Thanks for your contribution, @ikruglov! 👍🏻

PR #4190, opened 2025-01-12
Add Ukrainian Localization (Experimental Feature)

SerhiiSerdiukDev:
This Pull Request introduces Ukrainian localization to the project.

**Key Details**:

- Ukrainian localization is included but is currently marked as an experimental feature.
- Users can only enable the Ukrainian language by toggling the Experimental Features option in the settings.
- The localization was generated using ChatGPT-4o and subsequently refined through manual correction to ensure accuracy.


apps/client/src/app/app.component.html
```diff
@@ -180,6 +180,11 @@
             <li>
               <a href="../tr" title="Ghostfolio in Türkçe">Türkçe</a>
             </li>
+            @if (user?.settings?.isExperimentalFeatures) {
```
dtslvr, line 183: Can you use `<!-- -->` for now?

package-lock.json
dtslvr, line 1: Can you please revert these changes?

apps/client/src/app/app.component.html
SerhiiSerdiukDev: Done.

package-lock.json
SerhiiSerdiukDev: Done.
However, a new developer will have a problem using the existing `package-lock.json`. They will get an error:
```
NX   Cannot find project 'api'
```

dtslvr: Okay, could you create a separate PR addressing this problem?

PR #4243, opened 2025-01-23
Bugfix/fix format options needed in `getNumberFormatGroup`

### LINKED ISSUE #3652: [BUG] Cannot read properties of undefined (reading 'value')
**Bug Description**

When accessing the Overview page, it does not display how much I currently have invested. It does show the graph and the current ROI, but not the main "number". 

**Expected behavior**

**Screenshots**

image

**Logs**

No useful logs in the docker logs. 

Browser logs:
```
main.a9f5126504e8e9e5.js:3  ERROR TypeError: Cannot read properties of undefined (reading 'value')
    at Qd (main.a9f5126504e8e9e5.js:3:432741)
    at b.ngOnChanges (2287.33f0b3b24f935621.js:1:13315)
    at b.Gn (main.a9f5126504e8e9e5.js:3:1278482)
    at Uf (main.a9f5126504e8e9e5.js:3:1283427)
    at Th (main.a9f5126504e8e9e5.js:3:1283583)
    at Sh (main.a9f5126504e8e9e5.js:3:1283329)
    at gu (main.a9f5126504e8e9e5.js:3:1282999)
    at mT (main.a9f5126504e8e9e5.js:3:1318596)
    at G_ (main.a9f5126504e8e9e5.js:3:1319927)
    at Xb (main.a9f5126504e8e9e5.js:3:1319749)
```

**Environment**

- Ghostfolio Version 2.102.0
- Self-hosted
- Edge Browser
- Windows

**Additional context**

I saw in #1057 there is a similar error that has to do with currencies. I have mostly EUR but also some USD and crypto, my main is EUR, not sure if that could be the issue

Kaysera:
Fixes #3652 

I was able to reproduce and diagnose the bug on development. It is found in the function `getNumberFormatGroup()`.

https://github.com/ghostfolio/ghostfolio/blob/553c10ac91ebdfa27ddfbd945d62f225f1c1d388/libs/common/src/lib/helper.ts#L245-L251

More precisely, the line 

https://github.com/ghostfolio/ghostfolio/blob/553c10ac91ebdfa27ddfbd945d62f225f1c1d388/libs/common/src/lib/helper.ts#L246

Assumes that the locale uses [grouping separators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#usegrouping), but when the locale is `es-ES` it does not and the function returns `undefined`.

The solution would be adding the option to always using grouping separators:

```typescript
 export function getNumberFormatDecimal(aLocale?: string) { 
   const formatOptions = {'useGrouping': true}
   const formatObject = new Intl.NumberFormat(aLocale, formatOptions).formatToParts(9999.99); 
  
   return formatObject.find((object) => { 
     return object.type === 'decimal'; 
   }).value; 
 } 
```

### DIFF
--- a/libs/common/src/lib/helper.spec.ts
+++ b/libs/common/src/lib/helper.spec.ts
@@ -41,12 +41,71 @@ describe('Helper', () => {
   });
 
   describe('Get Number Format Group', () => {
+    let languageGetter;
+    beforeEach(() => {
+      languageGetter = jest.spyOn(window.navigator, 'language', 'get');
+    });
     it('Get en-US number format group', () => {
       expect(getNumberFormatGroup('en-US')).toEqual(',');
     });
 
+    it('Get en-US number format group when it is default', () => {
+      languageGetter.mockReturnValue('en-US');
+      expect(getNumberFormatGroup()).toEqual(',');
+    });
+
+    it('Get en-GB number format group', () => {
+      expect(getNumberFormatGroup('en-GB')).toEqual(',');
+    });
+
+    it('Get en-GB number format group when it is default', () => {
+      languageGetter.mockReturnValue('en-GB');
+      expect(getNumberFormatGroup()).toEqual(',');
+    });
+
     it('Get es-ES number format group', () => {
       expect(getNumberFormatGroup('es-ES')).toEqual('.');
+    });
+
+    it('Get es-ES number format group when it is default', () => {
+      languageGetter.mockReturnValue('es-ES');
+      expect(getNumberFormatGroup()).toEqual('.');
+    });
+
+    it('Get de-DE number format group', () => {
+      expect(getNumberFormatGroup('de-DE')).toEqual('.');
+    });
+
+    it('Get de-DE number format group when it is default', () => {
+      languageGetter.mockReturnValue('de-DE');
+      expect(getNumberFormatGroup()).toEqual('.');
+    });
+
+    it('Get de-CH number format group', () => {
+      expect(getNumberFormatGroup('de-CH')).toEqual('’');
+    });
+
+    it('Get de-CH number format group when it is default', () => {
+      languageGetter.mockReturnValue('de-CH');
+      expect(getNumberFormatGroup()).toEqual('’');
+    });
+
+    it('Get zh-CN number format group', () => {
+      expect(getNumberFormatGroup('zh-CN')).toEqual(',');
+    });
+
+    it('Get zh-CN number format group when it is default', () => {
+      languageGetter.mockReturnValue('zh-CN');
+      expect(getNumberFormatGroup()).toEqual(',');
+    });
+
+    it('Get ru-RU number format group', () => {
+      expect(getNumberFormatGroup('ru-RU')).toEqual(' ');
+    });
+
+    it('Get ru-RU number format group when it is default', () => {
+      languageGetter.mockReturnValue('ru-RU');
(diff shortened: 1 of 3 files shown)


libs/common/src/lib/helper.ts
```diff
@@ -251,7 +251,11 @@ export function getNumberFormatDecimal(aLocale?: string) {
 }
 
 export function getNumberFormatGroup(aLocale = getLocale()) {
-  const formatObject = new Intl.NumberFormat(aLocale).formatToParts(9999.99);
+  const formatOptions = { useGrouping: true };
+  const formatObject = new Intl.NumberFormat(
+    aLocale,
+    formatOptions
```
dtslvr, line 257: Could you please add `{ useGrouping: true }` directly here?

dtslvr (changes requested):
Thanks for your contribution, @Kaysera! 👍🏻

Could you please extend [helper.spec.ts](https://github.com/ghostfolio/ghostfolio/blob/main/libs/common/src/lib/helper.spec.ts) by a test case to demonstrate your improvement?

Kaysera:
Added the proposed change and expanded the test cases

libs/common/src/lib/helper.spec.ts
dtslvr, line 1: I'm not sure if these tests fully demonstrate the impact of your change. I was more thinking of an example where `extractNumberFromString()` with the locale `es-ES` and some value would previously fail but now works correctly after your change. Would you be able to add a test case like that? What do you think?

dtslvr (changes requested):

libs/common/src/lib/helper.spec.ts
Kaysera: Given that now we pass `useGrouping: true` to the `Intl` function, it won't fail anymore. Should I pass the options for `Intl` as parameter, so that if it is empty (as it was before) it would fail but if it has `useGrouping: true` then it does not fail? Something like:

```javascript
 export function getNumberFormatGroup(aLocale = getLocale(), options = {useGrouping: true}) { 
   const formatObject = new Intl.NumberFormat(aLocale, formatOptions).formatToParts(9999.99); 
  
   return formatObject.find((object) => { 
     return object.type === 'group'; 
   }).value; 
 } 
```
So that the test case can be something like
```javascript
expect(getNumberFormatGroup('es-ES', {})).toEqual(undefined)
expect(getNumberFormatGroup('es-ES')).toEqual('.')
```

What do you think?

dtslvr: Adding this `expect(getNumberFormatGroup('es-ES')).toEqual('.');` will do it as it fails without your change.

Kaysera: Understood, I misread the name of the function to be tested, sorry

Kaysera: Should I keep the tests for `getNumberFormatGroup` or are they out of scope?

Kaysera: > Adding this `expect(getNumberFormatGroup('es-ES')).toEqual('.');` will do it as it fails without your change.

Also, this test was also added in my previous test cases in line 67 :)

dtslvr (approved):
Thanks again for your contribution, @Kaysera! 😎

Ready to merge now.

Issue #4257, opened 2025-01-30
[BUG] Missing Sectors and Countries for some ETFs


Kaysera:
**Bug Description**

When adding an activity for some ETFs, the Sectors and Countries data is not properly stored, so the Pie Charts are not populated

**To Reproduce**

1. Add a new activity
2. Choose EUNK.DE ETF and add a new Buy
3.  Click on the activity and scroll down to see the graph

**Expected behavior**

The Pie Charts should appear but they do not because the data is not properly populated. This also happens in the Allocations page where a chunk in the groupings By Sector, By Continent and By Market is rendered with No Data Available

**Screenshots**

When selecting the VUAA.L ETF shows the expected behavior

![Image](https://github.com/user-attachments/assets/66c19cd9-f49d-4554-a156-0b2a1e0b592b)

But when selecting the EUNK.DE ETF it does not show anything

![Image](https://github.com/user-attachments/assets/2c0eadc7-bfa2-4c33-94f1-860e984f4e37)

**Logs**

**Environment**

- Ghostfolio Version 2.136.0
- Self-hosted
- Experimental Features disabled
- Brave Browser
- Windows

**Additional context**

I have reproduced the issue in the development environment using Firefox and Linux so it seems to be a problem when storing the data. I have managed to track the issue down to the gathering and enhancing where I find myself at a dead end. 

It seems that `TrackinsightDataEnhancerService` is being used to get the holdings and countries, but there are some tickers it cannot find. `YahooFinanceDataEnhancerService` is able to find the tickers but is not filling the sectors for anything other than a Mutual Fund 

https://github.com/ghostfolio/ghostfolio/blob/954cf765b8fb6bb1000a2b5662795547b48ba8f6/apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts#L200

I can think of two solutions:
- First, let Yahoo Finance complete the sectors if it is able to find it and they have not been completed yet, something like

```javascript
 if (!response.sectors && (assetSubClass === AssetSubClass.MUTUALFUND || assetSubClass === AssetSubClass.ETF)) {
```
however Yahoo Finance is not able to find the countries (or I'm unable to see how to do so)

- Second, we can find a similar ticker for Track Insight that has the same ISIN code. For EUNK there is a ticker named SMEA in a different market that has the same properties (although different price, but we are filling in just the ETF sectors/holdings and countries, should not matter)

[...]

PR #4258, opened 2025-01-30
Feature/Add sectors for ETFs using YahooFinanceDataEnhancer

Kaysera:
Addresses #4257 

This PR adds functionality to extract sectors using ETFs. The issue also talks about countries, which Yahoo does not provide, so it does not fix it completely.


apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -197,7 +197,7 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
         assetProfile.price.symbol
       );
 
-      if (assetSubClass === AssetSubClass.MUTUALFUND) {
+      if (assetSubClass in [AssetSubClass.ETF, AssetSubClass.MUTUALFUND]) {
```
dtslvr, line 200: Is `in` working as expected here?

When I try this
```
if ("yes" in ["yes", "no"]) {
  console.log("in");
} else {
  console.log("out");
}
```
I get `out` 🤔

dtslvr (approved):
Thanks for your contribution, @Kaysera! 👍🏻

PR #4237, opened 2025-01-22
Manual scraper: Fix json mimetype detection (closes #4233)

enen92:
This fixes https://github.com/ghostfolio/ghostfolio/issues/4233 as I was also interested in the same feature.
response.headers is a Headers interface (https://developer.mozilla.org/en-US/docs/Web/API/Headers) so accessing it like `response.headers['content-type']` will always return undefined

Issue #3320, opened 2024-04-22
Support activity in custom currency


dtslvr:
### Status Quo

Activities are always tracked in the currency of the underlying ticker. Although the user can set their own currency, the unit price is immediately converted into the original currency.

### Target Outcome

The unit price should no more be converted in the frontend, when the user choses a custom currency for the activity. Instead, the custom currency is stored in the activity (already done) and the value will be converted on the fly. Ideally in a single place of the code.

### Screenshot

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/52ffd348-9f1b-4299-a9b7-66419077218c" />

### Related Discussions

* https://github.com/ghostfolio/ghostfolio/issues/2700
* https://github.com/ghostfolio/ghostfolio/discussions/3141
* https://github.com/ghostfolio/ghostfolio/discussions/3344
* https://github.com/ghostfolio/ghostfolio/discussions/3786
* https://github.com/ghostfolio/ghostfolio/discussions/4098

jameskitt616:
I dont understand why it finds BTC-USD at yahoo finance, but not BTC-EUR, both exist.
And as a Person who only uses EUR, USD doesnt really help me. Is there anything i can do to add my BTC in EUR?
https://finance.yahoo.com/quote/BTC-USD/
https://finance.yahoo.com/quote/BTC-EUR/


![image](https://github.com/user-attachments/assets/65122f78-b656-46b5-9721-b725d21a4e6f)
![image](https://github.com/user-attachments/assets/04590407-84cd-4daf-b644-485e172b0967)

csehatt741:
@dtslvr, I can take a look at this issue

Dgram19:
Any news about this topic?

swissbuechi:
@jameskitt616 
Yes there is a solution called manual asset profile + scraper.

I'm using this workaround for `BTC-CHF`.
You just need to find a website to download the past data of `BTC-EUR` in a `.csv`, match the import format in excel and configure the scraper to fetch the price in real-time.

Scraper Configuration `BTC-CHF`:
- Mode: `instant`
- Selector: `.push-data`
- URL: `https://www.finanzen.ch/devisen/bitcoin-franken-kurs`

As a bonus, your `Absolute Curreny Performace` in the `Portfolio Analysis` will be fixed too. Since you also track the manual investment in EUR.

If you need assistance for creating a manual `BTC-EUR` configuration, I'd happily create one for you.

[...]

PR #4290, opened 2025-02-06
Yahoo Finance: Allow searching cryptocurrencies in all configured fiat currencies

enen92:
Hey there,

This pull request allows the yahoo finance data provider to also look for cryptocurrencies in other user configured currencies (apart from the default USD): e.g. BTC-EUR, BTC-CAD, etc. It seems counter intuitive that users have to create manual scrapers (pointing to yahoo finance) to simply have access to these symbols (ref: https://github.com/ghostfolio/ghostfolio/pull/4237, https://github.com/ghostfolio/ghostfolio/issues/3320#issuecomment-2585202408).

The change was quite straight forward apart from the fact that currencies are completely tied to `ExchangeRateDataModule` in the codebase. Importing this into the yahoo finance data provider results into circular dependencies. Hence, I've introduced a new module (`CurrencyModule` / `CurrencyService`) that at the moment only moves the user currency list getter previously in the `ExchangeRateDataService` there. Sounded like a possible clean way to go considering that in the future more logic could probably be added to this service. This is handled in https://github.com/ghostfolio/ghostfolio/commit/912f5ad9322e737ab3deb69d05f7afb143d3ac4d.

**Screenshots:**
User with only USD configured as currency:
image

- Can find BTC-USD:
image

- Can't find BTC-EUR:
image

User adds EUR as currency:
image

- User finds both BTC-USD and BTC-EUR:
image
image

### DIFF
--- a/apps/api/src/app/admin/admin.module.ts
+++ b/apps/api/src/app/admin/admin.module.ts
@@ -4,6 +4,7 @@ import { SubscriptionModule } from '@ghostfolio/api/app/subscription/subscriptio
 import { TransformDataSourceInRequestModule } from '@ghostfolio/api/interceptors/transform-data-source-in-request/transform-data-source-in-request.module';
 import { ApiModule } from '@ghostfolio/api/services/api/api.module';
 import { ConfigurationModule } from '@ghostfolio/api/services/configuration/configuration.module';
+import { CurrencyModule } from '@ghostfolio/api/services/currency/currency.module';
 import { DataProviderModule } from '@ghostfolio/api/services/data-provider/data-provider.module';
 import { ExchangeRateDataModule } from '@ghostfolio/api/services/exchange-rate-data/exchange-rate-data.module';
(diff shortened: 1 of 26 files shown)


enen92:
Finally found some time to adjust the format errors and the tests.
@dtslvr any chance you can take a look?

dtslvr (commented):
Hi @enen92,

Thanks a lot for taking the time to put this together! I really appreciate your effort in improving the _Yahoo Finance_ integration and making it more flexible for users.

Instead of solving this just for _Yahoo Finance_, I think it would be better to address it in a more general way, as we also support other data providers like _CoinGecko_. There’s an open issue (#3320) that aims to cover custom currencies (including ETFs), so it would make sense to tackle it there once and for all.

Would you be interested in contributing to that broader solution? Let me know what you think.

dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/issues/3320.

PR #4296, opened 2025-02-07
Feature/improve symbol lookup in Trackinsight data enhancer

### LINKED ISSUE #4257: [BUG] Missing Sectors and Countries for some ETFs

Kaysera:
Fixes #4257 

This PR enhances Trackinsight Data Enhancer in order to search for the adequate symbol. This has two main benefits:

1. It streamlines symbols like `VUAA` and `VUAA.DE`, which before needed two search attempts in case it needed to remove the `.DE` part.
2. It looks for the symbol stored in Trackinsight with the same ISIN as the symbol searched. Tickers can differ between markets to refer to the same ISIN, but the underlying asset is the same. So we can use this to obtain sectors, holdings and countries, which are informations of the asset rather than the market.

### DIFF
--- a/apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
+++ b/apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
@@ -44,28 +44,26 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
       return response;
     }
 
+    let trackinsightSymbol = await this.searchTrackinsightSymbol(
+      symbol,
+      requestTimeout
+    );
+    if (trackinsightSymbol === '') {
+      trackinsightSymbol = await this.searchTrackinsightSymbol(
+        symbol.split('.')?.[0],
+        requestTimeout
+      );
+    }
+
     const profile = await fetch(
-      `${TrackinsightDataEnhancerService.baseUrl}/funds/${symbol}.json`,
+      `${TrackinsightDataEnhancerService.baseUrl}/funds/${trackinsightSymbol}.json`,
       {
         signal: AbortSignal.timeout(requestTimeout)
       }
     )
       .then((res) => res.json())
       .catch(() => {
-        return fetch(
-          `${TrackinsightDataEnhancerService.baseUrl}/funds/${
-            symbol.split('.')?.[0]
-          }.json`,
-          {
-            signal: AbortSignal.timeout(
-              this.configurationService.get('REQUEST_TIMEOUT')
-            )
-          }
-        )
-          .then((res) => res.json())
-          .catch(() => {
-            return {};
-          });
+        return {};
       });
 
     const isin = profile?.isin?.split(';')?.[0];
@@ -75,7 +73,7 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     }
 
     const holdings = await fetch(
-      `${TrackinsightDataEnhancerService.baseUrl}/holdings/${symbol}.json`,
+      `${TrackinsightDataEnhancerService.baseUrl}/holdings/${trackinsightSymbol}.json`,
       {
         signal: AbortSignal.timeout(
           this.configurationService.get('REQUEST_TIMEOUT')
@@ -84,20 +82,7 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     )
       .then((res) => res.json())
       .catch(() => {
-        return fetch(
-          `${TrackinsightDataEnhancerService.baseUrl}/holdings/${
-            symbol.split('.')?.[0]
-          }.json`,
-          {
-            signal: AbortSignal.timeout(
-              this.configurationService.get('REQUEST_TIMEOUT')
-            )
-          }
-        )
-          .then((res) => res.json())
-          .catch(() => {
-            return {};
-          });
+        return {};
       });
 
     if (
@@ -171,6 +156,29 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     }
 
     return Promise.resolve(response);
+  }
+
+  private async searchTrackinsightSymbol(
+    symbol,
+    requestTimeout = this.configurationService.get('REQUEST_TIMEOUT')
+  ) {
+    return await fetch(
+      `https://www.trackinsight.com/search-api/search_v2/${symbol}/_/ticker/default/0/3`,
+      {
+        signal: AbortSignal.timeout(requestTimeout)
+      }
+    )
+      .then((res) => res.json())
+      .then((jsonRes) => {
+        if (jsonRes['results']['count'] === 1) {
+          // Return the only ticker that matches the one in the search
+          return jsonRes['results']['docs'][0]['ticker'];
+        }
+        return '';
+      })
+      .catch(() => {
+        return '';
+      });
   }
 
   public getName() {
--- a/apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
+++ b/apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
@@ -80,9 +80,7 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     const holdings = await fetch(
       `${TrackinsightDataEnhancerService.baseUrl}/holdings/${trackinsightSymbol}.json`,
       {
-        signal: AbortSignal.timeout(
-          this.configurationService.get('REQUEST_TIMEOUT')
-        )
+        signal: AbortSignal.timeout(requestTimeout)
       }
     )
       .then((res) => res.json())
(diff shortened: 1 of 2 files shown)


apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -44,28 +44,26 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+      symbol,
+      requestTimeout
+    );
+    if (trackinsightSymbol === '') {
+      trackinsightSymbol = await this.searchTrackinsightSymbol(
+        symbol.split('.')?.[0],
+        requestTimeout
+      );
+    }
+
```
dtslvr, line 57:
```
if (!trackinsightSymbol) {
  return response;
}
```

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -173,6 +158,29 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     return Promise.resolve(response);
   }
 
+  private async searchTrackinsightSymbol(
+    symbol,
```
dtslvr, line 162:
```
{
  requestTimeout,
  symbol
}: {
  requestTimeout: number;
  symbol: string;
}
```

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -173,6 +158,29 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
     return Promise.resolve(response);
   }
 
+  private async searchTrackinsightSymbol(
```
dtslvr, line 161: Move down ⬇️ (`private` functions should be at the end of the class)

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -173,6 +158,29 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+        signal: AbortSignal.timeout(requestTimeout)
+      }
+    )
+      .then((res) => res.json())
+      .then((jsonRes) => {
+        if (jsonRes['results']['count'] === 1) {
+          // Return the only ticker that matches the one in the search
+          return jsonRes['results']['docs'][0]['ticker'];
+        }
+        return '';
```
dtslvr, line 177: `return undefined;`

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -173,6 +158,29 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+      })
+      .catch(() => {
+        return '';
```
dtslvr, line 180: `return undefined;`

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -44,28 +44,26 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
       return response;
     }
 
+    let trackinsightSymbol = await this.searchTrackinsightSymbol(
+      symbol,
+      requestTimeout
+    );
+    if (trackinsightSymbol === '') {
```
dtslvr, line 51: Could you move this logic directly inside `searchTrackinsightSymbol()`?

dtslvr (changes requested):
Thanks for your contribution, @Kaysera! 😎

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -180,4 +163,51 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
   public getTestSymbol() {
     return 'QQQ';
   }
+
+  private async searchTrackinsightSymbol({
+    requestTimeout = this.configurationService.get('REQUEST_TIMEOUT'),
```
dtslvr, line 168: Just `requestTimeout,` (the assignment is not needed as it is already done in `enhance()`)

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -180,4 +163,51 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+    symbol
+  }: {
+    symbol: string;
```
dtslvr, line 171: Move down ⬇️ (alphabetical order)

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -180,4 +163,51 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+        return undefined;
+      })
+      .catch(() => {
+        return undefined;
+      });
+    if (newSymbol) {
+      return newSymbol;
+    }
+
+    return await fetch(
```
dtslvr, line 195: This duplicates the request above, consider refactoring to avoid repetition (call `searchTrackinsightSymbol()` twice in `enhance()` with different `symbol` parameter)

dtslvr (changes requested):
Thanks for the adaptions, @Kaysera. I have added further comments.

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
dtslvr: > Would something like this be okay?

At first glance, that's fine 👍🏻 I will have a closer look later.

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
```diff
@@ -180,4 +168,33 @@ export class TrackinsightDataEnhancerService implements DataEnhancerInterface {
...
+      `https://www.trackinsight.com/search-api/search_v2/${symbol}/_/ticker/default/0/3`,
+      {
+        signal: AbortSignal.timeout(requestTimeout)
+      }
+    )
+      .then((res) => res.json())
+      .then((jsonRes) => {
+        if (
+          jsonRes['results']?.['count'] === 1 ||
+          jsonRes['results']?.['docs']?.[0]?.['ticker'] === symbol
```
dtslvr, line 189: To support the very basic use case for symbol `VT` with many search results, I have added this additional condition

dtslvr (approved):
Thanks for your contribution, @Kaysera! 🙌🏻

apps/api/src/services/data-provider/data-enhancer/trackinsight/trackinsight.service.ts
Kaysera: Thanks, make sense! When I was playing with it, perfect match ETFs only returned a single symbol (3+ letter ones) so I thought all would do the same. Thanks again!

Issue #4318, opened 2025-02-15
[BUG] Assets missing in Storybook


dtslvr:
**Bug Description**

Running `npm run start:storybook` and opening `http://localhost:4400/?path=/story/logo--default` results in a `404` error for `ghost.svg`.

**Expected Behavior**

Assets are available in _Storybook_.

**Screenshot**

<img width="842" alt="Image" src="https://github.com/user-attachments/assets/0240f488-e46f-4f48-a779-f8ca14379e9a" />

KenTandrian:
@dtslvr Adding [`staticDirs`](https://storybook.js.org/docs/api/main-config/main-config-static-dirs) to the Storybook configuration in `main.js` will resolve the issue. I can work on this.

dtslvr:
I have created a follow-up issue: https://github.com/ghostfolio/ghostfolio/issues/4331

PR #4324, opened 2025-02-15
Bugfix/assets missing in Storybook

KenTandrian:
Hi @dtslvr, this PR is to resolve issue https://github.com/ghostfolio/ghostfolio/issues/4318. Please take a look :)

### Changes
* Added `staticDirs` to the Storybook config.
* Improved config type safety.


libs/ui/.storybook/main.js
```diff
@@ -1,9 +1,16 @@
+/** @type {import('@storybook/angular').StorybookConfig} */
```
dtslvr, line 1: Interesting 🤔

dtslvr (approved):
Well done, @KenTandrian! 😎

libs/ui/.storybook/main.js
KenTandrian: It is not a necessity, but the IntelliSense will definitely be useful! 😄

KenTandrian: We can also convert this file to TypeScript, if it is a better choice :)

dtslvr: > We can also convert this file to TypeScript, if it is a better choice :)

 _TypeScript_ would be even better, however I don't expect that this file is changing a lot 🙂

PR #4328, opened 2025-02-15
Feature/reload available tags after creating custom tag in holding detail dialog

dtslvr:


dtslvr:
@KenTandrian small UX improvement 🙂

Issue #4331, opened 2025-02-16
[FEATURE] Build Storybook and serve it


dtslvr:
Building _Storybook_ should be included in the build workflow and ideally be accessible at `http://localhost:3333/storybook`.

shaun-ak:
Hey @dtslvr , will try this one, can you please assign me, thanks!

PR #4340, opened 2025-02-19
Feature/add Storybook to build process

shaun-ak:
PR for #4331 

@dtslvr , I assumed we would deploy the storybook using github pages, please feel free to modify it.


shaun-ak:
@dtslvr , there is no gh-pages for the story book to be deployed.
Hence, the deploy part will break.

Do we need to deploy it or just a check for building the storybook is enough?

dtslvr (commented):
Thanks for the PR, @shaun-ak!

GitHub Pages is not needed, my issue description might be misleading mentioning GitHub Action. What about adding `build:storybook` to `build:production`, then moving (?) the output folder to the root `./storybook`.

What do you think about it, @KenTandrian and @shaun-ak?

dtslvr:
> Let me know if you have any specific preferences on handling this!

I would prefer to change `package.json` and `libs/ui/project.json` only.

shaun-ak:
@dtslvr , I did not remove the storybook workflow, please re-review.

dtslvr (approved):
Thanks for your contribution, @shaun-ak! 👍🏻

We’ll see how this works in the next release.

dtslvr:
> We’ll see how this works in the next release.

It does not work, I have created a follow-up issue: #4401

Issue #3834, opened 2024-09-28
[BUG] Delete custom currency


dtslvr:
**Bug Description**

The deletion of a custom currency fails.

**To Reproduce**

1. Go to _Admin Control_ -> _Market Data_
2. Click the + button, switch to _Add Currency_ and add `PLN`
3. Switch the preset to _Currencies_ (see _Filter by..._)
4. Delete the asset profile via the action menu

**Expected behavior**

The request `DELETE api/v1/admin/profile-data/YAHOO/USDPLN` completes successfully.

* It should update the array of custom currencies in `Property[PROPERTY_CURRENCIES]` within the request (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-overview/admin-overview.component.ts#L162) for an example).

**Screenshots**

<img width="1230" alt="admin-market-data-delete-currency" src="https://github.com/user-attachments/assets/5a89e12a-8b91-40bf-8876-89ed6fbd0742">

**Logs**

```
[Nest] 8285  - 09/28/2024, 9:53:24 AM   ERROR [ExceptionsHandler] 
Invalid `prisma.symbolProfile.delete()` invocation:

An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
PrismaClientKnownRequestError: 
Invalid `prisma.symbolProfile.delete()` invocation:

An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
```

**Environment**

- Ghostfolio Version 2.110.0
- Self-hosted

dtslvr:
@madhab-chandra-sahoo this is a follow-up task for https://github.com/ghostfolio/ghostfolio/issues/3773. Are you interested to work on it?

madhab-chandra-sahoo:
yes, please assign me this

dtslvr:
Hi @madhab-chandra-sahoo

Any progress on the issue? Let me know if you need support.

madhab-chandra-sahoo:
@dtslvr, sorry I was away & could not work on it, but I will update you in the next 2 days.

dtslvr:
> @dtslvr, sorry I was away & could not work on it, but I will update you in the next 2 days.

Hi @madhab-chandra-sahoo, it’s been a while since your last update. Any progress on the issue? Let me know if you need anything. Thanks!

madhab-chandra-sahoo:
[...]

PR #4354, opened 2025-02-24
Bugfix: Delete custom currency fixed

csehatt741:
Fixed the functionality to delete an asset profile of type currency.

Fixes https://github.com/ghostfolio/ghostfolio/issues/3834

### DIFF
--- a/apps/api/src/app/admin/admin.service.ts
+++ b/apps/api/src/app/admin/admin.service.ts
@@ -117,7 +117,9 @@ export class AdminService {
 
     if (customCurrencies.includes(currency)) {
       const updatedCustomCurrencies = customCurrencies.filter(
-        (aCurrency) => aCurrency !== currency
+        (customCurrency) => {
+          return customCurrency !== currency;
+        }
       );
 
       await this.putSetting(
--- a/apps/api/src/app/admin/admin.service.ts
+++ b/apps/api/src/app/admin/admin.service.ts
@@ -109,8 +109,24 @@ export class AdminService {
     symbol
   }: AssetProfileIdentifier) {
     await this.marketDataService.deleteMany({ dataSource, symbol });
-    await this.symbolProfileService.delete({ dataSource, symbol });
-    await this.deleteCustomCurrency(symbol);
+
+    const currency = getCurrencyFromSymbol(symbol);
+    const customCurrencies = (await this.propertyService.getByKey(
+      PROPERTY_CURRENCIES
+    )) as string[];
+
+    if (customCurrencies.includes(currency)) {
+      const updatedCustomCurrencies = customCurrencies.filter(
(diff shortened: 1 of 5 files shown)


csehatt741:
@dtslvr, could you take a look at this PR?

apps/api/src/app/admin/admin.service.ts
```diff
@@ -110,6 +110,7 @@ export class AdminService {
   }: AssetProfileIdentifier) {
     await this.marketDataService.deleteMany({ dataSource, symbol });
```
dtslvr, line 111: Could you check here if it is a custom currency? (via `this.propertyService.getByKey(PROPERTY_CURRENCIES`))
If yes, filter and update.
If no, delete profile.

apps/api/src/services/symbol-profile/symbol-profile.service.ts
```diff
@@ -24,8 +24,8 @@ export class SymbolProfileService {
   }
 
   public async delete({ dataSource, symbol }: AssetProfileIdentifier) {
```
dtslvr, line 26: With the check above, no change should be needed here.

dtslvr (changes requested):
Thanks for your contribution, @csehatt741!
I have reviewed your changes and added a suggestion.

apps/api/src/services/symbol-profile/symbol-profile.service.ts
csehatt741: Reverted

apps/api/src/app/admin/admin.service.ts
csehatt741: Refactored

apps/api/src/app/admin/admin.service.ts
```diff
@@ -109,7 +109,24 @@ export class AdminService {
...
-    await this.symbolProfileService.delete({ dataSource, symbol });
+
+    const currency = getCurrencyFromSymbol(symbol);
+    const customCurrencies = (await this.propertyService.getByKey(
+      PROPERTY_CURRENCIES
+    )) as string[];
+
+    if (customCurrencies.includes(currency)) {
+      const updatedCustomCurrencies = customCurrencies.filter(
+        (aCurrency) => aCurrency !== currency
```
dtslvr, line 120: Could you use the explicit form here?

```
(customCurrency) => {
  return customCurrency !== currency;
}
```

dtslvr (changes requested):
Thanks for the update, @csehatt741!
It looks fine, just one suggestion to keep the code base consistent.

Could you add a changelog entry to the _Fixed_ section please:
`- Fixed the functionality to delete an asset profile of type currency`

dtslvr (approved):
Ready to merge now.

PR #4355, opened 2025-02-24
Bugfix/handle exception in benchmark service related to unnamed asset profiles

matt-fidd:
When using an unnamed symbol as a benchmark, several pages fail to load and the below message is printed in the server logs.

```
[Nest] 1252  - ERROR [ExceptionsHandler] Cannot read properties of null (reading 'localeCompare')
  TypeError: Cannot read properties of null (reading 'localeCompare')
  at ghostfolio/dist/apps/api/main.js:1:120131
  at Array.sort (<anonymous>)
  at BenchmarkService.getBenchmarkAssetProfiles (ghostfolio/dist/apps/api/main.js:1:120111)
  at async Promise.all (index 0)
  at async InfoService.get (ghostfolio/dist/apps/api/main.js:1:789591)
```

As a sidenote, I couldn't work out how to add a "name" for an manual symbol, it doesn't seem to have autopopulated or anything and the Market Data page is showing as below
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks for your first contribution, @matt-fidd! 😎

> image

If you are able to reproduce this unnamed asset profile, please open a new issue with the steps. Thanks

PR #4370, opened 2025-02-28
Feature/holding table add summary to footer row

dandevaud:
This adds a summary row for all filtered positions to the holdings table (/home/holdings).
It summarizes the Values and the NetPerformanceWithCurrency effect. Additionally using both of these values it calculates the performance in percent:

> (Sum of NetPerf)./(Total Value - Sum of NetPerf.)

### DIFF
--- a/libs/ui/src/lib/holdings-table/holdings-table.component.html
+++ b/libs/ui/src/lib/holdings-table/holdings-table.component.html
@@ -145,6 +145,7 @@
       <td *matFooterCellDef class="px-1" mat-footer-cell>
         <div class="d-flex justify-content-end">
           <gf-value
+            [colorizeSign]="true"
             [isCurrency]="true"
             [locale]="locale"
             [value]="totalChange"
@@ -180,6 +181,7 @@
       <td *matFooterCellDef class="px-1" mat-footer-cell>
         <div class="d-flex justify-content-end">
           <gf-value
+            [colorizeSign]="true"
             [isPercent]="true"
             [locale]="locale"
             [value]="totalChangePercentage"
--- a/libs/ui/src/lib/holdings-table/holdings-table.component.html
+++ b/libs/ui/src/lib/holdings-table/holdings-table.component.html
@@ -64,7 +64,11 @@
           />
         </div>
       </td>
-      <td *matFooterCellDef class="px-1" mat-footer-cell></td>
+      <td
+        *matFooterCellDef
+        class="d-none d-lg-table-cell justify-content-end px-1"
+        mat-footer-cell
+      ></td>
     </ng-container>
 
     <ng-container matColumnDef="valueInBaseCurrency">
@@ -90,7 +94,7 @@
         </div>
       </td>
 
-      <td *matFooterCellDef class="px-1" mat-footer-cell>
+      <td *matFooterCellDef class="d-none d-lg-table-cell px-1" mat-footer-cell>
         <gf-value
           [isCurrency]="true"
           [locale]="locale"
--- a/libs/ui/src/lib/holdings-table/holdings-table.component.html
+++ b/libs/ui/src/lib/holdings-table/holdings-table.component.html
@@ -16,6 +16,7 @@
           [tooltip]="element.name"
         />
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell></td>
     </ng-container>
 
     <ng-container matColumnDef="nameWithSymbol">
@@ -38,6 +39,7 @@
           <small class="text-muted">{{ element.symbol }}</small>
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell>Total</td>
     </ng-container>
 
     <ng-container matColumnDef="dateOfFirstActivity">
@@ -62,6 +64,7 @@
           />
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell></td>
     </ng-container>
 
     <ng-container matColumnDef="valueInBaseCurrency">
@@ -86,6 +89,14 @@
           />
         </div>
       </td>
+
+      <td *matFooterCellDef class="px-1" mat-footer-cell>
+        <gf-value
+          [isCurrency]="true"
+          [locale]="locale"
+          [value]="totalValue"
+        ></gf-value>
+      </td>
     </ng-container>
 
     <ng-container matColumnDef="allocationInPercentage">
@@ -107,6 +118,7 @@
           />
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell></td>
     </ng-container>
 
     <ng-container matColumnDef="performance">
@@ -128,6 +140,15 @@
               isLoading ? undefined : element.netPerformanceWithCurrencyEffect
             "
           />
+        </div>
+      </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell>
+        <div class="d-flex justify-content-end">
+          <gf-value
+            [isCurrency]="true"
+            [locale]="locale"
+            [value]="totalChange"
+          ></gf-value>
         </div>
       </td>
     </ng-container>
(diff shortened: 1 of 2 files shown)


libs/ui/src/lib/holdings-table/holdings-table.component.ts
```diff
@@ -92,6 +96,16 @@ export class GfHoldingsTableComponent implements OnChanges, OnDestroy {
...
     this.dataSource.sort = this.sort;
+    this.totalValue = this.dataSource.data.reduce(
+      (sum, current) => (sum += current.valueInBaseCurrency),
+      0
+    );
+    this.totalChange = this.dataSource.data.reduce(
+      (sum, current) => (sum += current.netPerformanceWithCurrencyEffect),
+      0
+    );
+    this.totalChangePercentage =
```
dtslvr, line 107: I'm not sure if this is mathematically correct, because it is calculated differently than `netPerformancePercent` 🤔

libs/ui/src/lib/holdings-table/holdings-table.component.ts
```diff
@@ -92,6 +96,16 @@ export class GfHoldingsTableComponent implements OnChanges, OnDestroy {
     this.dataSource = new MatTableDataSource(this.holdings);
     this.dataSource.paginator = this.paginator;
     this.dataSource.sort = this.sort;
+    this.totalValue = this.dataSource.data.reduce(
```
dtslvr, line 99: Please move down to keep the alphabetical order

libs/ui/src/lib/holdings-table/holdings-table.component.ts
```diff
@@ -70,6 +70,10 @@ export class GfHoldingsTableComponent implements OnChanges, OnDestroy {
   public isLoading = true;
   public routeQueryParams: Subscription;
 
+  protected totalValue = 0;
```
dtslvr, line 73: Please move down to keep the alphabetical order

dtslvr, line 73: Why `protected` and not `public`?

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dtslvr, line 99: You could use the helper function `getSum()`

libs/ui/src/lib/holdings-table/holdings-table.component.ts
```diff
@@ -92,6 +96,16 @@ export class GfHoldingsTableComponent implements OnChanges, OnDestroy {
     this.dataSource = new MatTableDataSource(this.holdings);
     this.dataSource.paginator = this.paginator;
     this.dataSource.sort = this.sort;
+    this.totalValue = this.dataSource.data.reduce(
+      (sum, current) => (sum += current.valueInBaseCurrency),
+      0
+    );
+    this.totalChange = this.dataSource.data.reduce(
```
dtslvr, line 103: You could use the helper function `getSum()`

libs/ui/src/lib/holdings-table/holdings-table.component.html
```diff
@@ -38,6 +39,7 @@
           <small class="text-muted">{{ element.symbol }}</small>
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell>Total</td>
```
dtslvr, line 42: Add `i18n`

libs/ui/src/lib/holdings-table/holdings-table.component.html
```diff
@@ -86,6 +93,14 @@
           />
         </div>
       </td>
+
+      <td *matFooterCellDef class="d-none d-lg-table-cell px-1" mat-footer-cell>
+        <gf-value
+          [isCurrency]="true"
+          [locale]="locale"
+          [value]="totalValue"
+        ></gf-value>
```
dtslvr, line 102: Use short form:
```
/>
```

libs/ui/src/lib/holdings-table/holdings-table.component.html
```diff
@@ -130,6 +146,16 @@
...
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell>
+        <div class="d-flex justify-content-end">
+          <gf-value
+            [colorizeSign]="true"
+            [isCurrency]="true"
+            [locale]="locale"
+            [value]="totalChange"
+          ></gf-value>
```
dtslvr, line 156: Use short form:
```
/>
```

libs/ui/src/lib/holdings-table/holdings-table.component.html
```diff
@@ -156,6 +182,16 @@
...
         </div>
       </td>
+      <td *matFooterCellDef class="px-1" mat-footer-cell>
+        <div class="d-flex justify-content-end">
+          <gf-value
+            [colorizeSign]="true"
+            [isPercent]="true"
+            [locale]="locale"
+            [value]="totalChangePercentage"
+          ></gf-value>
```
dtslvr, line 192: Use short form:
```
/>
```

libs/ui/src/lib/holdings-table/holdings-table.component.html
```diff
@@ -175,6 +211,11 @@
           })
       "
     ></tr>
+    <tr
+      *matFooterRowDef="displayedColumns"
+      mat-footer-row
+      [ngClass]="{ hidden: isLoading }"
```
dtslvr, line 217: There is currently a gap while loading. I think you need to use `d-none` instead of `hidden`

dtslvr (changes requested):
Thanks for your contribution, @dandevaud!

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: I changed it to use the calculated performance instead

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: Was removed anyway now

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: Was removed anyway now

libs/ui/src/lib/holdings-table/holdings-table.component.html
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.html
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.html
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.html
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.html
dandevaud: done

libs/ui/src/lib/holdings-table/holdings-table.component.ts
dandevaud: As the value is only used in the component (ts & html), I chose protected instead of  public

dtslvr (changes requested):
Thanks for the changes, @dandevaud. During testing, I noticed a few things:

* For `performance`, could we fetch this data from the dedicated performance endpoint? This would help keep the responsibilities of each endpoint more focused.
* It looks like the account detail dialog might be broken with these changes.
* Similarly, the public page seems to be experiencing an issue now.
* Regarding the holdings page, the total value shown in the _Closed Holdings_ view appears to be incorrect. I'm also wondering if displaying a total value there is the intended behavior.

What do you think about conditionally enabling the summary row using a `showSummaryRow` input? This would give us more control over where the total value is displayed.

PR #4387, opened 2025-03-02
Refactor/replace lodash.uniq with Array.from + Set

slagiewka:


apps/client/src/app/components/admin-market-data/create-asset-profile-dialog/create-asset-profile-dialog.component.ts
```diff
@@ -87,7 +86,9 @@ export class CreateAssetProfileDialog implements OnInit, OnDestroy {
         this.createAssetProfileForm.get('addCurrency').value as string
       ).toUpperCase();
 
-      const currencies = uniq([...this.customCurrencies, currency]).sort();
```
slagiewka, line 90: BTW I'm not a fan of spread over arrays instead of push or concat. This can be easily abused and lead to baaad performance. I'm not touching these things this time, though.

dtslvr (approved):
Thanks for your contribution, @slagiewka! 👍🏻

Issue #4374, opened 2025-02-28
Extend export by account balances


dtslvr:
The `accounts` of the [export endpoint](https://github.com/ghostfolio/ghostfolio/tree/main/apps/api/src/app/export) should be extended by `balances` (`date: string`, `value: number`).

Import functionality is not in scope of this issue.

csehatt741:
@dtslvr, I would take a look at this issue

PR #4390, opened 2025-03-03
Feature/Export extended by account balances

csehatt741:
`export` endpoint extended by account balances.

Fixes: #4374

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Extended the export functionality by the account balances
 - Added a _Copy portfolio data to clipboard for AI prompt_ action to the analysis page (experimental)
 
 ### Changed


apps/api/src/app/account/account.service.ts
```diff
@@ -56,13 +62,14 @@ export class AccountService {
...
     (Account & {
+      balances?: AccountBalance[];
       Order?: Order[];
       Platform?: Platform;
     })[]
   > {
     const { include = {}, skip, take, cursor, where, orderBy } = params;
 
-    include.balances = { orderBy: { date: 'desc' }, take: 1 };
+    include.balances = { orderBy: { date: 'desc' } };
```
dtslvr, line 72: As this function is widely used, it is not optimal to always get all balances. It should fetch them based on the argument: either take 1 as before or use the `include` argument.

apps/api/src/app/export/export.service.ts
```diff
@@ -94,7 +96,14 @@ export class ExportService {
 
     return {
       meta: { date: new Date().toISOString(), version: environment.version },
-      accounts,
+      accounts: accounts.map(({ balances, ...account }) => {
+        return {
+          balances: balances.map(({ date, value }) => {
+            return { date: date.toISOString(), value };
+          }),
+          ...account
```
dtslvr, line 104: I would move this up ⬆️

dtslvr (changes requested):
Thanks for your contribution, @csehatt741! 😎

I have reviewed your changes and added a feedback.

apps/api/src/app/account/account.service.ts
csehatt741: Refactored

dtslvr: 👍🏻

dtslvr (approved):
Ready to merge now.

PR #4402, opened 2025-03-05
Bugfix/change client-side dates to be sent in UTC format

Summon528:
When the client requests anything from the server involving the date type, the server response in the format of ISOString.
However, when the client send back the date, it converts date back to string using the date-fns `format` function, which is timezone awared.

Here's an example that's unexpected if the client is running in a timezone that's behind UTC (e.g. US timezone):
`format(parseISO("2014-10-05T00:00:00.000Z"),'yyyy-MM-dd') => 2014-10-04`

This is problematic in the benchmark page, where the client fetch `firstOrderDate`
https://github.com/ghostfolio/ghostfolio/blob/ddb50295b262d1123c808cdc2db47923d3a993c2/apps/client/src/app/pages/portfolio/analysis/analysis-page.component.ts#L235-L236
And submit this value back to the server
https://github.com/ghostfolio/ghostfolio/blob/ddb50295b262d1123c808cdc2db47923d3a993c2/apps/client/src/app/pages/portfolio/analysis/analysis-page.component.ts#L320-L326
Causing the date to be off by one in my case (and also causes the `fetchBenchmark` api to be empty and therefore no benchmark is being shown in my analysis page)

This PR is to make the change so that client send date in UTC.


dtslvr (commented):
Thanks for your contribution, @Summon528! 👍🏻

Starting from version 4, `date-fns` is supporting [time zones](https://github.com/date-fns/date-fns/blob/main/docs/timeZones.md). Do you think `.split('T')[0]` can be replaced with a helper function?

Is `gatherSymbol()` in [AdminService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/services/admin.service.ts) also affected?

Summon528:
- date-fns v4 has support for this with the following syntax, however, it requires a sperate package (`@date-fns/utc`). I'm not sure if you want this, but I can make the change if you want to go this route.
```js
import { parseISO } from 'date-fns';
import { utc } from "@date-fns/utc"
format(date, DATE_FORMAT, { in: utc })
```
- Yeah, gatherSymbol is probably also affected although I'm not sure where it's being called with `date`

dtslvr:
> ```js
> format(date, DATE_FORMAT, { in: utc })
> ```

This makes it very clear, I like it like this! 👍🏻

> * Yeah, gatherSymbol is probably also affected although I'm not sure where it's being called with `date`

It doesn't seem to be needed anymore. Could you remove`date` [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/services/admin.service.ts#L194)?

Summon528:
Just updated the code 👍🏻

dtslvr (approved):
Thanks for the update, @Summon528! 😎
Ready to merge now.

Issue #4413, opened 2025-03-09
Respect filters in export


dtslvr:
When you export activities via the [export endpoint](https://github.com/ghostfolio/ghostfolio/tree/main/apps/api/src/app/export) with filters (e.g. all activities of a holding), it should respect these filters also in the included accounts and tags. Currently it includes every account / tag of the user.

csehatt741:
@dtslvr, I can take a look at this issue

csehatt741:
@dtslvr, how to apply the `assetClasses` and `tags` filters to `accounts`? Do they apply to the `Account.User`?

dtslvr:
Since the export is activity-driven, we could clean up the accounts at the end: check if there's at least one activity linked to each account and only include those that are.

The same approach could be used for tags.

What do you think?

csehatt741:
Yes, makes sense

PR #4425, opened 2025-03-11
Feature/improve export by applying filters on accounts and tags

csehatt741:
Export endpoint accounts and tags are also filtered.

Fixes: #4413

### DIFF
--- a/apps/api/src/app/export/export.service.ts
+++ b/apps/api/src/app/export/export.service.ts
@@ -38,7 +38,7 @@ export class ExportService {
       withExcludedAccounts: true
     });
 
-    if (activityIds) {
+    if (activityIds?.length > 0) {
       activities = activities.filter(({ id }) => {
         return activityIds.includes(id);
       });
@@ -57,7 +57,7 @@ export class ExportService {
       })
     )
       .filter(({ id }) => {
-        return filters.length > 0
+        return activities.length > 0
           ? activities.some(({ accountId }) => {
               return accountId === id;
             })
--- a/apps/api/src/app/export/export.service.ts
+++ b/apps/api/src/app/export/export.service.ts
@@ -39,8 +39,8 @@ export class ExportService {
     });
 
     if (activityIds) {
-      activities = activities.filter((activity) => {
-        return activityIds.includes(activity.id);
+      activities = activities.filter(({ id }) => {
+        return activityIds.includes(id);
       });
     }
 
@@ -56,9 +56,13 @@ export class ExportService {
         where: { userId }
       })
     )
-      .filter((account) =>
-        activities.some(({ accountId }) => accountId === account.id)
-      )
+      .filter(({ id }) => {
+        return filters.length > 0
+          ? activities.some(({ accountId }) => {
+              return accountId === id;
+            })
+          : true;
+      })
       .map(
         ({
           balance,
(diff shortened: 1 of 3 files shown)


apps/api/src/app/export/export.service.ts
```diff
@@ -28,6 +28,22 @@ export class ExportService {
...
+      userCurrency,
+      userId,
+      includeDrafts: true,
+      sortColumn: 'date',
+      sortDirection: 'asc',
+      withExcludedAccounts: true
+    });
+
+    if (activityIds) {
+      activities = activities.filter((activity) => {
```
dtslvr, line 42: `({ id })`

apps/api/src/app/export/export.service.ts
```diff
@@ -28,6 +28,22 @@ export class ExportService {
...
+        return activityIds.includes(activity.id);
```
dtslvr, line 43:
```
return activityIds.includes(id);
```

apps/api/src/app/export/export.service.ts
```diff
@@ -39,57 +55,49 @@ export class ExportService {
...
-        Platform: platform,
-        platformId
-      }) => {
-        if (platformId) {
-          platformsMap[platformId] = platform;
-        }
-
-        return {
+    )
+      .filter((account) =>
```
dtslvr, line 59:
```
.filter(({ id }) =>
  activities.some(({ accountId }) => {
    return accountId === id;
  })
)
```

apps/api/src/app/export/export.service.ts
```diff
@@ -39,57 +55,49 @@ export class ExportService {
...
+          };
+        }
+      );
 
     const tags = (await this.tagService.getTagsForUser(userId))
-      .filter(({ isUsed }) => {
-        return isUsed;
-      })
+      .filter(
+        (tag) =>
```
dtslvr, line 95:
```
({ id, isUsed }) =>
```

apps/api/src/app/export/export.service.ts
```diff
@@ -39,57 +55,49 @@ export class ExportService {
...
+          tag.isUsed &&
+          activities.some((activity) =>
```
dtslvr, line 97:
```
activities.some((activity) => {
  return activity.tags.some(({ id: tagId }) => {
    return tagId === id;
  });
})
```

apps/api/src/app/export/export.service.ts
dtslvr: This removes (cash) accounts with no activities. Should `.filter()` only apply when `filters.length > 0`?

dtslvr (changes requested):
Thank you for this contribution, @csehatt741! 😎

apps/api/src/app/export/export.service.ts
csehatt741: Replaced

apps/api/src/app/export/export.service.ts
csehatt741: Updated

apps/api/src/app/export/export.service.ts
csehatt741: Updated

apps/api/src/app/export/export.service.ts
csehatt741: Updated

apps/api/src/app/export/export.service.ts
csehatt741: Updated

dtslvr (approved):
After some further tweaks it seems to work now. Thanks again for your contribution, @csehatt741! 🙌🏻

Ready to merge.

PR #4445, opened 2025-03-15
Feature/improve symbol validation in Yahoo Finance service

dtslvr:


netcarlos:
Instead of disallowing cryptocurrency pairs just because they contain a dash (-) in their ticker, wouldn't it be better to support them? For example:

GRT6719-USD  
IOTA-USD  

These are all valid Yahoo tickers and should be supported.

dtslvr:
The intention was to fix the silent failure of the import you reported, to make it consistent. This change addresses that specific issue.

For the cryptocurrencies mentioned, I suggest adding them [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/assets/cryptocurrencies/custom.json).

netcarlos:
No, it only covers the underlying bug. This change disallows using a dash '-' in any ticker. Obviously you don't mind that Yahoo provides tickers with dash.

Issue #4477, opened 2025-03-23
Extend emergency fund X-ray rule to support assets


dtslvr:
The _X-ray_ rule **Emergency Fund: Set up** currently considers cash only (see [portfolio.service.ts](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/portfolio/portfolio.service.ts#L1266)). The rule should also be fulfilled for assets.

**To Reproduce**

1. Import the file [ok-btcusd.csv](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok-btcusd.csv)
2. Assign the _Emergency Fund_ tag to the activity
3. Open _Portfolio_ → _X-ray_
4. The rule **Emergency Fund: Set up** shows _No emergency fund has been set up_

**Expected Behavior**

The rule **Emergency Fund: Set up** is fulfilled (_An emergency fund has been set up_)

csehatt741:
Assign it to me, please

dtslvr:
> Assign it to me, please

Appreciate your willingness to help, @csehatt741! 🙂 Given the complexity of #3320 and the need for your expertise on it, would you mind focusing on that for now? We can assign this one to you after completion.

shaun-ak:
Hey @dtslvr , I wish to take a look on this, can you please assign it to me, thanks!

PR #4485, opened 2025-03-25
feature/add support to emergency fund for assets

shaun-ak:
PR for #4477


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1263,7 +1263,8 @@ export class PortfolioService {
         [
           new EmergencyFundSetup(
             this.exchangeRateDataService,
-            userSettings.emergencyFund
+            userSettings.emergencyFund ||
```
dtslvr, line 1266: I will change this, because `EmergencyFundSetup()` expects a `number` instead of a `boolean`.

dtslvr (approved):
Thanks for your contribution, @shaun-ak! 🙂

Issue #4484, opened 2025-03-24
Add checkbox to toggle isActive state in asset profile dialog


dtslvr:
### Goal

Implement a checkbox within the [asset profile dialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog) to allow administrators to easily toggle the `isActive` state of a `SymbolProfile`.

### Screenshot

<img width="797" alt="Image" src="https://github.com/user-attachments/assets/e0f10ecf-0d9d-4b95-ba16-40e52e66cf64" />

tobikugel:
Hi @dtslvr,
Can you assign me that feature?

PR #4497, opened 2025-03-30
Feature/add data gathering toggle to asset profile details dialog

tobikugel:
Hi @dtslvr,
This PR solves #4484.
Can you please review this PR and let me know what you think?

Changes:
- Added checkbox for toggling the isActive property of asset profiles.
- Adjusted update-asset-profile data transfer object
- Added the IsActive property to the backend admin service.


apps/api/src/app/import/import.service.ts
```diff
@@ -546,6 +546,7 @@ export class ImportService {
           SymbolProfile: {
             currency,
             dataSource,
+            isActive: false,
```
dtslvr, line 549: The default value is `true`, I would either get the value from `activitiesDto` or set to the default. What do you think?

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
```diff
@@ -88,6 +89,7 @@ export class AssetProfileDialog implements OnDestroy, OnInit {
     historicalData: this.formBuilder.group({
       csvString: ''
     }),
+    isActive: [false],
```
dtslvr, line 92: `true` ❓

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
```diff
@@ -538,6 +542,20 @@ export class AssetProfileDialog implements OnDestroy, OnInit {
       });
   }
 
+  public onToggleIsActive(event: MatCheckboxChange) {
```
dtslvr, line 545: `{ checked }: MatCheckboxChange`

dtslvr (changes requested):
Thanks for your contribution, @tobikugel! 😎

apps/api/src/app/import/import.service.ts
tobikugel: Definitely. I expected you'd mention that.

tobikugel: I am going to extend the activitiesDto and other necessary Dtos.

This is possible because the import method loads the assetProfiles (and therefore the isActive property) in the validateActivities method.

One thing I found confusing so far is that the Prisma Model SymbolProfile is used for every AssetProfile.
Is this on purpose? In my opinion, it would be better to rename the SymbolProfile model to AssetProfile.

tobikugel:
Hi @dtslvr,
I have made the changes you asked for and added a comment above.

Since I had to adjust some things in the import-activities service, I tested the file import service with the files in the test/import folder.

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
```diff
@@ -512,7 +512,17 @@ <h1 class="flex-grow-1 m-0" mat-dialog-title>
     </form>
```
dtslvr, line 512: Would it make sense to wrap the whole dialog into `<form></form>` (cf. [CreateOrUpdateTagDialog](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-tag/create-or-update-tag-dialog/create-or-update-tag-dialog.html)) and use `isActive` via `formControlName`?
You could get rid of the logic in `onToggleIsActive()` completely.

tobikugel: As for now it is not possible to wrap the whole dialog into a form because it would be nesting another form element (the assetProfileIdentifierForm). This is not possible. [Source](https://stackoverflow.com/questions/379610/can-you-nest-html-forms) 
I implemented the assetProfileIdentifierForm to validate the inputs of the new Asset Profile Identifier.

dtslvr (approved):
Thanks again for your contribution, @tobikugel! 😎

I have simplified the implementation by removing the activity-related logic in the import.

dtslvr:
I have created a follow-up issue: #4508

PR #4498, opened 2025-03-31
Feature/improve check for duplicates in preview step of activities import (comments)

ms32035:
In a situation when stocks are bought or sold in lots of say 100, it is not unusual that multiple lots are sold within the same second in separate transactions. In such cases, the comment field can be used to distinguish between the transactions, for example by putting external trade id.

### DIFF
--- a/apps/api/src/app/import/import.service.ts
+++ b/apps/api/src/app/import/import.service.ts
@@ -526,7 +526,8 @@ export class ImportService {
             activity.quantity === quantity &&
             activity.SymbolProfile.symbol === symbol &&
             activity.type === type &&
-            activity.unitPrice === unitPrice
+            activity.unitPrice === unitPrice &&
+            activity.comment === comment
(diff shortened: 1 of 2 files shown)


apps/api/src/app/import/import.service.ts
```diff
@@ -522,7 +522,8 @@ export class ImportService {
             activity.quantity === quantity &&
             activity.SymbolProfile.symbol === symbol &&
             activity.type === type &&
-            activity.unitPrice === unitPrice
+            activity.unitPrice === unitPrice &&
+            activity.comment === comment
```
dtslvr, line 526: Could you move up to keep the alphabetical order please 🙂

dtslvr (changes requested):
Thanks for your contribution, @ms32035! 😎

dtslvr (approved):

Issue #4496, opened 2025-03-30
Move historical market data editor component into expansion panel in asset profile dialog


dtslvr:
Move the historical market data editor component into an expansion panel in the asset profile dialog (cf. [Scraper Configuration](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html#L347)).

<img width="1154" alt="Image" src="https://github.com/user-attachments/assets/3c9bf677-28f1-4992-8aaa-fec86f40f973" />

tobikugel:
Hi @dtslvr,
can I take over this issue?

dtslvr:
Hi @tobikugel,

Any progress on the issue? Let me know if you need support.

PR #4550, opened 2025-04-17
Feature/add expansion panel for historical market data editor

tobikugel:
Hi @dtslvr,
This PR solves #4496.

Changes:
- Added expansion panel for the historical market data component.
- Added margin on the right of the expansion indicator such that a part of it is not cut off.


dtslvr (approved):
Thanks for your contribution, @tobikugel! 😎

PR #4572, opened 2025-04-20
Bugfix/fix typos in permissions

mclifford82:
Fixed two typos in `permissions.ts`


dtslvr (approved):
Thanks for the fix, @mclifford82 👍🏻

Issue #4566, opened 2025-04-19
Activity in custom currency


dtslvr:
This is a follow-up issue of https://github.com/ghostfolio/ghostfolio/issues/3320.

> As of today's release, this is still incorrect.
> 
> ![Image](https://github.com/user-attachments/assets/38a23159-1024-47e1-a88e-4feb233ed37c)
> 
> ![Image](https://github.com/user-attachments/assets/7e56e6de-218d-4e2b-983b-1d726427f0f0)
> 
> ![Image](https://github.com/user-attachments/assets/684c1aed-7919-415d-82d6-1bf58a9956d6) 

 _Originally posted by @MRobi1 in [#2700](https://github.com/ghostfolio/ghostfolio/issues/2700#issuecomment-2816384192)_

**Environment**

- Ghostfolio Version 2.153.0

csehatt741:
@dtslvr, assign it to me and I'll take a look!

csehatt741:
I've had a looked at this issue and here are my findings:
- Apple Inc. 272.74 CAD is the market price in CAD and it seems to be correct
- The diagram below is in AssetProfile currency (USD), which might be in Activity currency (CAD)
- The Investment value is displayed in AssetProfile currency (USD), which should be in Activity currency (CAD)

@dtslvr, what do you think?

dtslvr:
> * Apple Inc. 272.74 CAD is the market price in CAD and it seems to be correct

I agree, no change needed.

> * The diagram below is in AssetProfile currency (USD), which might be in Activity currency (CAD)

❓

> * The Investment value is displayed in AssetProfile currency (USD), which should be in Activity currency (CAD)

The investment should be in the user's account currency (probably `CAD` in this case). However, the value seems to be calculated in `USD`. I would expect: `1000.00 CAD`

csehatt741:
I would expect the diagram below `Apple Inc. 272.74 CAD` should also be in CAD (user's account currency). Currently it is shown in USD (asset profile currency). Though, I might be wrong.

dtslvr:
> I would expect the diagram below `Apple Inc. 272.74 CAD` should also be in CAD (user's account currency). Currently it is shown in USD (asset profile currency or base currency). Though, I might be wrong.

[...]

PR #4597, opened 2025-04-24
Bugfix/investment calculation for activities in custom currency

csehatt741:
This PR includes:
- min/max price calculation fixed
- tests created for portfolio calculation when activity currency and asset profile currency do not match

Fixes: https://github.com/ghostfolio/ghostfolio/issues/4566


csehatt741:
@dtslvr, fixing portfolio calculation for `Activity in custom currency` is ready to be reviewed.

I haven't found any `Investment` related issues and the calculation is proved to be correct by the newly added tests.

I suspect, cached `TransactionPoints` might be stale, which causes the issue in the deployed application.

test/import/ok-btceur.json
dtslvr, line 1: The unit test needs to pass without these changes. When you import this file via GUI, you get the following errors:
```
activities.0.property feeInAssetProfileCurrency should not exist
activities.0.property unitPriceInAssetProfileCurrency should not exist
```

test/import/ok-btcusd.json
dtslvr, line 1: The unit test needs to pass without these changes. When you import this file via GUI, you get the following errors:
```
activities.0.property feeInAssetProfileCurrency should not exist
activities.0.property unitPriceInAssetProfileCurrency should not exist
```

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -902,8 +902,8 @@ export abstract class PortfolioCalculator {
     let lastTransactionPoint: TransactionPoint = null;
 
     for (const {
-      fee,
       date,
+      fee,
```
dtslvr, line 906: 👍🏻

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-btceur.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+          dataSource: activity.dataSource,
+          name: 'Bitcoin',
+          symbol: activity.symbol
+        }
+      }));
+
+      const portfolioCalculator = portfolioCalculatorFactory.createCalculator({
+        activities,
+        calculationType: PerformanceCalculationType.ROAI,
+        currency: 'EUR',
```
dtslvr, line 128: Could you please set this to `USD`? I suspect that will result in a greater overlap between the two test scenarios.

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-btcusd.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+        ],
+        totalFeesWithCurrencyEffect: new Big('4.46'),
+        totalInterestWithCurrencyEffect: new Big('0'),
+        totalInvestment: new Big('44558.42'),
+        totalInvestmentWithCurrencyEffect: new Big('44558.42'),
+        totalLiabilitiesWithCurrencyEffect: new Big('0'),
+        totalValuablesWithCurrencyEffect: new Big('0')
+      });
+
+      expect(portfolioSnapshot.historicalData.at(-1)).toMatchObject(
```
dtslvr, line 226: This part is currently missing in `portfolio-calculator-btceur.spec.ts`, unsure if needed.

dtslvr (changes requested):
Thanks for your contribution, @csehatt741! 😎 I am confident that the dedicated test scenarios will make the application more robust.

test/import/ok-btceur.json
csehatt741: Moved into tests

test/import/ok-btcusd.json
csehatt741: Moved into test

dtslvr (approved):
Looks fine! Thanks for providing this fix and the test cases, @csehatt741 🙌🏻

ms32035:
@csehatt741 @dtslvr I have some doubts about it. For some reason after this change, in the orders endpoint `fee` and `feeInAssetProfileCurrency` are now equal to each other and both in the Asset Profile currency. I believe `fee` should be what was 
previously `feeInBaseCurrency`

dtslvr:
Thanks for trying out the changes and providing feedback, @ms32035!

Could you please submit a bug report using our template? It would be helpful if you could provide a reproducible scenario, ideally by adapting the existing test case to demonstrate the issue. Thank you.

PR #4641, opened 2025-05-02
MOEX data source implementation

aensidhe:
**Justification**

My main portfolio is on MOEX, so I need a way to manage it. Ghostfolio looks
cool but there's no way to proper manage assets on MOEX. So, first I've found
`moex-iss-api` npm package. Turns out it needed some help, so couple of PR
there. After that I was ready to work on ghostfolio.

This commit implements full `DataProviderInterface` spec:

- Disabled by default, so only those who need it would enabled it.
- We `canHandle` all symbols
- We aren't premium feature
- `getQuotes`, `getHistorical` and `search` were pretty straightforward
- `getTestSymbol` return `SBER` because if something happens to `SBER` stocks
  then MOEX would definetely doesn't matter anymore for sometime.
- `getAssetProfile` and `getDividends` proved to be tricky to implement, so
  I'll cover them below separately.

**getAssetProfile**

**Currency**

This is main method to get info about assets. Unfortunately, due to fall of
USSR and ruble denomination in 1998, we have three currency tickers: 'RUR',
'SUR', 'RUB' For convenience we use 'RUB' instead of all of them. I don't see
practical value to differentiate between them here, but I'm open to suggestions.
Also, some of the tickers do not return currency in which they're listed on
MOEX. Assumed that it's also 'RUB'.

**Name**

Every asset can have several things to identify it. And all of them are
optional in MOEX API, except `secid` which is `Security ID`. So we use them
for name in this order of preference:

1. Latin (usually English) name.
2. Latin short name.
3. Russian name.
4. Security ID.

**Country**

I try to detect country, parsing ISIN: first two letters should be country code.

**Sectors**

MOEX supports some industry related indices, so when we first encounter some
symbol, I check whether it's in those indices and assign sectors accordingly.

**AssetClass and AssetSubClass**

At first, I was tempted to leave them empty, but finally decided to look into.
I downloaded all asset types from MOEX and tried to best of my knowledge assign
asset classes and subclasses. If I wasn't able to find proper relation, I left
the cell empty. After that I took the table (you can check it in the comments
in the code) and made `SecurityTypeMap` interface.

**getDividends**

MOEX API for dividends isn't documented at all (or probably I didn't find
proper docs) and sometimes it doesn't return the newest dividends. Surprisingly,
you can get dividends for MOEX-related assets from YAHOO, but the date can
differ. So, there is heurestic implemented: if those date are no more than two
days apart and payout is the same, then it's the exact same payout and we
merge them.

**Why does migration have date 2024-05-01?**

That's simple. I've started to work on this PR a year ago, but it was not an easy year, so work was delayed. I can change this migration, but it looks good to me at this point.

Signed-off-by: Anatoly Popov <me@aensidhe.ru>


aensidhe:
@dtslvr where can I see/read docs about scraping feature of ghostfolio? And how do I configure my local instance to use it?

aensidhe:
So, I'm looking to `apps/api/src/services/data-provider/manual/manual.service.ts`, where we use scraper configuration.

If I read code correctly, there's no proper way to scrape historical prices using scraper, is there?

And also, dividends feature isn't really implemented, right?

dtslvr:
> If I read code correctly, there's no proper way to scrape historical prices using scraper, is there?

That's correct. You would need to import historical market data manually once.

aensidhe:
Looks like switching to scraping will not only make my setup more complex (+1 service to deploy and monitor), but also it'll become less comfortable experience in general: no automatic dividends and/or history. Also, if my home server will go offline, I'll need to resync history manually for all my assets. 

I get your concerns about support of the feature in future, given you have ghostfolio in cloud as a product. What I propose: merge the feature, keep it disabled by default (and on your platform), assign MOEX tickets (if they will be) to me. At the end, it's open source software, right? So, users shouldn't really expect commercial grade support with SLAs and such. If there would be some big refactoring that will render MOEX not working - fine, not a problem. I'll try and fix it. No SLAs given :)

Anyways, now it's up to you :) Maybe you and the team decide to up the scraper game in near future, who knows.

aensidhe:
@dtslvr so, if you decide against merging this feature, I'd need an easier way to manage my installation. So, I've separated fix commit from this pull request to separate one: https://github.com/ghostfolio/ghostfolio/pull/4715

And small fix to gitlab pipeline so I can override docker repo in my github repo: https://github.com/ghostfolio/ghostfolio/pull/4716

PR #4270, opened 2025-02-01
Feature/update NestJs to version 11 (#4251)

the13-HK:
Sorry @dtslvr, 
I re-created the PR because I made a mistake in rebase and the commit history seemed to be messed up.
There were also multiple commits, so I squashed the commits.(Origin PR is https://github.com/ghostfolio/ghostfolio/pull/4265)

this PR is to resolve issue https://github.com/ghostfolio/ghostfolio/issues/4251. Please check
Confirmed that it passes by running npm test in local

Changes

Update NestJs to version 11
Update cache manager to version 6 due to Nest 11 requirements
Delete Import Milliseconds from cache-manager, because Milliseconds Type has been removed in version 6

PR #4682, opened 2025-05-08
Bugfix/fix ApiKeyStrategy error

andiz2:
# Fix ApiKeyStrategy "done is not a function" Error

## Issue
When using the ApiKeyStrategy with Passport.js and NestJS, an error was occurring:
```
TypeError: done is not a function
    at ApiKeyStrategy.callback [as verify] (/path/to/node_modules/@nestjs/passport/dist/passport/passport.strategy.js:22:21)
```

This error happened because of a parameter mismatch between the NestJS PassportStrategy wrapper and the underlying passport-headerapikey library.

## Root Cause
The issue was in the ApiKeyStrategy constructor where `passReqToCallback` was set to `true`:

```typescript
super({ header: HEADER_KEY_TOKEN, prefix: 'Api-Key ' }, true);
```

When `passReqToCallback` is set to `true`, the HeaderAPIKeyStrategy expects a callback function with signature `(req, apiKey, done)`. However, NestJS's PassportStrategy adapter is providing a callback with signature `(apiKey, done)`. This parameter mismatch caused the `done` callback to be undefined when called.

## Solution
The solution is to set `passReqToCallback` to `false` in the constructor:

```typescript
super({ header: HEADER_KEY_TOKEN, prefix: 'Api-Key ' }, false);
```

This change aligns the expected callback signature with what NestJS's PassportStrategy adapter provides, resolving the error.

## Implementation Details
- Changed `passReqToCallback` from `true` to `false` in the ApiKeyStrategy constructor
- No other code changes were needed as the validate method already had the correct implementation

## Testing
- Confirmed the error no longer occurs
- Verified that authentication still works correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved the language localization for Catalan (`ca`)
 - Upgraded `nestjs` from version `10.4.15` to `11.0.12`
 
+### Fixed
+
+- Fixed an issue with the ApiKeyStrategy that made application break.
+
 ## 2.161.0 - 2025-05-06
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved the language localization for Catalan (`ca`)
 - Upgraded `nestjs` from version `10.4.15` to `11.0.12`
 
-### Fixed
-
-- Fixed an issue with the ApiKeyStrategy that made application break.
-
 ## 2.161.0 - 2025-05-06
 
 ### Added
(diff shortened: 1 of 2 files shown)


the13-HK:
@andiz2 @dtslvr 
Before the migration of nestjs, I implemented the done function myself, but I think the cause of the problem is that the implementation was deleted in this migration.
This fix will change the individual implementations, but will it have any impact on other things?

https://github.com/ghostfolio/ghostfolio/pull/4270/files#diff-76664be3336f989898ec214708777629c25178792f53070b4865ff80cffa127e

andiz2:
@the13-HK I can test to see if it works in other parts of the code, tell me where.

Yesterday I've looked and that were the only part where apikeystrategy was used, but maybe I am wrong. 
Is my first time contributing to this

andiz2:
Thanks for help and quick merge @dtslvr !
Looking forward for more contributions, project is awesome.

PR #4709, opened 2025-05-11
Feature/improve zh localization

KenTandrian:
Hi @dtslvr, I see that there are some incorrect translations. Opening this PR to fix them :)

### Changes
* Fixed translations that had incorrect context. For example, "export activities" was translated to mean logistic goods export instead of data export, and "bond" was translated as relationship instead of asset.
* Translated the "logout" button.


dtslvr (approved):
I have changed the original label from _Logout_ to _Log out_ at the same time. Ready to merge now.

PR #4733, opened 2025-05-22
Bugfix/fix exclude route with wildcard of serve static module

dtslvr:
https://docs.nestjs.com/migration-guide#express-v5


the13-HK:
Hi @dtslvr 
Thank you for fixing related to NestJs11.
I checked other point using wildcard as Express v5 but the other parts were mainly commented out, multiplication, and import.

PR #4737, opened 2025-05-23
Bugfix/rename snake-case hint to kebab-case in paths

dtslvr:


dtslvr:
CC @KenTandrian and @lukasbauer54 🤓

KenTandrian:
I believe we need to rerun `npm run extract-locales`?

Currently the notes look like this (should be kebab-case):
```xlf
<note priority="1" from="description">snake-case</note>
```

PR #4748, opened 2025-05-24
Bugfix/fix Redis cache health check

dtslvr:


dtslvr:
Noticed an issue related to the introduction of `keyv` via the _NestJS_ migration (cc @the13-HK).

PR #4752, opened 2025-05-25
Feature/refactor router links

dtslvr:


dtslvr:
@KenTandrian I have refactored even more router links

Issue #4578, opened 2025-04-20
Localize X-ray rule: FeeRatioInitialInvestment


dtslvr:
This issue aims to set the foundation for the X-ray rules localization (see _Portfolio_ → _X-ray_).

Localize the X-ray rule [FeeRatioInitialInvestment](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts) as a prototype, including interpolation for `thresholdMax` and `feeRatio`.  Use the [i18nService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/i18n/i18n.service.ts) together with the [i18n page](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/pages/i18n/i18n-page.html) for localization on the server.

tobikugel:
Hi @dtslvr,
Yes. However, I will have time to share more about it in about two days.

dtslvr:
I have created a follow-up issue: https://github.com/ghostfolio/ghostfolio/issues/4841

PR #4779, opened 2025-05-28
Feature/Localize X-ray rule

tobikugel:
Hi @dtslvr,

This feature localizes the X-ray rule [FeeRatioInitialInvestment](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts) as described in issue #4578.

There are a couple of points where I’d appreciate your feedback:

I used Unicode references for the curly brackets in the interpolation for the [i18n-page](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/pages/i18n/i18n-page.html) ({ = &#123;, } = &#125;). Does this overall approach make sense to you?

I believe the [i18nService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/i18n/i18n.service.ts) should be an injectable service, so it doesn’t call the loadFiles method for every instantiation. Should I implement it this way?


apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts
```diff
@@ -27,18 +30,30 @@ export class FeeRatioInitialInvestment extends Rule<Settings> {
       : 0;
 
     if (feeRatio > ruleSettings.thresholdMax) {
+      const evaluation = this.i18nService
+        .getTranslation({
+          id: 'rule.fee-ratio-initial-investment.exceed',
+          languageCode: DEFAULT_LANGUAGE_CODE
```
dtslvr, line 36: You could move the replace logic into `getTranslation()` to make it easier for the consumer:
```
this.i18nService.getTranslation({
  id: 'rule.fee-ratio-initial-investment.exceed',
  languageCode: DEFAULT_LANGUAGE_CODE,
  placeholders: {
    feeRatio: (feeRatio * 100).toPrecision(3),
    thresholdMax: (ruleSettings.thresholdMax * 100).toFixed(2)
  }
});
```

apps/client/src/app/pages/i18n/i18n-page.html
```diff
@@ -11,6 +11,14 @@
         performance, portfolio, software, stock, trading, wealth, web3
       </li>
       <li i18n="@@myAccount">My Account</li>
+      <li i18n="@@rule.fee-ratio-initial-investment.exceed">
+        The fees do exceed &#123;thresholdMax&#125;% of your initial investment
```
dtslvr, line 15: Could we use this (less verbose):
```
The fees do exceed ${thresholdMax} of your initial investment
```

apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts
```diff
@@ -1,10 +1,13 @@
 import { RuleSettings } from '@ghostfolio/api/models/interfaces/rule-settings.interface';
 import { Rule } from '@ghostfolio/api/models/rule';
 import { ExchangeRateDataService } from '@ghostfolio/api/services/exchange-rate-data/exchange-rate-data.service';
+import { I18nService } from '@ghostfolio/api/services/i18n/i18n.service';
+import { DEFAULT_LANGUAGE_CODE } from '@ghostfolio/common/config';
 import { UserSettings } from '@ghostfolio/common/interfaces';
 
 export class FeeRatioInitialInvestment extends Rule<Settings> {
   private fees: number;
+  private i18nService = new I18nService();
```
dtslvr, line 10: Changing this to an injectable service (with singleton behavior) sounds like a good idea. Would you consider implementing this change in a separate pull request?

dtslvr (changes requested):
Thanks for your contribution, @tobikugel! 👍🏻
You're on the right track with this. I have reviewed your changes and added a few comments.

apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts
tobikugel: @dtslvr - yes I will

apps/client/src/app/pages/i18n/i18n-page.html
tobikugel: I agree — the interpolation syntax above is the right way to do it.
However, I can’t use it in this case because it’s inside an HTML string and causes a compile error.
Do you have an idea on how I can make it?

dtslvr: Unfortunately, I don't have a solution. Let's keep it like this for now. At least it looks fine in the `messages.*.xlf` files 🙂

dtslvr:
@tobikugel I have reverted the `messages.*.xlf` files (they are generated automatically) and implemented the missing language code. I will test and go through the changes again later.

apps/api/src/models/rules/fees/fee-ratio-initial-investment.ts
dtslvr: PR: https://github.com/ghostfolio/ghostfolio/pull/4829

dtslvr (approved):
Thanks again for the i18n setup for the X-ray rules, @tobikugel! 🙌🏻
Ready to merge now!

Issue #4745, opened 2025-05-24
Import button remains disabled too long after file selection


dtslvr:
**Bug Description**

After selecting activities to import, the _Import_ button stays disabled for too long (it only switches after moving the cursor over the table).

**To Reproduce**

1. Navigate to _Portfolio_ → _Activities_
2. Import the file [ok.json](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok.json)
3. The _Import_ button stays disabled for too long

**Expected Behavior**

The _Import_ button should switch from disabled immediately

**Environment**

- Ghostfolio Version 2.161.0

prosperxo:
hey may i work on this issue?

PR #4791, opened 2025-05-30
fixed import button being disabled on import

prosperxo:
There was a bug #4745 that stated when the import button stayed disabled for too long. 

From my recreation of the bug, it seems the import button was disabled, but then gets enabled whenever an item on the table is selected/deselected. 

With these changes, the button "import" should be enabled right away. 

My approach: 

**1. auto-select all loaded activities**
this.selectedActivities = [...activities];
this.selectedActivities = [...data.activities]; 

- Right after populating the table (dataSource) and set totalItems, we copy the full list of validated activities into selectedActivities

**2. trigger OnPush change detection**
 this.changeDetectorRef.markForCheck();

- this allows the view to see the new selections and re-evaluates the import button state

**3. ensure stepper advances after selection**

stepper.next();

stepper.next is kept in the "finally" block but I moved the new lines to run before it. This allows for when users land on the "Select Activities" step, the import button will already be enabled! 

All in all, no other logic is touched such as error-handling, table setup, etc. This change should only boost the default selection for a butter user interface.

### DIFF
--- a/apps/client/src/app/pages/portfolio/activities/import-activities-dialog/import-activities-dialog.component.ts
+++ b/apps/client/src/app/pages/portfolio/activities/import-activities-dialog/import-activities-dialog.component.ts
@@ -289,6 +289,10 @@ export class ImportActivitiesDialog implements OnDestroy {
             this.activities = activities;
             this.dataSource = new MatTableDataSource(activities.reverse());
             this.totalItems = activities.length;
+
+            this.selectedActivities = [...activities];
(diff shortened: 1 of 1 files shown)


prosperxo:
@dtslvr Hi Thomas! 

Yes that would be a more concise fix as it does not contain duplicated logic!

dtslvr (approved):
Thanks again and congrats on your first contribution, @prosperxo! 🙌🏻

PR #4767, opened 2025-05-27
Feature/improve language localization PL 20250527

halcyonet:
**Description**
This PR updates the `messages.pl.xlf` file to fix a mistake in the Polish translation.

PR #4802, opened 2025-05-31
Feature/improve language localization for UK 20250531

halcyonet:
This PR includes :

- mistaken number fix ( was 100 000, should have been 80 000 )
- fixed two cases where the state was set to "new" while the content was translated.


halcyonet:
@dtslvr Hey, I found another place where the same mistake was made like the last PR I sent #4767. Please review this. Thanks

PR #4849, opened 2025-06-07
Feature/change interpolation syntax in i18n service

dtslvr:


dtslvr:
@tobikugel I have changed the interpolation syntax from `{var}` to `${var}`

PR #4925, opened 2025-06-14
Feature/improve language localization for PT 20250614

olindost:
- Related to #3591


olindost:
Hi @dtslvr, I've translated some parts of the locales to `pt`. Please check my work when you're free. Thank You.

PR #4926, opened 2025-06-14
Bugfix/Google Sign-in hangs on redirect

### LINKED ISSUE #4907: Google Sign-in hangs on redirect
After migration of HtmlTemplateMiddleware (https://github.com/ghostfolio/ghostfolio/pull/4889), the _Google_ Sign-in does not work anymore. After logging in on _Google_, it tries to redirect in [googleLoginCallback](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/auth/auth.controller.ts#L91) to `https://ghostfol.io/en/auth/eyJhb...`, but the page hangs (without error).

KenTandrian:
Hi @dtslvr, this PR resolves #4907. Please take a look :)

### Changes
* Added an exclusion pattern `/:lang/auth/*wildcard` to `app.setGlobalPrefix` in `main.ts`.

This critical change ensures that language-prefixed authentication redirect paths (e.g., `/en/auth/{jwt}`) are correctly handled by the application's middleware chain. By excluding them from the global API prefix, these requests now properly reach `HtmlTemplateMiddleware`, allowing the client-side application to correctly handle authentication redirects and avoid 404 errors or empty responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## 2.171.0-beta.2 - 2025-06-14
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 2.171.0-beta.3 - 2025-06-15
+## Unreleased
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


dtslvr (commented):
Hi @KenTandrian,

Thanks a lot for your investigation and for providing this fix! I created a pre-release and was able to successfully sign in with Google. Great work on that!

However, after further testing, I found another case that fails with the exact same symptoms:

However, after more testing, I realized another case that fails (exactly the same):
1. Open Chrome in Incognito Mode
2. Navigate directly to any (public) sub-route (e.g. `https://ghostfol.io/en/markets`)
3. The response is again `200 OK`, but the page remains blank

Would it help to exclude `'/:lang/*wildcard'`?

KenTandrian:
@dtslvr excluding `'/:lang/*wildcard'` will not work because it would also exclude API routes with more than 1 segment like:
* `/platform/:id`
* `/logo/:dataSource/:symbol`
* `/market-data/:dataSource/:symbol`
* `/tags/:id`
* `/watchlist/:dataSource/:symbol`

Let's define the language codes instead. Can you check if the latest change works now?

dtslvr:
Thanks for the clarification, @KenTandrian! That makes sense.

> Can you check if the latest change works now?

My initial tests look promising so far!

PR #4944, opened 2025-06-15
Feature/switch data provider service to OnModuleInit ensuring (currency) quotes are fetched only once

### LINKED ISSUE #4936: Two instances of DataProviderService
**Bug Description**

It looks like two instances of `DataProviderService` are initialized.

**To Reproduce**

1. Open the file `apps/api/src/services/data-provider/data-provider.service.ts`
2. Add the following code to the constructor: `console.log('Initialize DataProviderService');`
3. Start the server
4. `Initialize DataProviderService` appears twice in the log

**Expected Behavior**

`Initialize DataProviderService` only appears once in the log

**Environment**

- Ghostfolio Version `2.170.0`

monoetienne:
fixes #4936 

Description:

This PR addresses the issue: 
- where `DataProviderService` was being initialized multiple times.
- Introduced an `isInitialized` private flag in DataProviderService.
- Added a check in the `initialize()` method to ensure the service's initialization logic runs only once.
- Prevents duplicate instances of DataProviderService from being created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Changelog
+### Changed
 
+- Adapted the options of the date range selector in the assistant dynamically based on the user’s first activity
 - Switched the data provider service to `OnModuleInit`, ensuring (currency) quotes are fetched only once
 - Migrated the `@ghostfolio/ui/assistant` component to control flow
 - Migrated the `@ghostfolio/ui/value` component to control flow
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -30,7 +30,13 @@ Run `npm run start:server`
 
 ### Start Client
 
-Run `npm run start:client` and open https://localhost:4200/en in your browser
+#### English (Default)
+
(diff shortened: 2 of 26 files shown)


monoetienne:
Hey @dtslvr  — sent a PR your way, would love your feedback when you have time.

dtslvr (commented):
Hi @renegadesolutions,

Thanks for your contribution.

Have you been able to identify the root cause of the issue? I'm wondering if there's a more NestJS-like way to approach it to align with the framework's patterns.

monoetienne:
@dtslvr , the core issue is `DataProviderService` is being initialized twice because initialize() is called in the constructor.  In NestJS, services are typically singletons by default within their module, but if a service is provided in multiple modules or if there are multiple entry points initializing the application context, it can lead to multiple instantiations.

in my next commit, I  used the `OnModuleInit` lifecycle hook. This is a more robust and idiomatic NestJS solution that ensures the initialization logic runs exactly once when the module is initialized by the framework.

please review it, if it works for you and let me know your feedback.

monoetienne:
Hey @dtslvr, I’ve made the changes you suggested which is aligned with NestJS. Please review it when its convenient and let me know if you need anything else or if it's good to merge 🙌

dtslvr (approved):
That looks interesting. Thanks for your contribution, @renegadesolutions! 🙌🏻

PR #4945, opened 2025-06-15
Feature/conditionally display date range options based on user activity

### LINKED ISSUE #4739: Dynamically display date range selector options
The date range selector options in the [GfAssistantComponent](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/assistant) should be conditionally displayed based on `user.dateOfFirstActivity` to provide more relevant choices to the user. For example, if a user's first activity was less than a year ago, the '1Y' option might not be relevant.

Implement it for the options
* `1Y`
* `5Y`

fehranbit:
Fixes #4739

- Hide '1Y' option if user's first activity is less than 1 year ago
- Hide '5Y' option if user's first activity is less than 5 years ago

This change improves the user experience by only showing time ranges that are meaningful based on the user's portfolio history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Adapted the options of the date range selector in the assistant dynamically based on the user’s first activity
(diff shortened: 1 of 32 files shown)


fehranbit:
@dtslvr I have sent a PR #4945 . Kindly give it a look when you can.

PR #4955, opened 2025-06-16
Feature/include admin control panel in quick links of assistant

### LINKED ISSUE #4953: Exclude Admin Control from quick links based on user permission
The _Admin Control_ quick link in the [assistant](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/assistant) should be [excluded](https://github.com/ghostfolio/ghostfolio/blob/main/libs/common/src/lib/routes/routes.ts#L24) dynamically based on the user permission `permissions.accessAdminControl`.

hashpalk:
Fixes #4953 

Changes made:
- Removed static `excludeFromAssistant: true` flag from Admin Control route
- Added dynamic permission check in `searchQuickLinks` method to filter out Admin Control based on `permissions.accessAdminControl`
- The Admin Control quick link will now only appear for users who have the required admin permissions

This ensures that:
- Admin users can still see and access Admin Control through the quick links
- Regular users without admin permissions will not see the Admin Control option in quick links
- The functionality is consistent with the main navigation bar behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Included the admin control panel in the quick links of the assistant
+
 ### Changed
 
 - Adapted the options of the date range selector in the assistant dynamically based on the user’s first activity


libs/common/src/lib/routes/routes.ts
```diff
@@ -21,7 +21,6 @@ export const internalRoutes: Record<string, IRoute> = {
     title: $localize`Settings`
   },
   adminControl: {
-    excludeFromAssistant: true,
```
dtslvr, line 24:
```
excludeFromAssistant: (user: User) => {
  return hasPermission(
    user?.permissions,
    permissions.accessAdminControl
  );
};
```

dtslvr (changes requested):
Thanks for your contribution, @hashpalk.

Your solution addresses the problem. However, I would prefer an approach where the logic is defined in a single place (`routes.ts`).

If it helps, you could refactor e.g. `api` also to a functional concept:

```
api: {
  excludeFromAssistant: (_user: User) => {
    return true;
  },
  path: 'api',
  routerLink: ['/api'],
  title: 'Ghostfolio API'
},
```

Could you please iterate based on my suggestion?

hashpalk:
Good day @dtslvr. I have implemented the Admin Control quick link permission check using a functional approach in routes.ts. I also fixed the permission logic to correctly show or hide the link based on the user's accessAdminControl permission. The solution keeps all route-related logic centralized, as requested.

libs/ui/src/lib/assistant/assistant.component.ts
```diff
@@ -627,6 +628,10 @@ export class GfAssistantComponent implements OnChanges, OnDestroy, OnInit {
 
     const allRoutes = Object.values(internalRoutes)
       .filter(({ excludeFromAssistant }) => {
+        if (isFunction(excludeFromAssistant)) {
+          return excludeFromAssistant(this.user);
```
dtslvr, line 632: I had to invert this to make it work

dtslvr (approved):
Thanks for your contribution, @hashpalk 👍🏻

I have slightly adapted the code. Ready to merge now.

Issue #4605, opened 2025-04-26
[BUG] Wrong percentage values displayed when using impersonation Ids (grant account access)


enen92:
**Important Notice**

The issue tracker is **ONLY** used for reporting bugs. New features should be discussed in our [Slack](https://join.slack.com/t/ghostfolio/shared_invite/zt-vsaan64h-F_I0fEo5M0P88lP9ibCxFg) community or in [Discussions](https://github.com/ghostfolio/ghostfolio/discussions).

Incomplete or non-reproducible issues may be closed, but we are here to help! If you encounter difficulties reproducing the bug or need assistance, please reach out to our community channels mentioned above.

Thank you for your understanding and cooperation!

**Bug Description**

When a user (B) grants access to his account to another user (A), the information is wrongly presented in the Portofolio Analysis page when accessed from A. This is due to the fact the `gf-investment-chart` uses `[isInPercent]="hasImpersonationId || user.settings.isRestrictedView"`. The current code does not take the restricted view permission to the grantee into account and tries to obfuscate the values regardless (although wrong):

![Image](https://github.com/user-attachments/assets/5c9d6266-95e6-4ff3-b60e-1faec7033352)

With 

`[isInPercent]="false"`:

<img width="1052" alt="Image" src="https://github.com/user-attachments/assets/dff89961-46bc-4e5a-bdcc-159e0fa84707" />

The same issue is also shown in other parts of this page:

<img width="1101" alt="Image" src="https://github.com/user-attachments/assets/d3ac69a9-c4dc-491f-8483-13d8b5fd0a5c" />

In the screenshot above, no savings rate displayed, incorrect labels.
In dividends the problem is likely also reproducible but I don't have any assets with dividends on the account I am monitoring.

In the ETF holdings table, the same issue is also presented:

<img width="1185" alt="Image" src="https://github.com/user-attachments/assets/cc176b86-760d-4709-9ae9-ec23094d5cd3" />

Grantee has full read access, not restricted view.

**To Reproduce**

1. Create two accounts (A and B)
2. Give full access to to A from B (My Ghostfolio -> Access)
3. Create a few activities in B
4. Access account B with a login from account A and navigate to the analysis page

**Expected Behavior**

I should be able to view account B info the same of my main account A with real values and not percentages. I should see percentages if restricted view access was granted instead of full access but those should be 0-100%.

**Screenshots**

**Logs**

**Environment**

- Ghostfolio Version (current main)
- Self-hosted
[...]

PR #4958, opened 2025-06-17
Fix Incorrect Percentages in Impersonation View

felixjordandev:
Fixes [#4605](https://github.com/ghostfolio/ghostfolio/issues/4605)

What I have done:

- Removed `hasImpersonationId` from the `isInPercent` binding for `gf-investment-chart` components in `apps/client/src/app/pages/portfolio/analysis/analysis-page.html`.
- The `isInPercent` property now solely uses `user.settings.isRestrictedView`.

What it does:

- When impersonating a user with full access, actual values are displayed.
- Percentages are only shown when the viewed account is specifically set to `isRestrictedView`.
- The chart display aligns with the intended access permissions during impersonation.

### DIFF
--- a/apps/client/src/app/pages/portfolio/analysis/analysis-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/analysis/analysis-page.component.ts
@@ -11,8 +11,8 @@ import {
   ToggleOption,
   User
 } from '@ghostfolio/common/interfaces';
-import { hasPermission, hasReadRestrictedAccessPermission, permissions } from '@ghostfolio/common/permissions';
-import type { AiPromptMode, GroupBy, UserWithSettings } from '@ghostfolio/common/types';
+import { hasPermission, permissions } from '@ghostfolio/common/permissions';
+import type { AiPromptMode, GroupBy } from '@ghostfolio/common/types';
 import { translate } from '@ghostfolio/ui/i18n';
 
 import { Clipboard } from '@angular/cdk/clipboard';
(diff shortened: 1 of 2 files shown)


felixjordandev:
hi, @dtslvr could you please review this?

dtslvr (changes requested):
Thanks for your contribution, @felixjordandev!

I have tested it and noticed that these changes break the impersonation mode: it shows the unnecessary y-axis scale when you receive access with _restricted view_).

**Before**

portfolio-evolution-impersonation-mode-before


**After**

portfolio-evolution-impersonation-mode-new

Could you please take a look?

felixjordandev:
@dtslvr Good day .I have tried few ways to fix this .To fix the impersonation bugs, I can updated the backend authentication logic in jwt.strategy.ts. The new approach checks for an impersonation header on incoming requests, looks up the corresponding access permissions from the database, and then dynamically sets an isRestrictedView flag on the user object for that specific request. This ensures the frontend always has the correct permission status to render the view properly, resolving both the original issue and the regression by centralizing the logic on the backend. Please let me know if I should follow this approach to solve this issue.

felixjordandev:
@dtslvr Good day .I have tried few ways to fix this .To fix the impersonation bugs, I can updated the backend authentication logic in jwt.strategy.ts. The new approach checks for an impersonation header on incoming requests, looks up the corresponding access permissions from the database, and then dynamically sets an isRestrictedView flag on the user object for that specific request. This ensures the frontend always has the correct permission status to render the view properly, resolving both the original issue and the regression by centralizing the logic on the backend. Please let me know if I should follow this approach to solve this issue.

felixjordandev:
@dtslvr Good day .I have tried few ways to fix this .To fix the impersonation bugs, I can updated the backend authentication logic in jwt.strategy.ts. The new approach checks for an impersonation header on incoming requests, looks up the corresponding access permissions from the database, and then dynamically sets an isRestrictedView flag on the user object for that specific request. This ensures the frontend always has the correct permission status to render the view properly, resolving both the original issue and the regression by centralizing the logic on the backend. Please let me know if I should follow this approach to solve this issue.

dtslvr:
Thanks for looking into this and for your suggestion, @felixjordandev.

Updating the backend authentication logic for this bug might be a broader change than intended. Before a solution of this scope, it might be beneficial to re-evaluate the core impersonation logic itself.

PR #5019, opened 2025-06-24
Feature/change entrypoint.sh to run app with main PID

penguineer:
Using exec to call `node main` ensures that SIGTERM and SIGINT are propagated to the app when the container is requested to shut down.


dtslvr (approved):
Thanks for your contribution, @penguineer!

Congratulations, you are Ghostfolio’s 200th contributor! 🎉
As a thank you, we would like to offer you a Ghostfolio Premium subscription for a year. Please reach out to us at hi[at]ghostfol.io to claim it.

Issue #3628, opened 2024-08-02
Ghostfolio'nun Türkçesi


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Turkish-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.tr.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

geooner:
Hello @dtslvr . I would like to work on this issue. Can you please assign me?

dtslvr:
> Hello [@dtslvr](https://github.com/dtslvr) . Can I work on this PR?

Sure @halcyonet, please create a pull request with your translations. As this is an ongoing issue, I don't assign it to you now.

halcyonet:
Hope you are having a good day @dtslvr . I have send the pull request for the #4663 . Please review it and merge if it looks good.

tomaswoj-eth:
@dtslvr Can I work on this?

tomaswoj-eth:
@dtslvr  Can I work on this?

dtslvr:
> [@dtslvr](https://github.com/dtslvr) Can I work on this?

Sure! Feel free to create a PR with the translations, @tomaswoj-eth. Since it is an ongoing effort, I will not formally assign it, but your contributions are very welcome! 🙂

tomaswoj-eth:
@dtslvr I have send a PR #4738 . Could you kindly review?

aleskrin:
@dtslvr I would like to contribute in this issue. Can you assign me please?

dtslvr:
> [@dtslvr](https://github.com/dtslvr) I would like to contribute in this issue. Can you assign me please?

Please create a pull request with your translations, @aleskrin. Since it is an ongoing effort, I will not formally assign it, but your contributions are very welcome! 🙂

aleskrin:
@dtslvr A PR #4844 has been sent. Please give it a look when you can. Thank You.

dtslvr:
I have extended the instructions by a changelog entry.

PR #5030, opened 2025-06-25
Feature/improve language localization for TR 20250625

hashpalk:
#3628

### DIFF
--- a/apps/client/src/locales/messages.tr.xlf
+++ b/apps/client/src/locales/messages.tr.xlf
@@ -7674,7 +7674,7 @@
       </trans-unit>
       <trans-unit id="440264111109852789" datatype="html">
         <source>Live Demo</source>
-        <target state="new">Live Demo</target>
+        <target state="translated">Canlı Demo</target>
         <context-group purpose="location">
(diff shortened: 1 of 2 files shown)


apps/client/src/locales/messages.tr.xlf
```diff
@@ -7674,55 +7674,55 @@
...
+        <target state="translated"> Holdingler </target>
         <context-group purpose="location">
           <context context-type="sourcefile">libs/ui/src/lib/assistant/assistant.html</context>
           <context context-type="linenumber">68</context>
         </context-group>
       </trans-unit>
       <trans-unit id="routes.resources.personalFinanceTools" datatype="html">
         <source>personal-finance-tools</source>
-        <target state="new">personal-finance-tools</target>
+        <target state="translated">kişisel-finans-araçları</target>
```
dtslvr: Is it common to use special characters like `ç` and `ş` in urls? I would suggest the following:
* `ç` → `c`
* `ş` → `s`

hashpalk: alr, I changed it now.

Issue #5050, opened 2025-06-28
Refactor getByKey() in PropertyService


dtslvr:
The current implementation of `getByKey()` in the [PropertyService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/property/property.service.ts) relies on explicit type casting (e.g., `as string[]` in [AdminService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/admin/admin.service.ts)).

```
const customCurrencies = (await this.propertyService.getByKey(PROPERTY_CURRENCIES)) as string[];
```

To enhance type safety and readability, the method should leverage _TypeScript_ generics to explicitly define the expected return type.

```
const customCurrencies = await this.propertyService.getByKey<string[]>(PROPERTY_CURRENCIES);
```

### Implementation Details

* Refactor the `getByKey()` method itself to accept a generic type
* Adapt all existing usages of `getByKey()` and eliminate the casting

csehatt741:
@dtslvr, I'll take a look!

PR #5065, opened 2025-06-30
Feature/refactor getByKey in PropertyService

csehatt741:
Refactored the `PropertyService` by making `getByKey()` function generic.

Fixes: https://github.com/ghostfolio/ghostfolio/issues/5050


apps/api/src/services/property/property.service.ts
```diff
@@ -38,15 +38,14 @@ export class PropertyService {
     return response;
   }
 
-  public async getByKey(aKey: string) {
+  public async getByKey<TKey>(aKey: string): Promise<TKey | undefined> {
```
dtslvr: Why `undefined`? Could we drop the return type?

csehatt741: Return type removed

Issue #3587, opened 2024-07-20
Ghostfolio in Italiano


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Italian-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.it.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

sghebuz:
Hello, I can work on the Italian version, considering I have never worked on the translation of a project I have just a quick question: I see that there are some entries with `state="new"` that are already translated, shall I just verify them and then change their state to `"translated"`? thanks

dtslvr:
> I see that there are some entries with `state="new"` that are already translated, shall I just verify them and then change their state to `"translated"`?

Exactly. I don't know why the `state` sometimes changes while the translation remains the same.

sghebuz:
I'm sorry, is there a recommended tool to edit the xlf file? I tried with poedit but it's rewriting the xml too much. thanks

dtslvr:
> I'm sorry, is there a recommended tool to edit the xlf file?

You could either use a conventional text editor or edit the file directly on [GitHub](https://github.dev/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.it.xlf).

geooner:
@dtslvr Hi, I have made the translations in #4681 . Hope it was helpful.

fabiocarlos97:
@dtslvr  Hello, I have implemented the recent requested Italian translation changes. Please review PR #4717. Thank you

dtslvr:
I have extended the instructions by a changelog entry.

MImran2002:
@dtslvr Hi I am from codepath A301 class can I work on this issue? Is this available for me to take?

dtslvr:
> can I work on this issue?

Sure @MImran2002, please create a pull request with your translations. As this is an ongoing issue, we don't assign it to you now.

PR #5089, opened 2025-07-04
Feature/improve language localization for IT 20250704

geooner:
Translated #3587


geooner:
@dtslvr translated some parts to italian

PR #5152, opened 2025-07-11
Feature/Improve language localization for PT 20250711

olindost:
partially solves #3591


olindost:
hello @dtslvr, here is another PR for the translations. Can you please check?

PR #5154, opened 2025-07-11
Feature/improve language localization for ZH 20250712

imtms:
I have made some improvements to the inaccurate Chinese translations.

> What about using Pinyin for urls? Maybe you can create a separate PR for that.

As a native Chinese speaker, I prefer English to Pinyin when writing URLs in i18n development. There are two main factors that make Pinyin inaccurate and unreadable when writing URLs: lack of tones and prevalence of polyphones. It is not easy to associate Pinyin with specific meanings in Chinese.
Many Chinese words share the same Pinyin spelling but have completely different meanings when tones are applied. For example, "ma" can mean "mother" (mā), "hemp" (má), "horse" (mǎ), or "scold" (mà). In a URL like example.com/ma, it's impossible to accurately convey the intended meaning of the url.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improved the language localization for Chinese (`zh`)
 - Improved the language localization for Dutch (`nl`)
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks for your contribution, @imtms! 👍🏻

Issue #5149, opened 2025-07-10
Extend export by (custom) asset profiles


dtslvr:
The [export endpoint](https://github.com/ghostfolio/ghostfolio/tree/main/apps/api/src/app/export) should optionally include asset profile ids of activities and every asset profile in use. It should only consider custom asset profiles where `userId` is set (for reproduction you can use [sample.json](https://github.com/ghostfolio/ghostfolio/blob/main/test/import/ok/sample.json), but only import _Penthouse Apartment_).

* `activities` should have an `assetProfileId: string?` attribute (technically `symbolProfileId`)
* `assetProfiles` should include all used asset profiles (technically `SymbolProfile`) of that user similar to `accounts`

Import functionality is not in scope of this issue, but an exported file including asset profiles should work without breaking.

csehatt741:
@dtslvr, I'll take a look!

dtslvr:
I have created a follow-up issue: https://github.com/ghostfolio/ghostfolio/issues/5201

PR #5165, opened 2025-07-14
Feature/extend export by custom asset profiles

csehatt741:
Export extend by custom asset profiles.

Fixes: https://github.com/ghostfolio/ghostfolio/issues/5149

### DIFF
--- a/apps/api/src/app/export/export.service.ts
+++ b/apps/api/src/app/export/export.service.ts
@@ -1,6 +1,7 @@
 import { AccountService } from '@ghostfolio/api/app/account/account.service';
 import { OrderService } from '@ghostfolio/api/app/order/order.service';
 import { environment } from '@ghostfolio/api/environments/environment';
+import { MarketDataService } from '@ghostfolio/api/services/market-data/market-data.service';
 import { TagService } from '@ghostfolio/api/services/tag/tag.service';
 import { Filter, Export } from '@ghostfolio/common/interfaces';
 
@@ -12,6 +13,7 @@ import { groupBy, uniqBy } from 'lodash';
 export class ExportService {
   public constructor(
     private readonly accountService: AccountService,
+    private readonly marketDataService: MarketDataService,
     private readonly orderService: OrderService,
     private readonly tagService: TagService
   ) {}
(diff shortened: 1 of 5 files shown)


apps/api/src/app/export/export.service.ts
```diff
@@ -108,6 +108,15 @@ export class ExportService {
         }
       );
 
+    const assetProfiles = uniqBy(
+      activities.map(({ SymbolProfile }) => {
+        return SymbolProfile;
+      }),
+      ({ id }) => {
```
dtslvr, line 121: Could you filter by `userId` to only include custom asset profiles?

apps/api/src/app/export/export.service.ts
```diff
@@ -128,6 +137,54 @@ export class ExportService {
...
+            assetSubClass,
+            comment,
+            countries: countries as unknown as Prisma.JsonArray,
+            currency,
+            cusip,
+            dataSource,
+            figi,
+            figiComposite,
+            figiShareClass,
+            holdings: holdings as unknown as Prisma.JsonArray,
```
dtslvr, line 197: Could you include `marketData`?
```
marketData: Pick<MarketData, 'date' | 'marketPrice'>[];
```

apps/api/src/app/export/export.service.ts
```diff
@@ -145,6 +202,7 @@ export class ExportService {
           unitPrice
         }) => {
           return {
+            assetProfileId: SymbolProfile.userId ? SymbolProfile.id : undefined,
```
dtslvr: I think we can drop `assetProfileId` and change `symbol` to `symbol: SymbolProfile.symbol` to be closer to the internal data model. What do you think?

libs/common/src/lib/interfaces/enhanced-symbol-profile.interface.ts
```diff
@@ -14,6 +14,7 @@ export interface EnhancedSymbolProfile {
   countries: Country[];
   createdAt: Date;
   currency?: string;
+  cusip?: string;
```
dtslvr, line 17: 👍🏻

apps/api/src/app/export/export.service.ts
csehatt741: So do you mean `assetProfiles` should also include only custom ones?

"every asset profile in use" in the description made me uncertain about it.

csehatt741: Filtered

apps/api/src/app/export/export.service.ts
csehatt741: I added `marketData: { date: string; marketPrice: number }[]` to expose `date` as ISO string via the API.

apps/api/src/app/export/export.service.ts
csehatt741: Exporting an `UUID` might not make much sense, so can be a good idea. But as for the current implementation, `symbol` is mapped either from `SymbolProfile.name` or `SymbolProfile.symbol`, which is just another `UUID` for custom asset profiles.

dtslvr:
I have rebased your branch (⚠️  force push) and made a suggestion without `assetProfileId`: https://github.com/ghostfolio/ghostfolio/pull/5165/commits/a53502bda532583a681c220c43209fe9205e5f31

What do you think, @csehatt741?

dtslvr (approved):

PR #5176, opened 2025-07-16
Bugfix/fix clone functionality for custom asset profile activities

dtslvr:


KenTandrian:
Hi @dtslvr, I just realized that this PR brings a breaking change on my end, because now for all assets with `MANUAL` data source (sample symbol: `FR13.F6`), it keeps on creating a new symbol profile every time I create a new order.

So if I understand correctly, with this change, all `MANUAL` assets should have their symbols either start with `GF_` prefix or use UUIDv4?

dtslvr:
Thanks for the feedback, @KenTandrian.

You've correctly identified an unfortunate side effect of this change. The solution is not ideal, however I remember that it's not enough to simply check for a `MANUAL` data source.

Do you have any ideas for a more robust solution?

KenTandrian:
@dtslvr May I understand what is the clone functionality issue being fixed here?

dtslvr:
> @dtslvr May I understand what is the clone functionality issue being fixed here?

It created a new asset profile on clone instead of connecting it with the existing one. With the fix, symbols of type UUID and `GF_*` are handled correctly.

dtslvr:
https://github.com/ghostfolio/ghostfolio/issues/5324 seems to be related.

PR #5220, opened 2025-07-20
Feature/improve language localization for ES 20250720

fabiocarlos97:
translated some more of spanish


fabiocarlos97:
@dtslvr could you check this please

PR #5258, opened 2025-07-27
Bugfix/fix Storybook styles after upgrade

### LINKED ISSUE #5246: Storybook styles are broken after upgrade
It looks like some styles, particularly for _Angular Material_ components, are broken in _Storybook_ after the recent upgrade (#5194). Only development mode is affected, the build works as expected.

**To Reproduce**

1. Run `npm run start:storybook`

KenTandrian:
Hi @dtslvr, this PR resolves #5246. Please take a look :)

### Changes
* Added styles option to `ui:storybook` target.
* Removed `ionicons` loading scripts, since `ionicons` are now directly imported to the components.

### DIFF
--- a/libs/ui/.storybook/preview-head.html
+++ b/libs/ui/.storybook/preview-head.html
@@ -1,8 +1,0 @@
-<script
-  src="https://unpkg.com/ionicons@5.5.1/dist/ionicons/ionicons.esm.js"
-  type="module"
-></script>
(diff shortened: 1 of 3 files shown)


libs/ui/.storybook/preview-head.html
dtslvr, line 1: 👍🏻

PR #5303, opened 2025-08-01
Bugfix/fix style of slide toggle after Angular Material upgrade

dtslvr:


KenTandrian:
@dtslvr just curious, are we planning to migrate to Material 3? Since Angular Material v19 we can use `slide-toggle-overrides` mixin to override styles.

dtslvr:
> @dtslvr just curious, are we planning to migrate to Material 3? Since Angular Material v19 we can use `slide-toggle-overrides` mixin to override styles.

@KenTandrian I have actually started with the migration some time ago: https://github.com/ghostfolio/ghostfolio/pull/3661

Many details need to be fixed before it's ready. Do you have any ideas on how we could best organize this migration? Perhaps we could use that branch as a base, collect a list of glitches and make improvements as separate sub-tasks.

KenTandrian:
@dtslvr agree. I checked out the PR and it actually works, but I see some styling changes. We can work on those in separate tasks.

Issue #5316, opened 2025-08-03
Migrate from Angular Material design 2 to 3


dtslvr:

KenTandrian:
@dtslvr I see that dialogs are using the same background color as before (following `--dark-background: rgb(25, 25, 25)`). Can you point out which dialog should we restore?

KenTandrian:
Hi @dtslvr, I see that we have one task left here: `11 Restore active indicator color of tab to primary color`. Which "tab" are you referring to?

dtslvr:
@KenTandrian, these kind of tabs:

<img width="750" height="1334" alt="Image" src="https://github.com/user-attachments/assets/73268d33-f584-44ac-a54a-928614c550ae" />

KenTandrian:
@dtslvr anything else we should do here?

dtslvr:
I'm merging this into the [next](https://github.com/ghostfolio/ghostfolio/tree/next) branch now. I plan to do some additional testing there, but the current state looks good for now.

KenTandrian:
I see that there are two "theme" SCSS files: [m3-theme.scss](https://github.com/ghostfolio/ghostfolio/blob/4352062154af2684165c4cacfa3c0ba3000cc320/apps/client/src/styles/m3-theme.scss) and [theme.scss](https://github.com/ghostfolio/ghostfolio/blob/4352062154af2684165c4cacfa3c0ba3000cc320/apps/client/src/styles/theme.scss).

Are we going to remove `m3-theme.scss`?

dtslvr:
> Are we going to remove `m3-theme.scss`?

That is a good idea, @KenTandrian! A PR for this cleanup would be much appreciated.

KenTandrian:
@dtslvr Great, I will create a PR for that.

Meanwhile, it seems like changes from PR #6182 is missing in the `next` branch. Any idea why it happened?

dtslvr:
Oh no, the branch management has been a bit chaotic lately and it looks like those changes got lost! Hopefully, things will be smoother moving forward with just the [next](https://github.com/ghostfolio/ghostfolio/tree/next) branch.

My intention was to close this PR.

dtslvr:
Resolved via #6163.

PR #5341, opened 2025-08-09
Feature/decrease border radius and margin of dialogs on mobile

KenTandrian:
Hi @dtslvr, this PR is part of #5316. Please take a look :)

### Changes
- Added `dialog-overrides`.
- Moved some `tabs-overrides` to `.page.has-tabs` so that they only impact the navigation sidebar.

### DIFF
--- a/apps/client/src/styles/theme.scss
+++ b/apps/client/src/styles/theme.scss
@@ -213,6 +213,14 @@ $_tertiary: map.merge(map.get($_palettes, tertiary), $_rest);
     );
   }
 
+  @include mat.dialog-overrides(
+    (
+      container-max-width: 80vw,
+      container-shape: 4px,
+      container-small-max-width: 96vw
+    )
+  );
+
   @include mat.fab-overrides(
     (
       container-color: var(--gf-theme-primary-500)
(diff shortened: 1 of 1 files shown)


apps/client/src/styles/theme.scss
```diff
@@ -213,6 +213,14 @@ $_tertiary: map.merge(map.get($_palettes, tertiary), $_rest);
     );
   }
 
+  @include mat.dialog-overrides(
+    (
+      container-max-width: 80vw,
+      container-shape: 4px,
```
dtslvr: I think the new radius is fine, only on mobile (full-screen) it should be tiny.

What about moving `container-shape: 4px` into `@media (max-width: 575.98px) {}`?

apps/client/src/styles/theme.scss
```diff
@@ -222,14 +230,21 @@ $_tertiary: map.merge(map.get($_palettes, tertiary), $_rest);
     (
```
dtslvr, line 248: Add line break before `@include mat.tabs-overrides(`

dtslvr (changes requested):
I have tried your changes and added a few suggestions.

apps/client/src/styles/theme.scss
```diff
@@ -206,30 +206,51 @@ $_tertiary: map.merge(map.get($_palettes, tertiary), $_rest);
...
-      label-text-tracking: normal
+      inactive-ripple-color: var(--gf-theme-primary-500)
     )
   );
+
+  .page.has-tabs {
+    @include mat.tabs-overrides(
+      (
+        active-indicator-height: 0,
+        container-height: 2rem,
```
dtslvr: Do we need to move this up again? It overrides the `container-height` of `3rem` on mobile.

dtslvr:

Issue #5177, opened 2025-07-16
Filter asset sub class options based on selected asset class in create activity dialog


dtslvr:
When selecting an asset class in the [create activity dialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/pages/accounts/create-or-update-account-dialog) (_Valuable_), the asset sub class dropdown should dynamically update to show only the options that belong to the selected asset class. Reuse the functionality from the asset profile dialog (#5131).

Rustix69:
@dtslvr Can I pick up this Issue ?

dtslvr:
Hi @Rustix69,

Any progress on the issue? Let me know if you need support.

Rustix69:
@dtslvr Sorry i got busy in another stuff, I will process the PR Tomorrow

csehatt741:
@dtslvr, I'll take a look!

PR #5404, opened 2025-08-24
Feature/filter asset sub class options

csehatt741:
Asset sub class options filtered based on selected asset class.

Fixes: https://github.com/ghostfolio/ghostfolio/issues/5177

### DIFF
--- a/libs/common/src/lib/interfaces/index.ts
+++ b/libs/common/src/lib/interfaces/index.ts
@@ -8,6 +8,7 @@ import type {
   AdminMarketDataItem
 } from './admin-market-data.interface';
 import type { AdminUsers } from './admin-users.interface';
+import { AssetClassSelectorOption } from './asset-class-selector-option.interface';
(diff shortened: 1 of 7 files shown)


libs/common/src/lib/interfaces/index.ts
```diff
@@ -8,6 +8,7 @@ import type {
   AdminMarketDataItem
 } from './admin-market-data.interface';
 import type { AdminUsers } from './admin-users.interface';
+import { AssetClassSelectorOption } from './asset-class-selector-option.interface';
```
dtslvr: `import type { AssetClassSelectorOption } from './asset-class-selector-option.interface';` ❓

dtslvr (commented):
Thanks for your contribution, @csehatt741! 😎

libs/common/src/lib/interfaces/index.ts
csehatt741: fixed

PR #5416, opened 2025-08-27
Fix average price for short positions

ms32035:
This improves average price calculation for short positions. I am aware some tests are still to be added, @dtslvr please advise.

A few comments:
1) This only improves average price. In general, many other calculations still remain incorrect for short positions.
2) The following assumptions are made about short positions values:
- quantity - negative
- price - positive
- investment - positive
3) A case where positions are flipped, is still not supported. For example, going from +50 to -50 by selling 100 units, in an ideal world should be broken down into 2 parts for performance calculation

### DIFF
--- a/apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
+++ b/apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
@@ -926,13 +926,21 @@ export abstract class PortfolioCalculator {
           .mul(factor)
           .plus(oldAccumulatedSymbol.quantity);
 
-        if (type === 'BUY') {
+        if (type === 'BUY' && oldAccumulatedSymbol.investment.gte(0)) {
           investment = oldAccumulatedSymbol.investment.plus(
             quantity.mul(unitPrice)
           );
-        } else if (type === 'SELL') {
+        } else if (type === 'BUY' && oldAccumulatedSymbol.investment.lt(0)) {
+          investment = oldAccumulatedSymbol.investment.plus(
+            quantity.mul(oldAccumulatedSymbol.averagePrice)
+          );
+        } else if (type === 'SELL' && oldAccumulatedSymbol.investment.gt(0)) {
           investment = oldAccumulatedSymbol.investment.minus(
             quantity.mul(oldAccumulatedSymbol.averagePrice)
+          );
+        } else if (type === 'SELL' && oldAccumulatedSymbol.investment.lte(0)) {
+          investment = oldAccumulatedSymbol.investment.minus(
+            quantity.mul(unitPrice)
           );
         }
 
@@ -942,8 +950,8 @@ export abstract class PortfolioCalculator {
           investment,
           skipErrors,
           symbol,
-          averagePrice: newQuantity.gt(0)
-            ? investment.div(newQuantity)
+          averagePrice: !newQuantity.eq(0)
+            ? investment.div(newQuantity).abs()
             : new Big(0),
           dividend: new Big(0),
           fee: oldAccumulatedSymbol.fee.plus(fee),


apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -926,14 +926,22 @@ export abstract class PortfolioCalculator {
           .mul(factor)
           .plus(oldAccumulatedSymbol.quantity);
 
-        if (type === 'BUY') {
+        if (type === 'BUY' && oldAccumulatedSymbol.investment.gte(0)) {
```
dtslvr: Could you distinguish inside of `if (type === 'BUY') { ... }`?

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -926,14 +926,22 @@ export abstract class PortfolioCalculator {
...
           investment = oldAccumulatedSymbol.investment.plus(
             quantity.mul(unitPrice)
           );
-        } else if (type === 'SELL') {
```
dtslvr, line 933: Could you distinguish inside of `if (type === 'SELL') { ... }`?

dtslvr (changes requested):
Thanks for the suggestion, @ms32035.

Could you setup a test similar to [portfolio-calculator-btcusd.spec.ts](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-btcusd.spec.ts) by using `loadActivityExportFile()` with e.g. `btcusd-short.json` that demonstrates your scenario(s)?

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -927,13 +927,25 @@ export abstract class PortfolioCalculator {
...
+          } else {
+            investment = oldAccumulatedSymbol.investment.plus(
+              quantity.mul(oldAccumulatedSymbol.averagePrice)
+            );
+          }
         } else if (type === 'SELL') {
-          investment = oldAccumulatedSymbol.investment.minus(
-            quantity.mul(oldAccumulatedSymbol.averagePrice)
-          );
+          if (oldAccumulatedSymbol.investment.gt(0)) {
```
dtslvr, line 940: Why is it here `gt(0)` and  `gte(0)` in `BUY`?

test/import/ok/btcusd-short.json
```diff
@@ -0,0 +1,42 @@
...
+  "platforms": [],
+  "tags": [],
+  "activities": [
+    {
+      "accountId": null,
+      "comment": null,
+      "fee": 4.46,
+      "quantity": 1,
+      "type": "SELL",
+      "unitPrice": 40000.00,
```
dtslvr: Could you use real market prices? (`44558.42` for `2021-12-12`)

test/import/ok/btcusd-short.json
```diff
@@ -0,0 +1,42 @@
...
+      "symbol": "BTCUSD",
+      "tags": []
+    },
+    {
+      "accountId": null,
+      "comment": null,
+      "fee": 4.46,
+      "quantity": 1,
+      "type": "SELL",
+      "unitPrice": 50000.00,
```
dtslvr: Could you use real market prices?

dtslvr (changes requested):

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
ms32035: because if you are at 0 and are selling, then you're opening a short position, which means it should fall into the else case

ms32035:
@dtslvr anything more needed on this?

Issue #5301, opened 2025-08-01
Migrate execution of seed.ts without additional configuration


dtslvr:
As of Node.js `22.18.0`, [type stripping is enabled by default](https://nodejs.org/en/blog/release/v22.18.0#type-stripping-is-enabled-by-default) for `.ts` files. This allows us to execute _TypeScript_ files directly without a separate compilation step.

The goal of this issue is to upgrade the invocation of `seed.ts` to leverage this new functionality by removing the need for `ts-node` for this script.

KenTandrian:
Hi @dtslvr can I take this?

KenTandrian:
@dtslvr The official Docker image for Node.js 22.18.0 is not released yet in Docker Hub. We need to wait for https://github.com/docker-library/official-images/pull/19580 to be merged before we can move forward.

dtslvr:
> [@dtslvr](https://github.com/dtslvr) The official Docker image for Node.js 22.18.0 is not released yet in Docker Hub. We need to wait for [docker-library/official-images#19580](https://github.com/docker-library/official-images/pull/19580) to be merged before we can move forward.

https://github.com/docker-library/official-images/pull/19580 has been merged now.

dtslvr:
The log of Ghostfolio [2.189.0](https://github.com/ghostfolio/ghostfolio/releases) shows the following:

```
Seeding the database

+ echo Seeding the database

+ npx prisma db seed

Running seed command `node prisma/seed.ts` ...

(node:120) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///ghostfolio/apps/api/prisma/seed.ts is not specified and it doesn't parse as CommonJS.

Reparsing as ES module because module syntax was detected. This incurs a performance overhead.

To eliminate this warning, add "type": "module" to /ghostfolio/apps/api/package.json.

(Use `node --trace-warnings ...` to show where the warning was created)
```

I'm a bit confused. Is it still running `npx prisma db seed` (see [entrypoint.sh](https://github.com/ghostfolio/ghostfolio/blob/main/docker/entrypoint.sh#L9))? And does the warning indicate a module import issue?

dtslvr:
I have created a follow-up issue: https://github.com/ghostfolio/ghostfolio/issues/5334

PR #5489, opened 2025-09-08
Task/migrate to prisma configuration file

### LINKED ISSUE #5472: Migrate to Prisma Configuration File
This goal of this issue is to migrate to [Prisma Configuration File](https://www.prisma.io/docs/orm/reference/prisma-config-reference) (`prisma.config.ts`) which will become the default in Prisma 7.

For more information, please see this [video](https://www.youtube.com/shorts/HnTf1yNQ7jM).

DavidReque:
hi @dtslvr, this PR resolves #5472 

### Changes
- Added prisma.config.ts to centralize Prisma's setup, aligning with the upcoming Prisma 7 standard.


prisma.config.ts
```diff
@@ -0,0 +1,11 @@
+import 'dotenv/config';
+import path from 'node:path';
```
dtslvr: `import { join } from 'node:path';`

prisma.config.ts
```diff
@@ -0,0 +1,11 @@
...
+import { defineConfig } from 'prisma/config';
+
+export default defineConfig({
+  schema: path.join('prisma', 'schema.prisma'),
+  migrations: {
+    path: path.join('prisma', 'migrations'),
+    seed: 'npx tsx prisma/seed.mts'
```
dtslvr: In [package.json](https://github.com/ghostfolio/ghostfolio/blob/main/package.json#L214), we are using `node prisma/seed.mts` (without `npx` / `tsx`).

Could you also remove `seed` from `package.json`?

prisma.config.ts
```diff
@@ -0,0 +1,11 @@
+import 'dotenv/config';
+import path from 'node:path';
+import { defineConfig } from 'prisma/config';
+
+export default defineConfig({
+  schema: path.join('prisma', 'schema.prisma'),
+  migrations: {
```
dtslvr, line 6: Move up ⬆️

prisma.config.ts
DavidReque: > In [package.json](https://github.com/ghostfolio/ghostfolio/blob/main/package.json#L214), we are using `node prisma/seed.mts` (without `npx` / `tsx`).
> 
> Could you also remove `seed` from `package.json`?

Hi @dtslvr, 

Thanks for the feedback! I've now removed the `prisma` section from `package.json` as you suggested.

Regarding the seed command, I've tried changing it to `node prisma/seed.mts`, but unfortunately it fails in my local environment with this error:

It seems that Node.js cannot execute the `.mts` file directly and requires a TypeScript runner. The command `npx tsx prisma/seed.mts` works correctly by handling the ES Module TypeScript file.

Would it be acceptable to use the `tsx` command to ensure the seeding process is functional across different environments?

dtslvr: Direct execution of _TypeScript_ has been introduced in _Node.js_ `22.18.0` (see https://github.com/ghostfolio/ghostfolio/issues/5301 for more details). Which version of are you using?

DavidReque: > Direct execution of _TypeScript_ has been introduced in `22.18.0` (see #5301 for more details). Which version of _Node.js_ are you using?

that was the problem, i had a node.js outdated version, now it works for me.
check out the changes

dtslvr (approved):
Thanks again for your contribution, @DavidReque! 🙌🏻

PR #5508, opened 2025-09-11
Feature/move holdings table into holdings section on public page

gmag11:
Maybe holdings close to allocation chart improves data context

### DIFF
--- a/apps/client/src/app/pages/public/public-page.html
+++ b/apps/client/src/app/pages/public/public-page.html
@@ -77,8 +77,6 @@
             [keys]="['symbol']"
             [showLabels]="deviceType !== 'mobile'"
           />
-        </mat-card-content>
-        <mat-card-content>
           <gf-holdings-table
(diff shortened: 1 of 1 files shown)


apps/client/src/app/pages/public/public-page.html
```diff
@@ -78,6 +78,16 @@ <h1 class="h4 mb-3 text-center" i18n>
             [showLabels]="deviceType !== 'mobile'"
           />
         </mat-card-content>
+        <mat-card-content>
```
dtslvr: There should only be one `mat-card-content` element per card. Could you move to the existing one?

gmag11: Done in [1a99fa0](https://github.com/ghostfolio/ghostfolio/pull/5508/commits/1a99fa0e0b8629a0cabbf4bdbd4f7f0e63c3b736)

PR #5492, opened 2025-09-09
Feature/migrate markets page to standalone

DavidReque:
Hi @dtslvr this PR resolves #5478, please take a look

### Changes
- Refactored markets page to standalone.
- Updated changelog

PR #5530, opened 2025-09-15
Feature/add symbol to benchmark component

Batwam:
Hello, 

I would like to propose adding the symbols to the watchlist to help with identification and keep the format consistent with the Holdings (and Activities) lists. 

The code is transposed from [here](https://github.com/ghostfolio/ghostfolio/blob/main/libs/ui/src/lib/holdings-table/holdings-table.component.html#L31-L39) and has been tested using the dev environment.

Before:
image

After:
image


dtslvr (changes requested):
This is a great addition for the watchlist. Thanks for your contribution, @Batwam! 😎

For the markets overview, I believe it would be better to keep the view without symbol.

Could you add `@Input() showSymbol = true;` to `benchmark.component.ts` and conditionally show the symbol in the template?

Then, in `apps/client/src/app/components/home-market/home-market.html` and `app/components/markets/markets.html`, set `[showSymbol]="false"` for the benchmark components.

Batwam:
Thanks for the inputs. I hadn't noticed that it was also missing from the Market list. What is the concern with including symbols to the Market list as well?

Frankly, I'd be inclined to have it there too for consistency and easier identification. In my experience, I rely heavily on tickers names as a lot if ETFs have very similar descriptions. Also, maybe more of an edge case but I believe that it is possible for multiple tickers to have the same Fund Name (eg IWDA, & SWDA).

dtslvr:
The markets overview is designed for high-level market trends, rather than specific assets. It contains a mix of major indices and commodities, not all of which have a known symbol.

You can see the current view here: https://ghostfol.io/en/markets

Batwam:
Hello, I made the changes requested above.

I tried to open both en/markets and en/home/markets to make sure each html worked properly but both pages appears to display the content of home-markets.html. the content of markets.html doesn't seem to be used. In other words, if I edit markets.html, nothing happens. If I edit home-markets.html, both pages get edited.

Is there a reason en/markets and markets.html doesn't seem to appear anywhere?  this makes it difficult to make sure the changes work. Edit: maybe related to this recent change: https://github.com/ghostfolio/ghostfolio/pull/5492

Batwam:
btw, I copied the `text-truncate` class from the original code but I'm looking for the scss file which defines it and I can't find it. same with `text-nowrap`

Probably a stupid question but, is it defined anywhere? search result [here](https://github.com/search?q=repo%3Aghostfolio%2Fghostfolio++text-truncate&type=code)

dtslvr:
> Probably a stupid question but, is it defined anywhere? search result [here](https://github.com/search?q=repo%3Aghostfolio%2Fghostfolio++text-truncate&type=code)

These are style classes from bootstrap (see import [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/styles/bootstrap.scss#L43)).

PR #5570, opened 2025-09-24
Feature/Interactive Overview Graph

Batwam:
Hello,

**Current Status:** At the moment, the overview graph is essentially static, only showing the timeline of % returns `netPerformanceInPercentageWithCurrencyEffect`. underneath the graph, we can see the current value for this indicator, alongside the current total value and Performance.

**Proposal:** I would like to propose this page to be more interactive by allowing to show the historical graph for the 2 other components (`netPerformanceWithCurrencyEffect` and `value`) when clicking on the respective values, allowing to do a deep dive into them: 

- Default clean look remains unchanged, no additional clutter while making this page more useful (which is important as it is the landing page). Default graph type (% Performance) on page opening is also maintained unless the user choses to display a different graph type (see below). 
- minimal code change, reusing the existing dataset and graph function
- No calculations are made, I am simply displaying data readily stored in `historicalDataItems` so it doesn't slow the page down.
- The type of graph is saved automatically so the interface continues to display's the user's preferred graph type
- Graph update on click is disabled if you click on the parameter which is already being displayed on the graph. It is also disabled when in Presenter mode since there is only one value/graph. 

Apologies in advance if this isn't coded the way you might have done it as I'm pretty new to the project. Happy to take any comment to improve.

https://github.com/user-attachments/assets/0d1e05ce-0553-42ab-84bf-38f81f4f5698


Batwam:
Now updated so the type of graph is saved automatically (using `localStorage`) so the interface continues to display the user's preferred graph type each time the page is opened. 

if changing to Presenter mode, the graph reverts to % and once out of Presented, goes back to the previously selected graph type.

Batwam:
@dtslvr is this of interest? I'd really love to see this integrated so the overview page can be more useful.

Batwam:
@dtslvr understand that this is more of a nice to have than an urgent bug fix but is this of interest? That would be good to know so I can close the PR if it's not deemed useful.

Batwam:
Is this of interest?

Issue #5579, opened 2025-09-26
[BUG] Incorrect alignment in some activity table headers


gmag11:
**Bug Description**

Numeric columns in activities table have their header aligned to left while content is aligned to right

**Expected Behavior**

Headers should be aligned as content in teh same column

**Screenshots**

<img width="1121" height="149" alt="Image" src="https://github.com/user-attachments/assets/9b9e113a-729e-4657-af53-cbf90ec3feae" />

**Environment**

- Ghostfolio Version 2.201.0
- Self-hosted
- Experimental Features disabled
- Firefox
- Windows

PR #5580, opened 2025-09-26
Bugfix/improve table headers’ alignment of activities table

gmag11:
Fix proposal for #5579 

Before:
Image

After:
image

The alignment isn’t yet perfect because the mat-sort-header is to the right of the text, but I think it’s fine as it is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
-
-- Make table headers to inherit alignment of column content
-
 ### Added
 
 - Added the safe withdrawal rate to the user settings (experimental)
@@ -19,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved the wording of the 4% rule in the _FIRE_ section
 - Improved the language localization for German (`de`)
+
+### Fixed
+
+- Improved the table headers’ alignment of the activities table
 
 ## 2.203.0 - 2025-09-27
 


libs/ui/src/lib/activities-table/activities-table.component.scss
dtslvr, line 1: What do you think of adding this
```
...
--mat-table-background-color: var(--light-background);

//////////

th {
   .mat-sort-header-container {
     justify-content: inherit;
  }
}

//////////

.mat-footer-row,
...
```
to [table.scss](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/styles/table.scss)?

gmag11: I used the same CSS structure as in holdings-table.component.scss.

Maybe I'm not understanding what you mean. Sorry.

dtslvr: Now I see you followed the existing style of the holdings table component.

Since the style seems to be applicable across many tables, it would be better to move it to `table.scss`. That way, we fix it once for all tables, rather than fixing each individual component. What do you think?

gmag11: I will test it

gmag11: I've moved this CSS rule under `@mixin gf-table` in [table.scss](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/styles/table.scss), removing it from accounts, activities, benchmark and holdings tables CSS.

It seems not to do anything as right aligned content columns get their headers aligned to the left.

I cannot find where this rule is overwritten.

gmag11: I've removed the rule from table and created a mixing that is reused on those 4 tables. Check if you like that solution.

dtslvr: The mixin solution is slightly better, but we would still need to include it in every component.

Have you tried adding
```
th {
   .mat-sort-header-container {
     justify-content: inherit;
  }
}
```
to [table.scss](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/styles/table.scss) without `::ng-deep { ... }`?

gmag11: That works. Thanks

PR #5650, opened 2025-10-01
Feature/extend holdings endpoint by performance with currency effect for cash

### LINKED ISSUE #5612: Extend holdings endpoint by performance with currency effect for cash
The [holdings](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/portfolio/portfolio.controller.ts#L395) endpoint includes `netPerformancePercentWithCurrencyEffect` and `netPerformanceWithCurrencyEffect` for all held assets. These metrics reflect the impact of currency fluctuations differing from your base currency even if the markets are closed.

However, these metrics are currently missing for cash positions (`LIQUIDITY`) derived from accounts’ cash balances. The goal of this is to extend the holdings endpoint to include these metrics for all cash holdings.

KenTandrian:
Hi @dtslvr, this PR resolves #5612. Please take a look :)

### Changes
* Extended holdings endpoint by performance with currency effect for cash:
  * The `getDetails` method in `PortfolioService` now fetches historical exchange rate data for all currencies associated with the user's cash accounts. This is done once per call to ensure efficiency.
  * The `getCashPositions` method has been updated to accept this historical data. It then calculates the `netPerformanceWithCurrencyEffect` for each cash position by comparing the current value of the cash in the base currency against its value at the time the account was created.
  * The `netPerformancePercentWithCurrencyEffect` is then derived from `netPerformanceWithCurrencyEffect` value relative to the initial investment.
* Updated changelog.
* Added comments to clarify the new logic.

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -662,9 +662,26 @@ export class PortfolioService {
       };
     }
 
+    const { endDate, startDate } = getIntervalFromDateRange(
+      dateRange,
+      portfolioCalculator.getStartDate()
+    );
+
+    // Gather historical exchange rate data for all currencies in cash positions
+    const exchangeRatesByCurrency =
+      await this.exchangeRateDataService.getExchangeRatesByCurrency({
+        currencies: [
+          ...new Set(cashDetails.accounts.map(({ currency }) => currency))
+        ],
+        endDate,
+        startDate,
+        targetCurrency: userCurrency
+      });
+
     if (filters?.length === 0 || isFilteredByAccount || isFilteredByCash) {
       const cashPositions = this.getCashPositions({
         cashDetails,
+        exchangeRatesByCurrency,
         userCurrency,
         value: filteredValueInBaseCurrency
       });
@@ -690,6 +707,7 @@ export class PortfolioService {
     ) {
       const emergencyFundCashPositions = this.getCashPositions({
         cashDetails,
+        exchangeRatesByCurrency,
         userCurrency,
         value: filteredValueInBaseCurrency
       });
@@ -1625,10 +1643,14 @@ export class PortfolioService {
 
   private getCashPositions({
     cashDetails,
+    exchangeRatesByCurrency,
     userCurrency,
     value
   }: {
     cashDetails: CashDetails;
+    exchangeRatesByCurrency: Awaited<
+      ReturnType<ExchangeRateDataService['getExchangeRatesByCurrency']>
+    >;
     userCurrency: string;
     value: Big;
   }) {
@@ -1650,24 +1672,48 @@ export class PortfolioService {
         continue;
       }
 
+      const exchangeRates =
+        exchangeRatesByCurrency[`${account.currency}${userCurrency}`];
+
+      // Calculate the performance of the cash position including currency effects
+      const netPerformanceWithCurrencyEffect = new Big(account.balance)
+        .mul(exchangeRates?.[format(new Date(), DATE_FORMAT)] ?? 1)
+        .minus(
+          new Big(account.balance).mul(
+            exchangeRates?.[format(account.createdAt, DATE_FORMAT)] ?? 1
+          )
+        )
+        .toNumber();
+
       if (cashPositions[account.currency]) {
         cashPositions[account.currency].investment += convertedBalance;
+        cashPositions[account.currency].netPerformanceWithCurrencyEffect +=
+          netPerformanceWithCurrencyEffect;
         cashPositions[account.currency].valueInBaseCurrency += convertedBalance;
       } else {
         cashPositions[account.currency] = this.getInitialCashPosition({
           balance: convertedBalance,
           currency: account.currency
         });
+
+        cashPositions[account.currency].netPerformanceWithCurrencyEffect =
+          netPerformanceWithCurrencyEffect;
       }
     }
 
     for (const symbol of Object.keys(cashPositions)) {
-      // Calculate allocations for each currency
+      // Calculate allocations and net performances for each currency
       cashPositions[symbol].allocationInPercentage = value.gt(0)
         ? new Big(cashPositions[symbol].valueInBaseCurrency)
             .div(value)
             .toNumber()
         : 0;
+      cashPositions[symbol].netPerformancePercentWithCurrencyEffect =
+        cashPositions[symbol].investment > 0
+          ? new Big(cashPositions[symbol].netPerformanceWithCurrencyEffect)
+              .div(cashPositions[symbol].investment)
+              .toNumber()
+          : 0;
     }
 
     return cashPositions;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Extended holdings endpoint by performance with currency effect for cash
+
 ### Changed
 
 - Improved the language localization for German (`de`)


apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1650,24 +1672,48 @@ export class PortfolioService {
         continue;
       }
 
+      const exchangeRates =
+        exchangeRatesByCurrency[`${account.currency}${userCurrency}`];
+
+      // Calculate the performance of the cash position including currency effects
+      const netPerformanceWithCurrencyEffect = new Big(account.balance)
```
dtslvr: I think we might have a date range issue for the calculation here.

Currently, the net performance seems to be calculated from `account.createdAt` to `new Date()`. This yields to a high number if the account is 3 years old and we are only looking at the period between `startDate` and `endDate` (`1d`).

Should we calculate the net performance specifically between the `startDate` and `endDate` parameters instead?

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1625,10 +1643,14 @@ export class PortfolioService {
 
   private getCashPositions({
     cashDetails,
+    exchangeRatesByCurrency,
     userCurrency,
     value
   }: {
     cashDetails: CashDetails;
+    exchangeRatesByCurrency: Awaited<
```
dtslvr: How about we introduce an interface for `getExchangeRatesByCurrency()`?

KenTandrian: Agree!

apps/api/src/app/portfolio/portfolio.service.ts
KenTandrian: yes you are right, aligned!

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1650,24 +1683,51 @@ export class PortfolioService {
...
       }
 
+      const exchangeRates =
+        exchangeRatesByCurrency[`${account.currency}${userCurrency}`];
+
+      // Calculate the performance of the cash position including currency effects
+      const netPerformanceWithCurrencyEffect = new Big(account.balance)
+        .mul(exchangeRates?.[format(endDate, DATE_FORMAT)] ?? 1)
+        .minus(
+          new Big(account.balance).mul(
```
dtslvr: Thanks for the update, @KenTandrian!

I tested it again and at first glance, it looks fine.

However, I think you've brought to light a deeper problem with how the system treats cash holdings. At the moment, this calculation assumes the account balance remains constant.

Shouldn't it be weighted similarly to stock holdings? Treating cash holdings the same way as assets in the performance calculator seems to be the better (but more complex) long-term solution. What do you think?

KenTandrian: @dtslvr Thanks for the detailed feedback. You've raised a very important point, and after digging into it, I completely agree. My initial implementation had a significant flaw.

To make sure I'm on the same page, let me summarize my understanding of the problem with an example:

Let's say my base currency is USD and I have a cash account in EUR.

   * Jan 1st: The account has 10,000 EUR. The exchange rate is 1 EUR = 1.10 USD.
   * Jan 30th: I deposit another 5,000 EUR. Let's assume the rate is 1 EUR = 1.115 USD on this day.
   * Jan 31st: The final balance is 15,000 EUR. The exchange rate is 1 EUR = 1.12 USD.

My previous logic was using the final balance for the entire period, calculating `(15,000 * 1.12) - (15,000 * 1.10)`, which equals $300. This doesn't reflect the actual performance, as it incorrectly applies the currency gain to the 5,000 EUR that wasn't in the account for the whole month.

The correct performance should be calculated by breaking the period around the cash flow:
   1. Performance before deposit (Jan 1 - Jan 30):
      The gain on the initial 10,000 EUR is `(10,000 * 1.115) - (10,000 * 1.10) = $150`.

   2. Performance after deposit (Jan 30 - Jan 31):
      The gain on the new total of 15,000 EUR is `(15,000 * 1.12) - (15,000 * 1.115) = $75`.

   3. Total Correct Performance:
      The total is `$150 + $75 = $225`.

As you suggested, the correct approach is to treat cash with the same time-weighted methodology as other assets. This will allow it to correctly account for the timing of cash flows and calculate the accurate performance figure of $225. Does this match your understanding?

Thanks again for the insightful catch!

dtslvr: Apologies for the delay @KenTandrian, I hadn't realized you were waiting for my feedback.

I agree with your detailed analysis and example.

Regarding implementation, the existing portfolio calculator already handles this time-weighted calculation for assets.

To avoid duplicating that complex logic, it might be possible to preprocess the cash balances and then pass this data to the existing portfolio calculator. This would treat the cash as a "synthetic" asset per currency.

KenTandrian:
@dtslvr I have reverted my previous changes and created a new method called `getCashActivities`, which basically fetches the balances of each account and iterate through them to generate synthetic activities.

I have also modified the `getDetails` method to use these synthetic activities. However, somehow it stucks at portfolio calculator creation without any error. I can see that `computeTransactionPoints` is executed successfully but there is no response from `getDetails`. Any idea why is it happening?

dtslvr:
Have you tested with a user account with activities or only cash balances? My initial idea was that it might block calculations if there is no activity. However, I was not able to fix it.

What I've observed while trying to find the cause:
* Instead of `YAHOO` we should use `this.dataProviderService.getDataSourceForExchangeRates()` [here](https://github.com/ghostfolio/ghostfolio/pull/5650/files#diff-b76b6c44c1bfd51ef638e858cc1226d3f730918ed2ffd4bf6890f3f97313cbdeR364)
* `[...activities, ...cashActivities]` → the order by date might be critical for subsequent calculations
* `cashActivities` might be also required in `getPerformance()` [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/portfolio/portfolio.service.ts#L1152)

KenTandrian:
I realized that `this.calculatorFactory.createCalculator()` is being called in many places, while I am only handling one of them. Therefore, the cash synthetic orders calculation is moved to `OrderService` so that it takes effect to all of them.

dtslvr:
Let’s make another attempt at getting this finalized. What is the current state, @KenTandrian?

KenTandrian:
hi @dtslvr, sorry for the delayed finalization of this PR.
The PR is ready to review. You can also take a look at the changes in the PR description. Thanks!

Anyway, I think we should work on adding JSDoc documentations on some important methods so that contributors (and also potentially AI agents) can understand better.

apps/api/src/app/order/order.service.ts
```diff
@@ -317,6 +323,108 @@ export class OrderService {
...
+  }: {
+    cashDetails: CashDetails;
+    userCurrency: string;
+    userId: string;
+  }): Promise<ActivitiesResponse> {
+    const activities: Activity[] = [];
+
+    for (const account of cashDetails.accounts) {
+      const { balances } = await this.accountBalanceService.getAccountBalances({
+        filters: [{ id: account.id, type: 'ACCOUNT' }],
```
dtslvr: Move down ⬇️

apps/api/src/app/order/order.service.ts
```diff
@@ -317,6 +323,108 @@ export class OrderService {
...
+        userCurrency,
+        userId
+      });
+
+      let currentBalance = 0;
+      let currentBalanceInBaseCurrency = 0;
+      for (const balanceItem of balances) {
```
dtslvr, line 357: Add line break

apps/api/src/app/order/order.service.ts
```diff
@@ -317,6 +323,108 @@ export class OrderService {
...
+          activities.push({
+            ...syntheticActivityTemplate,
+            quantity: currentBalance - balanceItem.value,
+            type: 'SELL',
+            value: currentBalance - balanceItem.value,
+            valueInBaseCurrency:
+              currentBalanceInBaseCurrency - balanceItem.valueInBaseCurrency
+          });
+        }
+        currentBalance = balanceItem.value;
```
dtslvr, line 420: Add line break

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator.ts
```diff
@@ -294,6 +298,7 @@ export class RoaiPortfolioCalculator extends PortfolioCalculator {
       itemType: 'start',
       quantity: new Big(0),
       SymbolProfile: {
+        assetSubClass: isCash ? 'CASH' : undefined,
```
dtslvr: Move down ⬇️

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator.ts
```diff
@@ -307,6 +312,7 @@ export class RoaiPortfolioCalculator extends PortfolioCalculator {
       feeInBaseCurrency: new Big(0),
       itemType: 'end',
       SymbolProfile: {
+        assetSubClass: isCash ? 'CASH' : undefined,
```
dtslvr: Move down ⬇️

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator.ts
```diff
@@ -347,6 +353,7 @@ export class RoaiPortfolioCalculator extends PortfolioCalculator {
           feeInBaseCurrency: new Big(0),
           quantity: new Big(0),
           SymbolProfile: {
+            assetSubClass: isCash ? 'CASH' : undefined,
```
dtslvr: Move down ⬇️

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -136,12 +136,15 @@ export abstract class PortfolioCalculator {
           }
 
           return {
-            SymbolProfile,
             tags,
             type,
             date: format(date, DATE_FORMAT),
             fee: new Big(feeInAssetProfileCurrency),
             quantity: new Big(quantity),
+            SymbolProfile: {
```
dtslvr: Why is this needed?

KenTandrian: It's actually not needed. Thanks for pointing that out!

apps/api/src/app/order/order.service.ts
```diff
@@ -43,7 +47,9 @@ import { randomUUID } from 'node:crypto';
 export class OrderService {
   public constructor(
     private readonly accountService: AccountService,
+    private readonly accountBalanceService: AccountBalanceService,
```
dtslvr, line 49: Move up ⬆️

apps/api/src/app/order/order.service.ts
```diff
@@ -317,6 +323,110 @@ export class OrderService {
...
+            assetSubClass: 'CASH',
+            countries: [],
+            createdAt: new Date(balanceItem.date),
+            currency: account.currency,
+            dataSource:
+              this.dataProviderService.getDataSourceForExchangeRates(),
+            holdings: [],
+            id: account.currency,
+            isActive: true,
+            sectors: [],
```
dtslvr, line 385: `name: account.currency,` ❓

The names for currencies are also missing in `GET api/v1/portfolio/holdings`. I don't know why but you can see it in the holding table: ✅ _USD (Cash) `USD`_ vs. ❌ `USD` only

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  let configurationService: ConfigurationService;
+  let currentRateService: CurrentRateService;
+  let exchangeRateDataService: ExchangeRateDataService;
+  let portfolioCalculatorFactory: PortfolioCalculatorFactory;
+  let portfolioSnapshotService: PortfolioSnapshotService;
+  let redisCacheService: RedisCacheService;
+
+  beforeAll(() => {
+    exportResponse = loadExportFile(
```
dtslvr: This is not really in use as data is hard-coded below...

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      jest.useFakeTimers().setSystemTime(parseDate('2025-01-01').getTime());
+
+      // Override activities with synthetic cash orders
+      // Scenario from User:
+      // 2023-12-31T00:00:00.000Z: 1000
+      // 2024-12-31T00:00:00.000Z: 2000
+
+      const accountId = exportResponse.accounts[0].id; // Use exportResponse to simulate tying to real account
+
+      const syntheticActivities: Activity[] = [
```
dtslvr: Could we reuse `getOrdersForPortfolioCalculator()` somehow to not hard code this? 🤔

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+          date: parseDate('2023-12-31'),
+          feeInAssetProfileCurrency: 0,
+          quantity: 1000,
+          type: 'BUY',
+          SymbolProfile: {
+            ...symbolProfileDummyData,
+            assetSubClass: 'CASH',
+            currency: 'USD',
+            dataSource: 'MANUAL',
+            name: 'US Dollar',
```
dtslvr: `name: 'USD'`

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+          date: parseDate('2024-12-31'),
+          feeInAssetProfileCurrency: 0,
+          quantity: 1000, // +1000 to reach 2000 total
+          type: 'BUY',
+          SymbolProfile: {
+            ...symbolProfileDummyData,
+            assetSubClass: 'CASH',
+            currency: 'USD',
+            dataSource: 'MANUAL',
+            name: 'US Dollar',
```
dtslvr: `name: 'USD'`

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+          },
+          unitPriceInAssetProfileCurrency: 1
+        }
+      ];
+
+      jest
+        .spyOn(exchangeRateDataService, 'getExchangeRatesByCurrency')
+        .mockResolvedValue({
+          USDCHF: {
+            '2023-12-31': 0.85,
```
dtslvr: Could you move this to the existing `exchange-rate-data.service.mock.ts`?

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+            '2024-01-01': 0.86,
+            '2024-12-31': 0.9,
+            '2025-01-01': 0.91
+          }
+        });
+
+      jest.spyOn(currentRateService, 'getValues').mockResolvedValue({
+        dataProviderInfos: [],
+        values: [],
+        errors: []
```
dtslvr: Move up ⬆️

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+        activities: syntheticActivities,
+        calculationType: PerformanceCalculationType.ROAI,
+        currency: 'CHF', // User default currency overridden to CHF
+        userId: userDummyData.id
+      });
+
+      const portfolioSnapshot = await portfolioCalculator.computeSnapshot();
+
+      const historicalData20231231 = portfolioSnapshot.historicalData.find(
+        (d) => d.date === '2023-12-31'
```
dtslvr: `({ date }) => date === '2023-12-31'`

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      );
+      const historicalData20240101 = portfolioSnapshot.historicalData.find(
+        (d) => d.date === '2024-01-01'
```
dtslvr: `({ date }) => date === '2024-01-01'`

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      );
+      const historicalData20241231 = portfolioSnapshot.historicalData.find(
+        (d) => d.date === '2024-12-31'
```
dtslvr: `({ date }) => date === '2024-12-31'`

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash-performance.spec.ts
KenTandrian: yeah, we are only using it for getting `accountId`. I am replacing it with `randomUUID()`

apps/api/src/app/order/order.service.ts
KenTandrian: `name: account.currency` added, checking for `GET api/v1/portfolio/holdings`

KenTandrian:
@dtslvr I have made some changes to show the names for currencies in the holdings table.

### Problem
Cash assets (e.g., USD, CHF) were appearing with blank names in the holdings list. This occurred because `PortfolioService.getDetails` relies on `assetProfile.name`, which is `undefined` for currency positions.

This happened because `assetProfile` depends on `symbolProfiles`, which somehow does not include cash.

https://github.com/ghostfolio/ghostfolio/blob/95cbd01a4c5becac550c69a69f74fa35fcf5502d/apps/api/src/app/portfolio/portfolio.service.ts#L556-L558

Previously, cash assets were not included in `positions`, and therefore it is okay that `symbolProfiles` does not include cash. Cash positions were previously inserted through this block of codes in `apps/api/src/app/portfolio/portfolio.service.ts`:
```ts
if (filters?.length === 0 || isFilteredByAccount || isFilteredByCash) {
  const cashPositions = this.getCashPositions({
    cashDetails,
    userCurrency,
    value: filteredValueInBaseCurrency
  });

  for (const symbol of Object.keys(cashPositions)) {
    holdings[symbol] = cashPositions[symbol];
  }
}
```

Now that we have changed the logic to calculate cash performance in the same pipeline, this block of codes is removed.

### Solution

I have implemented a new method called `getCashSymbolProfiles` for cash assets. The symbol profiles returned from this method will be pushed into `symbolProfiles` array so that `symbolProfileMap` includes cash assets.

dtslvr (approved):
Huge thanks for your persistence on this one, @KenTandrian! 🙌🏻
Really appreciate the hard work to get it merged.

Issue #5419, opened 2025-08-28
Import of (custom) asset profile fails


dtslvr:
**Bug Description**

The import of a (custom) asset profile fails.

**To Reproduce**

1. Import the following file: [penthouse-apartment.json](https://github.com/user-attachments/files/22031387/penthouse-apartment.json)
2. The _Name_ column stays empty
3. Click the _Import_ button
4. The import fails

**Expected Behavior**

The _Name_ column shows `Penthouse Apartment` and the activity is imported successfully.

**Environment**

- Ghostfolio Version 2.194.0
- Cloud and Self-hosted

inevs:
@dtslvr I can reproduce the error, let me try to fix it

dtslvr:
I have created a follow-up issue: https://github.com/ghostfolio/ghostfolio/issues/5740

PR #5670, opened 2025-10-03
fix issue #5419: Import of (custom) asset profile fails

inevs:
Root Causes

  Issue 1: Import Validation Failure
  - The validateActivities() method (called at apps/api/src/app/import/import.service.ts#L374) validates activities before asset profiles from the import file are created in the database (apps/api/src/app/import/import.service.ts#L245)
  - For MANUAL data source, ManualService.getAssetProfile() queries the database, which returns undefined for new profiles
  - Validation at apps/api/src/app/import/import.service.ts#L774 requires assetProfile.name to exist, causing import to fail

  Issue 2: Empty Name Column
  - Even if validation passes, the validateActivities() method only checked for the existence of the name but didn't populate it from the import data
  - The asset profile object used for display lacked the complete profile information from the import file

Solution

  Modified apps/api/src/app/import/import.service.ts:

  1. Pass import data to validation (apps/api/src/app/import/import.service.ts#L376)
    - Added assetProfilesWithMarketDataDto parameter to validateActivities() call
  2. Check import data during validation (apps/api/src/app/import/import.service.ts#L755)
    - When asset profile doesn't exist in database (no name), check if it exists in import data
    - If found in import data, merge all asset profile fields (name, currency, assetClass, etc.) into the validation object
    - This ensures the name and other fields are available for display in the UI
  3. Maintain validation rules (apps/api/src/app/import/import.service.ts#L788)
    - Keep existing validation that requires name for non-MANUAL BUY, DIVIDEND, and SELL activities
    - Separate the data population logic from validation logic

### DIFF
--- a/apps/api/src/app/import/import.service.ts
+++ b/apps/api/src/app/import/import.service.ts
@@ -373,6 +373,7 @@ export class ImportService {
 
     const assetProfiles = await this.validateActivities({
       activitiesDto,
+      assetProfilesWithMarketDataDto,
       maxActivitiesToImport,
       user
     });
@@ -698,10 +699,12 @@ export class ImportService {
 
   private async validateActivities({
     activitiesDto,
+    assetProfilesWithMarketDataDto,
     maxActivitiesToImport,
     user
   }: {
     activitiesDto: Partial<CreateOrderDto>[];
+    assetProfilesWithMarketDataDto: ImportDataDto['assetProfiles'];
     maxActivitiesToImport: number;
     user: UserWithSettings;
   }) {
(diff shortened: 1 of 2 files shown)


apps/api/src/app/import/import.service.ts
```diff
@@ -749,6 +752,37 @@ export class ImportService {
           )?.[symbol]
         };
 
+        if (!assetProfile?.name) {
+          const assetProfileInImport = assetProfilesWithMarketDataDto?.find(
```
dtslvr, line 756: `assetProfileWithMarketDataDto` ❓

apps/api/src/app/import/import.service.ts
```diff
@@ -749,6 +752,37 @@ export class ImportService {
...
+            (profile) =>
+              profile.dataSource === dataSource && profile.symbol === symbol
```
dtslvr:
```
{ return profile.dataSource === dataSource && profile.symbol === symbol; }
```

apps/api/src/app/import/import.service.ts
```diff
@@ -749,6 +752,37 @@ export class ImportService {
...
+          );
+
+          if (assetProfileInImport) {
+            Object.assign(assetProfile, {
```
dtslvr, line 766: Could you add a comment? Something like:
```
Merge all fields of custom asset profiles in the into the validation object
```

apps/api/src/app/import/import.service.ts
```diff
@@ -749,6 +752,37 @@ export class ImportService {
...
+              assetClass: assetProfileInImport.assetClass,
+              assetSubClass: assetProfileInImport.assetSubClass,
+              comment: assetProfileInImport.comment,
+              countries: assetProfileInImport.countries,
+              currency: assetProfileInImport.currency ?? assetProfile.currency,
```
dtslvr: Why is this the only special case?

dtslvr (changes requested):
Thanks for your contribution, @inevs! 👍🏻

Could you also add a changelog entry please? Thanks.

inevs:
@dtslvr thanks for your review. I applied your feedback

PR #5716, opened 2025-10-10
Bugfix/fix server startup message to properly display IPv6 addresses

slagiewka:
This allows to click and open from the log since v6 addresses are now properly formatted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Refactored various components to use self-closing tags
-- Refactored server started message to print IPv6 addresses correctly
+
+### Fixed
+
+- Fixed the server startup message to properly display IPv6 addresses
 
 ## 2.207.0 - 2025-10-08
 


apps/api/src/main.ts
```diff
@@ -90,7 +90,17 @@ async function bootstrap() {
 
   await app.listen(PORT, HOST, () => {
     logLogo();
-    Logger.log(`Listening at http://${HOST}:${PORT}`);
+
+    let address = app.getHttpServer().address();
+    if (typeof address === 'object') {
```
dtslvr, line 96: Add line break

apps/api/src/main.ts
```diff
@@ -90,7 +90,17 @@ async function bootstrap() {
...
+      const addressObj = address;
+      let host = addressObj.address;
+      if (addressObj.family === 'IPv6') {
```
dtslvr: Add line break

apps/api/src/main.ts
```diff
@@ -90,7 +90,17 @@ async function bootstrap() {
...
+        host = `[${addressObj.address}]`;
+      }
+      address = `${host}:${addressObj.port}`;
```
dtslvr: Add line break

apps/api/src/main.ts
```diff
@@ -90,7 +90,17 @@ async function bootstrap() {
...
+    }
+    Logger.log(`Listening at http://${address}`);
```
dtslvr, line 107: Add line break

apps/api/src/main.ts
```diff
@@ -90,7 +90,17 @@ async function bootstrap() {
 
   await app.listen(PORT, HOST, () => {
     logLogo();
-    Logger.log(`Listening at http://${HOST}:${PORT}`);
+
+    let address = app.getHttpServer().address();
+    if (typeof address === 'object') {
+      const addressObj = address;
```
dtslvr: `addressObject`

slagiewka:
Added these code style changes, but I wasn't sure if you meant before or after the line it was commented on. Seems unnatural to me, but let's go anyway.

PR #5792, opened 2025-10-19
Bugfix/database seed

### LINKED ISSUE #5783: Database seed fails with Error: P1000 due to unresolved env variables
Following the development instructions, the database setup (`npm run database:setup`) command fails due to an authentication error:
```
Error: P1000: Authentication failed against database server, the provided database credentials for `${POSTGRES_USER}` are not valid.
```

The root cause appears to be that the environment variables specified within the [DATABASE_URL](https://github.com/ghostfolio/ghostfolio/blob/main/.env.dev#L15) are not being substituted correctly before the connection is attempted.

**Environment**

- Ghostfolio `2.209.0`

Fullstack-Arshad:
resolved #5783

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added support for data gathering by date range in the asset profile details dialog of the admin control panel
 
+### Fixed
+
+- Fixed an issue in the database seeding process caused by unresolved environment variables in `DATABASE_URL`
+
 ## 2.209.0 - 2025-10-18
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,6 +65,8 @@
         "countries-list": "3.1.1",
         "countup.js": "2.9.0",
         "date-fns": "4.1.0",
+        "dotenv": "17.2.3",
+        "dotenv-expand": "12.0.3",
         "envalid": "8.1.0",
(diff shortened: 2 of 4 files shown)


dtslvr (commented):
Interesting!

Do you think it's also possible to achieve the same natively with the integrated environment variables handling of _Node.js_, @Fullstack-Arshad?

Fullstack-Arshad:
I am not 100% sure, but it should be possible, because dotenv alone does not support environment variable interpolation, but combining it with dotenv-expand does support it.

dtslvr:
Thanks for looking into this, @Fullstack-Arshad.

I would prefer a native _Node.js_ solution if possible. Otherwise, we would need to explicitly add `dotenv` / `dotenv-expand` to our dependencies in `package.json`.

Fullstack-Arshad:
could you please specify why do you need a nodejs native solution exactly. While working on this issue I haven't installed any dependency separately as there is no change in package.json file, and also these libraries are already a part of your dependencies,
as you can see in this image.
image

this could be the reasonn it worked for me on my local without installing it separately.

dtslvr:
Yes, `dotenv` and `dotenv-expand` are available transitively.

> could you please specify why do you need a nodejs native solution exactly.

My preference is driven by reducing external dependencies.

dtslvr (approved):
As a native solution does not seem to be possible at the moment, we will proceed with this approach.

Thanks for your contribution, @Fullstack-Arshad! 🙌🏻

Issue #5721, opened 2025-10-10
Support connecting to Redis over IPv6


slagiewka:
Currently it's not possible for Ghostfolio to connect to Redis running in IPv6 only network.

Not sure what's wrong and what are possible solutions. I will try to figure something out over the weekend, but I'm lodging a ticket so this doesn't get lost.

DB connects just fine, so this is and #5712 would be the only remaining pieces for v6-only deployment. This is of course not including any Internet services that might be serving on v4 only.

PR #5813, opened 2025-10-21
Task/upgrade ioredis to forfeit overriding defaults

slagiewka:
This follows up #5721 which was a working fix, but since I fixed the upstream, the defaults are now sensible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Extracted the portfolio filter form of the assistant to a reusable component
 - Formatted the holdings table in the _Copy AI prompt to clipboard for analysis_ action on the analysis page (experimental)
 - Formatted the holdings table in the _Copy portfolio data to clipboard for AI prompt_ action on the analysis page (experimental)
+- Reverted the explicit configuration of the _Redis_ address family in the job queue module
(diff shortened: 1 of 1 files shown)


package-lock.json
dtslvr, line 1: I appreciate the fix, but this does not seem to be a robust long-term solution.

Could you submit an `ioredis` version upgrade to https://github.com/optimalbits/bull via a PR?

Until that upstream PR is merged, please use [npm overrides](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) as a temporary fix.

slagiewka: Why would I use overrides? The range provided by `bull` is `^5.3.2` which covers `5.8.2`. If it wasn't for Ghostfolio's lockfile any new install would get the latest version and be done with it. 

Running override will also yield the same lockfile change. No transitive dependency needs to be updated this way if there's no range conflict somewhere. Not to mention that a manual override is even less robust than just running `npm update ioredis` to bump the lockfile.

dtslvr: Got it, @slagiewka 👍🏻
Is there a command to refresh the lockfile for a specific dependency?

slagiewka: Yes, `npm update ioredis` is the easiest. It doesn't control to which version it's going to be updated, though. 
This is the method I've used here.

Overrides can serve the same purpose for more precise update, then the override can be removed (IIRC).

PR #5858, opened 2025-10-27
Feature/atomic data replacement during historical market data gathering

inevs:
Changes:
 - Add `replaceAllForSymbol()` method to MarketDataService that  performs delete and insert within a Prisma transaction
 - Remove upfront `deleteMany()` call from `gatherSymbol()` method
 - Add `replaceExistingData` flag to DataGatheringItem interface to distinguish manual refresh from scheduled updates
 - Update data gathering processor to use atomic replace only for manual operations while keeping normal upsert behavior for scheduled updates
 - Remove unused MarketDataService dependency from DataGatheringService

### DIFF
--- a/apps/api/src/services/market-data/market-data.service.ts
+++ b/apps/api/src/services/market-data/market-data.service.ts
@@ -147,8 +147,8 @@ export class MarketDataService {
         let minTime = Infinity;
         let maxTime = -Infinity;
 
-        for (const item of data) {
-          const time = (item.date as Date).getTime();
+        for (const { date } of data) {
+          const time = (date as Date).getTime();
 
           if (time < minTime) {
             minTime = time;


inevs:
this fixes issue #5686

apps/api/src/services/market-data/market-data.service.ts
```diff
@@ -205,4 +205,41 @@ export class MarketDataService {
...
+
+  /**
+   * Atomically replace all market data for a symbol.
+   * Deletes existing data and inserts new data within a single transaction
+   * to prevent data loss if the operation fails.
+   */
+  public async replaceAllForSymbol({
+    dataSource,
+    symbol,
+    data
```
dtslvr: Move up ⬆️

apps/api/src/services/market-data/market-data.service.ts
```diff
@@ -205,4 +205,41 @@ export class MarketDataService {
...
+  }: {
+    dataSource: DataSource;
```
dtslvr: `{ AssetProfileIdentifier & { data: Prisma.MarketDataUpdateInput[] } }`

apps/api/src/services/market-data/market-data.service.ts
```diff
@@ -205,4 +205,41 @@ export class MarketDataService {
...
+    symbol: string;
+    data: Prisma.MarketDataUpdateInput[];
+  }): Promise<void> {
```
dtslvr: Remove `Promise<void>`

apps/api/src/services/queues/data-gathering/data-gathering.processor.ts
```diff
@@ -100,7 +100,7 @@ export class DataGatheringProcessor {
     name: GATHER_HISTORICAL_MARKET_DATA_PROCESS_JOB_NAME
   })
   public async gatherHistoricalMarketData(job: Job<DataGatheringItem>) {
-    const { dataSource, date, symbol } = job.data;
+    const { dataSource, date, symbol, replaceExistingData } = job.data;
```
dtslvr: `{ dataSource, date, replaceExistingData, symbol }`

apps/api/src/services/interfaces/interfaces.ts
```diff
@@ -20,4 +20,5 @@ export interface DataProviderResponse {
 
 export interface DataGatheringItem extends AssetProfileIdentifier {
   date?: Date;
+  replaceExistingData?: boolean;
```
dtslvr: `force?: boolean;` ❓

apps/api/src/services/queues/data-gathering/data-gathering.processor.ts
```diff
@@ -157,7 +157,15 @@ export class DataGatheringProcessor {
         currentDate = addDays(currentDate, 1);
       }
 
-      await this.marketDataService.updateMany({ data });
+      if (replaceExistingData) {
+        await this.marketDataService.replaceAllForSymbol({
+          dataSource,
+          symbol,
+          data
```
dtslvr: Move up ⬆️

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Ensured the locale is available in the settings dialog to customize the rule thresholds of the _X-ray_ page
+- Ensured atomic data replacememt for historical market data fetching
```
dtslvr: Move into `### Changed` ❓

dtslvr (changes requested):
Thanks for your contribution, @inevs! 😎

dtslvr:
Could you please review this PR, @KenTandrian?
We would appreciate your perspective on the implementation and help with testing.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improved the language localization for German (`de`)
+- Ensured atomic data replacememt for historical market data fetching
```
KenTandrian: `replacememt` should be `replacement`?

KenTandrian:
I have tested the code. I think it works really well, now it makes sure the data are fetched successfully before running the atomic transaction of deleting and creating. Thanks for the fix, @inevs!

One small note I have here:
```ts
await prisma.marketData.deleteMany({
  where: {
    dataSource,
    symbol
  }
});
```
Currently it is deleting **all** data but creating new data based on the date range.

Do you think we should only delete data based on the date range, or the current behavior looks good? Any thoughts, @dtslvr?

dtslvr:
> Do you think we should only delete data based on the date range, or the current behavior looks good? Any thoughts, @dtslvr?

Yes, I agree. For fixing a minor issue via `MTD` range, we should not wipe the full history.

How about we use the earliest date from the incoming `data` array as the start date for deletion?

inevs:
So then we first have to retrieve the data in order to get the earliest dat range, don‘t we? 
And is that to be fixed in this issue or shall we create a new one? @dtslvr

KenTandrian:
I believe `replaceAllForSymbol` actually receives a parameter called `data`. If I understand correctly, `data` comes from historical data retrieval, right?

dtslvr:
> I believe `replaceAllForSymbol` actually receives a parameter called `data`. If I understand correctly, `data` comes from historical data retrieval, right?

Exactly. Let’s adapt it in this PR, @inevs.

apps/api/src/services/market-data/market-data.service.ts
```diff
@@ -132,6 +132,64 @@ export class MarketDataService {
...
+        // Find the earliest and latest dates in the incoming data
+        const dates = data.map(({ date }) => {
+          return date as Date;
+        });
+
+        const minDate = new Date(
+          Math.min(
+            ...dates.map((date) => {
+              return date.getTime();
+            })
```
KenTandrian: For efficiency, can we just move this process to `const dates = data.map(...)` (line 148)? Hence we don't need to run `.map()` 3 times (when initializing `dates`, `minDate` and `maxDate`).

The `Array.prototype.map()` function has a time complexity of `O(n)`, hence running it 3 times can increase latency.

@dtslvr any thoughts on this?

inevs: good catch

PR #5912, opened 2025-11-04
Feature/OIDC auth

gmag11:
This implements OIDC auth. Not thoroughly tested yet, this is a first attempt to include OIDC. It is tested with Authentik only.

To configure it refer to `.env.example` placeholders. There is no documentation update yet.

Besides OIDC integration, I've added some usability improvements. 

- Allow disable token login
- Do not show token related items (reset user token) if token is disabled
- Always disallow signup if token login is disabled. No option is shown on admin UI.
- Show OIDC button only (and do not show google button ) if only oidc is configured
- Assign Admin role to first created user on any source. Before only first token created user was assigned to admin role.

The idea with this PR is to decide how the final implementation could be.

Signin dialog would be like this:

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,7 +15,7 @@ ACCESS_TOKEN_SALT=<INSERT_RANDOM_STRING>
 DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${POSTGRES_DB}?connect_timeout=300&sslmode=prefer
 JWT_SECRET_KEY=<INSERT_RANDOM_STRING>
 
-ROOT_URL=https://<your_domain>
+ROOT_URL=https://ghostfolio.domain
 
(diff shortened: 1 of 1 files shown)


Citrus0299:
Thanks for the work, however there is absolutely no reason to request this much user information by default.
You only need to request `openid` scope which is completely anonymous. The combination of `iss` and `sub` ensures stable user identification as required by the spec. There is no technical reason why the other scopes would be required for a privacy focussed app like this.

gmag11:
> Thanks for the work, however there is absolutely no reason to request this much user information by default. You only need to request `openid` scope which is completely anonymous. The combination of `iss` and `sub` ensures stable user identification as required by the spec. There is no technical reason why the other scopes would be required for a privacy focussed app like this.

Hi, thank you for your comment. Finally this PR has been splitted in several ones. This one will not be merged.

I'll take your comments into account for final version. It makes sense.

PR #5914, opened 2025-11-04
Bugfix/assign admin role to first user signing up

gmag11:
This centralizes the logic for automatically assigning the ADMIN role to the first user created in the system, regardless of the authentication method used.


dtslvr (approved):
Thanks for your contribution, @gmag11! 😎
I have made a few minor changes. Ready to merge now.

Issue #5950, opened 2025-11-15
No exchange rate has been found for GBpGBP


s0129:
**Bug Description**

There doesn't seem to be a way to import an exchange rate for GBp to GBP (pounds to pence).
Everything seems to work ok if the "Base Currency" is set to `GBp` but when set to `GBP` I get errors in the logs.
```
No exchange rate has been found for GBpGBP
```
Under Market data I have USDGBP and USDGBp but I can't see any way to import GBpGBP.

**To Reproduce**

1. Set base currency to `GBP`
2. Add a buy transaction for a stock listed on yahoo in GBp such as `PGOO.L`
3. View holdings page for the stock
4. Check logs for error messages

**Expected Behaviour**

Handle converting between GBP and GBp automatically.

**Screenshots**

<img width="1859" height="681" alt="Image" src="https://github.com/user-attachments/assets/dd08c384-3bef-4b4f-a7a2-de0e64f3cfe6" />

**Logs**

```
ERROR [ExchangeRateDataService] No exchange rate has been found for GBpGBP at 2019-08-04. Please complement market data for USDGBp and USDGBP.
```

**Environment**

- Ghostfolio Version 2.217.0
- Self-hosted
- Experimental Features disabled
- Browser Chrome/Edge
- OS win11

inevs:
I've analyzed this issue and identified the root cause:

The Problem:
Ghostfolio treats GBp (pence) and GBP (pounds) as two independent foreign currencies that need exchange rates, when they actually have a fixed mathematical relationship (1 GBP = 100 GBp).

What's happening:
1. When you have a stock priced in GBp and your base currency is GBP, the system tries to find an exchange rate for "GBpGBP"
2. When that doesn't exist, it attempts to convert indirectly via USD: GBp → USD → GBP
3. This fails when historical USD exchange rate data is missing for the transaction date, resulting in the error message

The Missing Logic:
The getExchangeRates() function in ExchangeRateDataService has special handling for identical currencies (returns factor of 1), but has no special handling for derived currencies like GBp/GBP, ILA/ILS, or ZAc/ZAR.

Proposed Solution:
Add logic to detect when converting between a derived currency and its root currency, and apply the fixed mathematical factor directly (0.01 for GBp→GBP, 100 for GBP→GBp) instead of trying to fetch exchange rates.

This affects not just GBp/GBP but also Israeli Agorot/Shekel and South African Cents/Rand conversions.

@dtslvr what do you think?

[...]

PR #5961, opened 2025-11-18
Bugfix/exchange rate calculation when converting derived currencies

inevs:
…s root currency, and apply the fixed mathematical factor directly (0.01 for GBp→GBP, 100 for GBP→GBp) instead of trying to fetch exchange rates.

this fixes issue #5950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed an issue with the exchange rate calculation when converting between derived currencies and their root currencies
+
 ## 2.219.0 - 2025-11-23
 
 ### Added
@@ -25,7 +31,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed an issue with the edit of future activities (drafts)
-- Fixed an issue with the exchange rate calculation when converting between derived currencies and their root currencies
 
 ## 2.218.0 - 2025-11-20
 


inevs:
@dtslvr I applied a fix for this issue

apps/api/src/services/exchange-rate-data/exchange-rate-data.service.ts
```diff
@@ -271,56 +271,65 @@ export class ExchangeRateDataService {
...
+        aFromCurrency,
+        aToCurrency
+      );
 
-      if (marketData?.marketPrice) {
-        factor = marketData?.marketPrice;
+      if (derivedCurrencyFactor !== null) {
+        factor = derivedCurrencyFactor;
       } else {
-        // Calculate indirectly via base currency
```
dtslvr, line 287: I don't understand what's the diff here

apps/api/src/services/exchange-rate-data/exchange-rate-data.service.ts
```diff
@@ -469,6 +496,35 @@ export class ExchangeRateDataService {
     return factors;
   }
 
+  private getDerivedCurrencyFactor(
```
dtslvr: Wouldn't it be more efficient, if a `derivedCurrenciesFactorMap` in both directions is created upon initialization? 🤔

apps/api/src/services/exchange-rate-data/exchange-rate-data.service.ts
inevs: The existing market data fetching logic now runs inside an else block, meaning it only executes when we're NOT dealing with a derived currency conversion

dtslvr (commented):
I have simplified the implementation. Could you please review it to confirm that the update achieves the intended improvement, @inevs?

apps/api/src/services/exchange-rate-data/exchange-rate-data.service.ts
```diff
@@ -357,7 +368,22 @@ export class ExchangeRateDataService {
...
+
+    if (derivedCurrencyFactor) {
+      for (const date of dates) {
+        factors[format(date, DATE_FORMAT)] = derivedCurrencyFactor;
+      }
+
+      return factors;
+    }
+
+    {
```
dtslvr: This block looks weird and could probably be removed, but it messes up the diff 😵‍💫

inevs:
@dtslvr As the result of my review I have to say, the code is more efficient with simpler syntax and it works the same. Thank you for the refactoring.

dtslvr (approved):
Thanks again for your contribution, @inevs! 🙌🏻
Ready to merge now.

PR #5987, opened 2025-11-24
Remove min-width from currency and status containers

bennypowers:
Currency Codes wrap to the next line on mobile widths, disrupting the visual flow of the homepage and causing a horizontal overflow. This change removes the min-width, which causes the flex overflow.

Before:

Screenshot from 2025-11-24 13-23-58

After:

Screenshot from 2025-11-24 13-25-45

### DIFF
--- a/apps/client/src/app/components/portfolio-performance/portfolio-performance.component.scss
+++ b/apps/client/src/app/components/portfolio-performance/portfolio-performance.component.scss
@@ -4,7 +4,6 @@
   .currency-container,
   .status-container {
     flex: 1;
-    min-width: 2.5rem;
   }
 
   .status-container {


dtslvr:
Thanks for creating a PR, @bennypowers.

The `min-width` is there to ensure the number remains centered.

The question is why your setup isn't recognizing the `device` as `mobile`, which seems to prevent the automatic `precision` adaption.

bennypowers:
I experience this both in firefox and chrome, on android and desktop linux

dtslvr:
Could you please debug and find out exactly why `deviceType` is not `mobile` [here](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/home-overview/home-overview.component.ts#L144)?

bennypowers:
There aren't any source maps, and the containerized version doesn't appear to have options to enable them :/

I don't have time to pursue this further, although the problem persists. I'll close it for now.

PR #6010, opened 2025-11-30
Task/update VS Code extension of Prettier

dtslvr:


KenTandrian:
@dtslvr I believe we will also need to update `.vscode/settings.json` so that `editor.defaultFormatter` is set to the new `prettier.prettier-vscode` extension.

dtslvr:
Thanks for the hint, @KenTandrian.

I have updated the default formatter, however I see a warning now: `Value is not accepted.` 🤔

KenTandrian:
I got the same problem as well, but once I installed the new Prettier extension, the warning disappeared.

dtslvr:
I switched back to the legacy extension. That might explain it, @KenTandrian!

Issue #6008, opened 2025-11-29
Portfolio proportion chart component shows incorrect percentage values


dtslvr:
**Bug Description**

The [portfolio proportion chart component](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/portfolio-proportion-chart) shows incorrect values when data is provided in percentage format.

**To Reproduce**

1. Start _Storybook_ locally via `npm run start:storybook`
2. Open the story URL: http://localhost:4400/?path=/story/portfolio-proportion-chart--in-percentage
3. Observe the rendered value for _United States_: `88.89%`

**Expected Behavior**

Based on the _Storybook_ [data](https://github.com/ghostfolio/ghostfolio/blob/main/libs/ui/src/lib/portfolio-proportion-chart/portfolio-proportion-chart.component.stories.ts#L44), the value for _United States_ should be rendered as `65.15%`

**Screenshots**

<img width="312" height="315" alt="Image" src="https://github.com/user-attachments/assets/767aad01-dce6-4969-b16c-4b01c6ae6d31" />

**Environment**

- Ghostfolio Version `2.220.0`

Raj-G07:
- Our dataset represents 73.29% approx of the whole, not a full 100%.

- If we calculate 65/73.29 it is close to 88.89%

DavidReque:
Hi @dtslvr is anyone else working on this issue? If not, I would like to do it

dtslvr:
@DavidReque if the total of the dataset is less than 100%, the remaining percentage should be allocated to unknown (see `UNKNOWN_KEY`) to ensure the total distribution is correctly represented.

PR #6013, opened 2025-11-30
fix(ui): correct percentage-based story data for portfolio proportion chart (#6008)

abdullahhafizh:
This pull request addresses issue **#6008**, where the *PortfolioProportionChart* incorrectly rendered percentage values in the Storybook `InPercentage` story.

### Problem

The component displayed erroneous proportions because the Storybook sample data did not represent normalized percentage values, resulting in a mismatched rendering (e.g., United States shown as `88.89%` instead of the expected `65.15%`).

### Fix

* Updated `InPercentage` Storybook data with normalized values so the percentage mode displays accurate proportions.
* Ensured that all country values sum to one hundred percent when interpreted in percentage mode.
* No changes to component behavior, only correction of story sample data.

### Verification

1. Run Storybook via `npm run start:storybook`.
2. Open: `/?path=/story/portfolio-proportion-chart--in-percentage`.
3. Confirm that United States now renders as `65.15%` and other values match the updated distribution.


dtslvr:
Thank you for your contribution, @abdullahhafizh.
We are closing this PR because the suggested approach is not the direction we intend to take.

PR #6015, opened 2025-12-01
Task/testimprovement authservice

inevs:
I added some tests to AuthService in order to improve test coverage


inevs:
@dtslvr will you review it?

apps/api/src/app/auth/auth.service.spec.ts
dtslvr, line 1: I'm unsure of the actual benefit here, as these tests appear to rely mostly on mocked logic.

DEVELOPMENT.md
```diff
@@ -52,7 +52,28 @@ npm run database:push
...
+npm test              # All tests
+```
+
+### Writing Tests
+
+**Always add tests for:**
+
+- New services and controllers (required)
+- Bug fixes (recommended)
+- focus on business logic
```
dtslvr, line 67: **F**ocus...

DEVELOPMENT.md
```diff
@@ -52,7 +52,28 @@ npm run database:push
...
+
+```bash
+npm test              # All tests
+```
+
+### Writing Tests
+
+**Always add tests for:**
+
+- New services and controllers (required)
```
dtslvr, line 65: I appreciate the goal of improving quality. However, I have reservations about strictly enforcing this rule.

It could unnecessarily slow down development and it conflicts with the current testing state of the existing codebase.

I suggest we focus practically on critical functional testing (calculations and imports) first to establish the standard, rather than introducing a strict requirement.

apps/api/src/app/auth/auth.service.spec.ts
inevs: This is a common concern about unit tests with heavy mocking. There are different testing schools or styles: 
- London (Freeman et al) school relies on testing behavior by using mocks (and of cause we need to have tests for mocked services as well=
- Detroit school (Kent Beck et al.) will test the implementation without mocks. 
As you see, I am a fan of the first one.

Let me explain the real value:

What These Tests Actually Validate
Despite the mocking, the auth.service.spec.ts tests provide significant value:

1. Business Logic & Control Flow
The tests verify the actual logic in AuthService:
  - Does it call the right methods in the right order?
  - Does it pass the correct parameters?
  - Does it handle the response correctly?

i.e. the test validates: "When user doesn't exist AND signup is disabled, does it throw an error?" This is real business logic, not mock behavior.

2. Integration Points & Contracts
  Tests document and verify:
  - What dependencies does this service need?
  - What methods does it call on those dependencies?
  - What data format does it expect back?

If you change userService.users() to require different parameters, the test will fail - alerting you to update all callers.

3. Error Handling
Tests verify the service handles failures correctly: 
it('should reject when users query throws error', async () => {
    userService.users.mockRejectedValue(new Error('Database error'));

    await expect(service.validateAnonymousLogin(accessToken))
      .rejects.toBeUndefined();
  });
This tests: "When the database fails, does AuthService handle it gracefully?"

4. Prevents Regressions
Someone might refactor and accidentally:
  - Skip the signup check
  - Forget to hash the access token
  - Pass wrong parameters to JWT
  - Change the order of operations
The tests catch these immediately.

I am fine if we use the other testing style as well.

DEVELOPMENT.md
inevs: you are right. How would you phrase it?

PR #6026, opened 2025-12-01
Task/upgrade envalid to version 8.1.1

### LINKED ISSUE #5712: Support serving over the unspecified IPv6 address
Ghostfolio can't serve over the unspecified IPv6 address `::`, which is equivalent to running on `0.0.0.0` (current default).

### Default state
Test for default host and port:
```sh
$ curl -v "[::]:3333"
*   Trying [::]:3333...
* connect to :: port 3333 from ::1 port 33490 failed: Connection refused
* Failed to connect to :: port 3333 after 0 ms: Could not connect to server
* closing connection #0
curl: (7) Failed to connect to :: port 3333 after 0 ms: Could not connect to server
```

### Changing `HOST` to `::`

This fails to parse due to `envalid.host` not supporting this address in their regexp.
Submitting a fix upstream is definitely an option, although I'm not sure it they're open to it yet.

### Changing `HOST` to `""` (empty string)

This also triggers `envalid.host` validation and does not pass.

### Result of using `::` or `""`

If the constant `DEFAULT_PORT` is changed (or `main.ts` is adjusted directly) using both an empty string and `::` still result in Ghostfolio being available over IPv4. This behaviour is expected, as per node.js docs[^1].

[^1]:https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback

### Possible solutions:

I have identified a couple:
1. Remove `DEFAULT_HOST` - only use non-standard host if clients provide `HOST` explicitly.
2. Skip providing hostname to the HTTP server if `configService.get<string>('HOST') === DEFAULT_HOST`. 
3. Bypass `host` validation for special case where `HOST` env is `""` or `::`.
4. Drop the notion `DEFAULT_HOST` - use runtime defaults when possible. In cases where hinting an address is required, `localhost` can be used or the following accurate lookup:
```ts
    const { address, port, family } = app
      .getHttpServer()
      .address() as AddressInfo;
    let host = address;
    if (family === 'IPv6') {
      host = `[${address}]`;
    }
    Logger.log(`Listening at http://${host}:${port}`);
```

slagiewka:
This will allow the HTTP server to listen on `::`.

Closes #5712

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-#### Chagned
+#### Changed
 
 - Upgraded `envalid` from version `8.1.0` to `8.1.1`
 


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+#### Chagned
```
dtslvr: Changed

PR #6027, opened 2025-12-02
Task/upgrade to prisma 7

### LINKED ISSUE #5969: Upgrade to Prisma 7
The _Prisma_ dependencies need to be upgraded to the latest version (`7.0`).

Upgrade guide from to _Prisma ORM 7_: https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-7

KenTandrian:
Hi @dtslvr, this PR resolves #5969. Please take a look :)

### Changes
* Bumped the `prisma` packages from 6.19.0 to 7.0.1.
* Move database URL to Prisma config file.
* Implement Prisma Postgres adapter.

### DIFF
--- a/libs/common/src/test-setup.ts
+++ b/libs/common/src/test-setup.ts
@@ -1,3 +1,3 @@
-import { TextEncoder, TextDecoder } from 'node:util';
+import { TextDecoder, TextEncoder } from 'node:util';
 
 Object.assign(global, { TextDecoder, TextEncoder });


KenTandrian:
@dtslvr I am not able to reproduce the issue. The server runs successfully on my end. Are you using SSL for the database connection?

Seems like the Prisma client tries to send a TLS session request but the database is not SSL enabled.

dtslvr:
Interesting. My `DATABASE_URL` contains `sslmode=prefer` (see [here](https://github.com/ghostfolio/ghostfolio/blob/main/.env.dev#L15)). Is it not supported anymore?

KenTandrian:
I don't think that's the case. I am using the same configuration (`sslmode=prefer`)

KenTandrian:
I think I am able to reproduce the issue now. It only happens when connecting to a local database which is not SSL-enabled.
Might be related to this issue: https://github.com/prisma/prisma/issues/27611.

dtslvr:
So, would the migration to Prisma 7 be considered a breaking change regarding `sslmode=prefer`?

KenTandrian:
Yes true, we will need to wait for that issue to be fixed. For the time being:
* If the database doesn't have SSL enabled, remove `sslmode=prefer`.
* If the database has SSL enabled, set `sslmode=prefer` or `sslmode=require`.

KenTandrian:
@dtslvr I am thinking of implementing an async SSL detection fix to resolve the [Prisma 7 SSL bug](https://github.com/prisma/prisma/issues/27611).

By adding a function that performs a lightweight TCP handshake during app startup, we can automatically fallback to `sslmode=disable` when SSL is not supported. This restores the robust "zero-config" SSL behavior from Prisma 6 for all self-hosted environments.

What do you think?

dtslvr:
To avoid adding fallback logic, I would prefer to focus on Ghostfolio 3 (including _Prisma_ 7). This gives us the opportunity to introduce a clean breaking change for the database connection. What do you think, @KenTandrian?

KenTandrian:
@dtslvr I see, introducing breaking changes in a new major version actually makes sense.

That being said, since `ssl=prefer` is not working as intended, are we going to force all database connections to be SSL-enabled? Or we will introduce a new environment variable (e.g., `POSTGRES_SSL=true`) so that the user can choose to enable SSL or not by themselves?

dtslvr:
My understanding is that we would update the default in:
https://github.com/ghostfolio/ghostfolio/blob/main/.env.example#L15

Then, we will document that existing self-hosters need to adapt their `DATABASE_URL` environment variable accordingly as part of the migration.

dtslvr:
Could you please retarget this PR to the [next](https://github.com/ghostfolio/ghostfolio/tree/next) branch, @KenTandrian? Also, please add an entry to the `CHANGELOG.md` under `### Todo` with the `DATABASE_URL` instruction.

PR #6075, opened 2025-12-16
Feature/OIDC user link

### LINKED ISSUE #6063: Link user account with identity provider
The goal of this issue is to implement an account linking feature in the user settings that allows an existing token-authenticated user to associate their profile with an _OpenID Connect_ (OIDC) provider. Integration with the _Google_ provider is currently out of scope, but the wording should be reusable.

gmag11:
This pull request implements "link mode" for OIDC authentication, allowing users with token-based authentication to link their account to an OIDC provider.


apps/api/src/app/auth/auth.controller.ts
```diff
@@ -111,26 +113,67 @@ export class AuthController {
...
   @Get('oidc/callback')
   @UseGuards(AuthGuard('oidc'))
   @Version(VERSION_NEUTRAL)
-  public oidcLoginCallback(@Req() request: Request, @Res() response: Response) {
-    const jwt: string = (request.user as any).jwt;
+  public async oidcLoginCallback(
+    @Req() request: Request,
+    @Res() response: Response
+  ) {
+    const result = request.user as OidcValidationResult;
```
dtslvr: `const { linkState, thirdPartyId }  = ...`

apps/api/src/app/auth/auth.service.ts
```diff
@@ -40,35 +45,104 @@ export class AuthService {
...
 
-        // Create new user if not found
-        user = await this.userService.createUser({
-          data: {
-            provider,
-            thirdPartyId
-          }
-        });
+      if (!isUserSignupEnabled) {
+        throw new Error('Sign up forbidden');
```
dtslvr: Could you throw a native forbidden error?

apps/api/src/app/auth/auth.service.ts
```diff
@@ -40,35 +45,104 @@ export class AuthService {
...
+   *
+   * @param userId - The ID of the user to link
+   * @param thirdPartyId - The OIDC subject identifier
+   * @returns JWT token for the linked user
+   * @throws ConflictException if thirdPartyId is already linked to another user
+   * @throws Error if user not found or has invalid provider
+   */
+  public async linkOidcToUser(
+    userId: string,
+    thirdPartyId: string
```
dtslvr: Move up ⬆️

dtslvr: Use a params object

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -111,26 +113,67 @@ export class AuthController {
...
+        const errorMessage =
+          error instanceof Error ? error.message : 'Unknown error';
+        Logger.error(
+          `OIDC callback: Link failed - ${errorMessage}`,
+          'AuthController'
+        );
+
+        // Determine error type for frontend
+        let errorCode = 'unknown';
+        if (errorMessage.includes('already linked')) {
```
dtslvr: Could you check for the error type instead of the message?

apps/api/src/app/auth/auth.service.ts
```diff
@@ -40,35 +45,104 @@ export class AuthService {
...
+      throw new ConflictException(
+        'This OIDC account is already linked to another user'
+      );
+    }
+
+    // Get the current user
+    const user = await this.userService.user({ id: userId });
+
+    if (!user) {
+      throw new Error('User not found');
```
dtslvr: Throw not found exception

apps/api/src/app/auth/auth.service.ts
```diff
@@ -40,35 +45,104 @@ export class AuthService {
...
+    }
+
+    if (user.provider !== 'ANONYMOUS') {
+      throw new Error('Only users with token authentication can link OIDC');
+    }
+
+    // Update user with thirdPartyId and switch provider to OIDC
+    await this.userService.updateUser({
+      where: { id: userId },
+      data: { thirdPartyId, provider: 'OIDC' }
```
dtslvr: Move up ⬆️

apps/api/src/app/auth/oidc-state.store.ts
```diff
@@ -1,40 +1,109 @@
+import { Logger } from '@nestjs/common';
+import * as jwt from 'jsonwebtoken';
 import ms from 'ms';
 
+export interface OidcLinkState {
```
dtslvr: Could you extract to `./interfaces/interfaces.ts`?

apps/api/src/app/auth/oidc-state.store.ts
```diff
@@ -1,40 +1,109 @@
...
  * Custom state store for OIDC authentication that doesn't rely on express-session.
  * This store manages OAuth2 state parameters in memory with automatic cleanup.
+ * Supports link mode for linking existing token-authenticated users to OIDC.
  */
 export class OidcStateStore {
   private readonly STATE_EXPIRY_MS = ms('10 minutes');
 
+  private pendingLinkState?: OidcLinkState;
+
+  private jwtSecret?: string;
```
dtslvr: Move up ⬆️

apps/api/src/app/auth/oidc-state.store.ts
```diff
@@ -1,40 +1,109 @@
+import { Logger } from '@nestjs/common';
+import * as jwt from 'jsonwebtoken';
```
dtslvr: Could you use `JwtService`? `jsonwebtoken` is not a direct dependency (yet).

apps/api/src/app/auth/oidc-state.store.ts
```diff
@@ -111,4 +187,21 @@ export class OidcStateStore {
       Date.now().toString(36)
     );
   }
+
+  /**
+   * Set link state for an existing or upcoming state entry.
+   * This allows the controller to attach user information before the OIDC flow starts.
+   */
+  public setLinkStateForNextStore(linkState: OidcLinkState) {
```
dtslvr: Move up ⬆️

apps/api/src/app/auth/oidc-state.store.ts
```diff
@@ -111,4 +187,21 @@ export class OidcStateStore {
...
+    this.pendingLinkState = linkState;
+  }
+
+  /**
+   * Get and clear pending link state (used internally by store)
+   */
+  public getPendingLinkState(): OidcLinkState | undefined {
```
dtslvr: Move up ⬆️

apps/api/src/app/auth/oidc.strategy.ts
```diff
@@ -11,25 +11,44 @@ import {
   OidcParams,
   OidcProfile
 } from './interfaces/interfaces';
-import { OidcStateStore } from './oidc-state.store';
+import { OidcLinkState, OidcStateStore } from './oidc-state.store';
+
+export interface OidcValidationResult {
```
dtslvr: Move to `./interfaces/interfaces.ts`

apps/api/src/app/auth/oidc.strategy.ts
```diff
@@ -11,25 +11,44 @@ import {
...
+  jwt?: string;
+  linkState?: OidcLinkState;
+  thirdPartyId: string;
+}
+
+export interface OidcStrategyOptions extends StrategyOptions {
```
dtslvr: Move to `./interfaces/interfaces.ts`

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -110,17 +115,42 @@ export class GfUserAccountSettingsComponent implements OnDestroy, OnInit {
...
     public webAuthnService: WebAuthnService
   ) {
-    const { baseCurrency, currencies } = this.dataService.fetchInfo();
+    const { baseCurrency, currencies, globalPermissions } =
+      this.dataService.fetchInfo();
 
     this.baseCurrency = baseCurrency;
     this.currencies = currencies;
 
+    // Check global permissions for auth methods
```
dtslvr: Remove comment

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -110,17 +115,42 @@ export class GfUserAccountSettingsComponent implements OnDestroy, OnInit {
...
+    this.hasPermissionForAuthOidc = hasPermission(
+      globalPermissions,
+      permissions.enableAuthOidc
+    );
+    this.hasPermissionForAuthToken = hasPermission(
```
dtslvr, line 131: Add line break

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -110,17 +115,42 @@ export class GfUserAccountSettingsComponent implements OnDestroy, OnInit {
...
 
+          // Check if user can link OIDC
+          // Both OIDC and Token auth must be enabled to show linking feature
+          // Only show for users with token auth (provider ANONYMOUS)
+          this.hasOidcLinked =
+            this.hasPermissionForAuthOidc &&
+            this.hasPermissionForAuthToken &&
+            this.user.provider === 'ANONYMOUS' &&
+            !!this.user.thirdPartyId;
+          this.canLinkOidc =
```
dtslvr, line 148: Add line break

apps/client/src/app/components/user-account-settings/user-account-settings.component.ts
```diff
@@ -143,11 +173,57 @@ export class GfUserAccountSettingsComponent implements OnDestroy, OnInit {
...
+      });
+  }
+
+  public getAuthProviderDisplayName(): string {
+    switch (this.user?.provider) {
+      case 'ANONYMOUS':
+        return 'Security Token';
+      case 'GOOGLE':
+        return 'Google';
+      case 'INTERNET_IDENTITY':
```
dtslvr: Remove

dtslvr (changes requested):
Thanks for your contribution, @gmag11. The implementation seems complete, but I'm questioning if the complexity introduced here is proportional to the added value.

gmag11:
> Thanks for your contribution, @gmag11. The implementation seems complete, but I'm questioning if the complexity introduced here is proportional to the added value.

In my first implementation I used a separate flow to start the linkage. It was simpler but it required an additional endpoint in the api. In order to reuse oidc code, with no changes in the endpoints, I had to add a parameter to request and modify OIDC flow to keep this parameter along the flow.

Authentication service did not send that parameter back (at least I did not find a way to do it with Authentik), so I had to store it temporarily so that it is remembered and no new user is created when getting the OIDC callback. This was not easy and it is the main reason of most complexity.

gmag11:
> Thanks for your contribution, @gmag11. The implementation seems complete, but I'm questioning if the complexity introduced here is proportional to the added value.

Hi. I'm analyzing if this feature may be implemented with a simpler flow. Wait a couple days. I will come back with my analysis and a new implementation.

gmag11:
Finally it was shorter. I've redesigned the flow to not use user intent as the marker to link the new openID to the user. Now if callback is received while the user is logged with token in then the id is linked. If now a new user is created as it happened before.

This implementation is simpler, but UI integration is the same.

I've left some comments in the code to help reading. Let me know if you prefer removing it after review.

flowcool:
Very good ! Any news for the merge in main ?

flowcool:
## Community review + use case

I'm a self-hosted Ghostfolio user who needs this feature: I have an existing token-based account with years of portfolio data and I need to migrate to OIDC (Authelia) without losing that data. This PR is the only path forward — today I'm stuck either re-entering everything or staying on token auth forever.

I went through the diff in detail. Here are my findings.

---

### What works well

**Architecture** — passing the `linkToken` through the OAuth state (via `OidcStateStore`) is clever. It avoids adding a new endpoint, keeps the flow entirely within the existing OIDC callback path, and the logic split in `OidcStrategy.validate()` between "link mode" and "normal login" is clean and easy to follow.

**Error handling** — `linkOidcToUser()` covers the expected failure cases correctly: already linked to another user (`ConflictException`), user not found (`NotFoundException`), wrong provider (`ConflictException`). Idempotency is handled: re-linking to the same account returns success without throwing.

**Security validation** — the `linkToken` is verified with `jwtService.verify()` before extracting `userId`. No blind trust of the token content.

**State store fix** — `OidcStateStore.verify()` callback signature was corrected from `(err, appState, ctx)` to `(err, ctx, state)`, aligning with `passport-openidconnect`. This is a real bug fix bundled in.

**UI** — the confirmation dialog before linking and the snackbar feedback (success/error with specific messages per error code) are a good user experience.

---

### One concern worth discussing

**`linkToken` passed as a GET query parameter**

```ts
// user-account-settings.component.ts
input.name = 'linkToken';
input.value = token; // full JWT in GET param → visible in OIDC provider access logs
```

The user's JWT ends up in the URL when the browser navigates to `../api/auth/oidc?linkToken=...`. This means it's:
- Visible in the OIDC provider's (e.g. Authelia) access logs if it logs full URLs
- Potentially in browser history

Mitigating factors: the JWT has a limited lifespan, the OIDC provider is on the same infrastructure, and this is a one-time explicit user action with a confirmation dialog. So the practical risk is low for self-hosted setups.

If the maintainer wants to address this, a short-lived opaque nonce stored server-side (mapped to the userId) would avoid exposing the JWT in the URL. But that adds exactly the kind of complexity @dtslvr wanted to avoid. I'd consider this a known trade-off rather than a blocker, especially given the narrow self-hosted context.

---

### Summary

The refactored implementation is clean, proportionate to the feature, and the residual complexity is inherent to the problem (threading state through an OAuth round-trip without a new endpoint). I'd call this merge-ready.

Is there anything specific blocking a second review from @dtslvr?

hgrgic:
Any timeline to merge and release this one? It would super useful for me to link newly added OIDC provider with PocketID with my token based account.

flowcool:
@dtslvr ? Apology Thomas, but what's going on ? It would be also super usefull for your paid subscribers ?

PR #6077, opened 2025-12-18
Bugfix/localize date formatting in FIRE calculator

### LINKED ISSUE #6070: Localize month names in FIRE calculator
The goal of this issue is to localize the retirement date in the [FIRE calculator component](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/fire-calculator) based on the provided `locale` (see `@Input()`).

Image

DavidReque:
hi @dtslvr, this PR resolves #6070

### Problem
The retirement date in the FIRE calculator displays month names in English regardless of the user's locale setting (e.g., "December 2035" instead of "Dezember 2035" for German users).

### Cause
The Angular date pipe was used without the locale parameter, defaulting to English:
```html
{{ calculatorForm.get('retirementDate')?.value | date: 'MMMM yyyy' }}
```

### Solution
Pass the component's locale input to the date pipe to properly localize month names:
```html
{{ calculatorForm.get('retirementDate')?.value | date: 'MMMM yyyy' : undefined : locale }}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the deprecated _Angular CLI_ decorator (`decorate-angular-cli.js`)
 - Refreshed the cryptocurrencies list
 
+### Fixed
+
+- Localized date formatting across the _FIRE_ section
+
 ## 2.223.0 - 2025-12-14
 
 ### Added


DavidReque:
> Thanks for your contribution, @DavidReque! 😎
> 
> To avoid the boilerplate of passing the locale manually in each pipe, could we handle this at the app level instead?

Updated `main.ts` to use `$localize.locale` (automatically set by Angular during i18n builds) instead of the hardcoded value. This ensures all Angular pipes use the correct locale at the application level, eliminating the need to pass `locale` manually in each pipe.

PR #6089, opened 2025-12-22
Bugfix: Localize month names in FIRE calculator

DavidReque:
Hi @dtslvr this PR fixes issue #6070 by localizing the retirement date display in the FIRE calculator component without causing regressions in Angular Material tables.

### Background
A previous fix (#6077) was reverted because it modified `LOCALE_ID `globally, which broke Angular Material tables. There was also a discrepancy between how dev and production builds handled locales.

### What Changed
Instead of changing global locale settings, I'm using date-fns directly in the FIRE calculator component:
Before:
`{{ calculatorForm.get('retirementDate')?.value | date: 'MMMM yyyy' }}`

After:
```
// Added method to component
public getFormattedRetirementDate(): string {
  const retirementDate = this.calculatorForm.get('retirementDate')?.value;
  if (!retirementDate) return '';
  
  const dateFnsLocale = getDateFnsLocale(this.locale);
  return format(retirementDate, 'MMMM yyyy', { locale: dateFnsLocale });
}
```

`{{ getFormattedRetirementDate() }}`

### Why This Works
- Only affects the FIRE calculator component
- Doesn't touch `main.ts` or global `LOCALE_ID`
- Angular Material tables keep working as before
- Uses the `locale `that's already passed as an `@Input()` to the component
- `date-fns` behaves the same in dev and production

### DIFF
--- a/libs/ui/src/lib/fire-calculator/fire-calculator.component.html
+++ b/libs/ui/src/lib/fire-calculator/fire-calculator.component.html
@@ -31,9 +31,7 @@
         <mat-form-field appearance="outline" class="w-100">
           <mat-label i18n>Retirement Date</mat-label>
           <div>
-            {{
-              calculatorForm.get('retirementDate')?.value | date: 'MMMM yyyy'
-            }}
+            {{ getFormattedRetirementDate() }}
           </div>
           <input
             class="d-none"
--- a/libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
+++ b/libs/ui/src/lib/fire-calculator/fire-calculator.component.ts
@@ -3,7 +3,7 @@ import {
   transformTickToAbbreviation
 } from '@ghostfolio/common/chart-helper';
 import { primaryColorRgb } from '@ghostfolio/common/config';
-import { getLocale } from '@ghostfolio/common/helper';
+import { getDateFnsLocale, getLocale } from '@ghostfolio/common/helper';
 import { FireCalculationCompleteEvent } from '@ghostfolio/common/interfaces';
 import { ColorScheme } from '@ghostfolio/common/types';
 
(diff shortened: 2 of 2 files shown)


DavidReque:
> Thanks for providing this fix, @DavidReque. While I appreciate avoiding the previous regressions, I’m not entirely convinced by this approach.
> 
> Manual formatting in the component is easy to forget. I think we should aim for a solution that uses the standard pipe.

Thanks for the feedback, @dtslvr! 

I understand the concern about manual formatting. However, I'm not sure how to use the standard `date` pipe without modifying `LOCALE_ID` globally, which caused the regressions in #6077.

**The problem:**
- Angular's `date` pipe depends on global `LOCALE_ID`
- Changing `LOCALE_ID` breaks Angular Material tables
- We need localization only in this component

**Possible alternatives:**

1. **Custom pipe with locale parameter:**
```   
{{ retirementDate | localizedDate: 'MMMM yyyy' : locale }}
```
More reusable and still pipe-based.
Component-level `LOCALE_ID` provider:
```
   providers: [{ provide: LOCALE_ID, useValue: this.locale }]
```
Not sure if this avoids the global issue though.
Question: How would you approach using the standard pipe while avoiding the global `LOCALE_ID` problem? I'm happy to implement whatever solution you prefer

omkarg01:
> Thanks for providing this fix, @DavidReque. While I appreciate avoiding the previous regressions, I’m not entirely convinced by this approach.
> 
> Manual formatting in the component is easy to forget. I think we should aim for a solution that uses the standard pipe.

Given the current setup with MAT_DATE_LOCALE + a custom DateAdapter driven by LanguageService, introducing LOCALE_ID would reintroduce global side effects and risks we’ve already seen. Formatting in the component is safe but easy to forget. 

A custom pipe using date-fns with an explicit locale seems like the only reliable and maintainable long-term solution.

PR #6090, opened 2025-12-22
Bugfix/fix asset profile deletion and allow editing asset profile identifiers with MANUAL data source

edualm:
## Summary
This PR fixes two UI bugs in the asset profile dialog of the admin control panel:

1. **Fix delete button always being disabled**: The delete permission check was missing the `watchedByCount` parameter, causing the delete button to be incorrectly disabled.

2. **Allow editing MANUAL symbol identifiers**: Removes an UI restriction that prevented editing symbol identifiers for MANUAL data sources. It seems this is arbitrary, as doing an API call for this works just fine (and after the change, it started working too).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Added the missing currency suffix to the cash balance field in the create or update account dialog
 - Fixed the delete button in the asset profile details dialog of the admin control panel by providing the missing `watchedByCount` parameter
 
 ## 2.224.2 - 2025-12-20
--- a/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
+++ b/apps/client/src/app/pages/accounts/create-or-update-account-dialog/create-or-update-account-dialog.html
@@ -39,7 +39,7 @@
           (keydown.enter)="$event.stopPropagation()"
         />
         <span class="ml-2" matTextSuffix>{{
-          accountForm.get('currency')?.value?.value
+          accountForm.get('currency')?.value
         }}</span>
       </mat-form-field>
     </div>


apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
```diff
@@ -261,10 +261,7 @@ export class GfAssetProfileDialogComponent implements OnDestroy, OnInit {
   }
 
   public get canEditAssetProfileIdentifier() {
```
dtslvr: Do you have any idea why this check is still necessary?

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
```diff
@@ -71,9 +71,10 @@ <h1 class="flex-grow-1 m-0" mat-dialog-title>
         type="button"
         [disabled]="
           !adminMarketDataService.hasPermissionToDeleteAssetProfile({
-            activitiesCount: assetProfile?.activitiesCount,
+            activitiesCount: assetProfile?.activitiesCount ?? 0,
```
dtslvr: `?? 0` is not needed

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
```diff
@@ -71,9 +71,10 @@ <h1 class="flex-grow-1 m-0" mat-dialog-title>
...
             isBenchmark: isBenchmark,
-            symbol: data.symbol
+            symbol: data.symbol,
+            watchedByCount: assetProfile?.watchedByCount ?? 0
```
dtslvr: `?? 0` is not needed

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.component.ts
edualm: I'd say yes, since it prevents editing things like currencies which don't have an asset class.

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
edualm: Right, fixed!

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
edualm: Right, fixed!

edualm:
Done @dtslvr, also added the changelog entries.

Apologies about not following the "one change per PR" rule, I'll keep that in mind for future contributions!

dtslvr (approved):
Thanks again and congrats on your first contribution, @edualm! 🙌🏻

PR #6103, opened 2025-12-28
Bugfix/time in market for impersonation mode

### LINKED ISSUE #6052: Time in Market does not adapt in impersonation mode
**To Reproduce**

1. Activate _Experimental Features_ in _My Ghostfolio_
2. Impersonate a user account you have access
3. _Time in Market_ in the _Portfolio Summary_ shows my value

**Expected behavior**

_Time in Market_ should be the value of the impersonated user account

**Environment**

- Ghostfolio Version `2.222.0`
- _Experimental Features_ enabled

**Additional Context**

Integrate `dateOfFirstActivity` into `PortfolioSummary`.

VahantSharma:
When impersonating a user, the "Time in Market" value was still showing my own data instead of the impersonated user's.

The issue was that the component pulled `dateOfFirstActivity` from the user endpoint, which always returns the logged-in user regardless of impersonation. 

Following the suggestion in the issue, I moved `dateOfFirstActivity` into the portfolio summary response instead — that endpoint already respects impersonation, so it was a straightforward fix.

Resolves #6052

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Added the missing currency suffix to the cash balance field in the create or update account dialog
+- Fixed the time in market display of the portfolio summary tab on the home page for the impersonation mode
 - Fixed the delete button in the asset profile details dialog of the admin control panel by providing the missing `watchedByCount` parameter
 
 ## 2.224.2 - 2025-12-20


dtslvr:
Could you please comment in the issue to get assigned, @VahantSharma?

libs/common/src/lib/interfaces/portfolio-summary.interface.ts
```diff
@@ -8,6 +8,7 @@ export interface PortfolioSummary extends PortfolioPerformance {
   cash: number;
   committedFunds: number;
   dividendInBaseCurrency: number;
+  dateOfFirstActivity?: Date;
```
dtslvr: Please move up to keep the alphabetical order

dtslvr: Could we remove the optional flag (`?`)?

apps/api/src/app/portfolio/portfolio.service.ts
```diff
@@ -1953,6 +1953,7 @@ export class PortfolioService {
         return ['BUY', 'SELL'].includes(type);
       }).length,
       committedFunds: committedFunds.toNumber(),
+      dateOfFirstActivity: firstOrderDate,
```
dtslvr: Move down to keep the alphabetical order

apps/client/src/app/components/portfolio-summary/portfolio-summary.component.ts
```diff
@@ -84,10 +84,13 @@ export class GfPortfolioSummaryComponent implements OnChanges {
         this.precision = 0;
       }
 
-      if (this.user.dateOfFirstActivity) {
-        this.timeInMarket = formatDistanceToNow(this.user.dateOfFirstActivity, {
-          locale: getDateFnsLocale(this.language)
-        });
+      if (this.summary?.dateOfFirstActivity) {
```
dtslvr:
```
if (this.summary.dateOfFirstActivity) {
```
because we already check for `this.summary` above

apps/client/src/app/services/data.service.ts
```diff
@@ -564,6 +564,11 @@ export class DataService {
             }
           }
 
+          if (response.summary?.dateOfFirstActivity) {
+            response.summary.dateOfFirstActivity = parseISO(
+              response.summary.dateOfFirstActivity
+            );
+          }
           return response;
```
dtslvr, line 573: Add line break

dtslvr (changes requested):
Thanks for your contribution, @VahantSharma! 😎

apps/client/src/app/services/data.service.ts
VahantSharma: Added line break

VahantSharma:
@dtslvr Thanks for the review! Applied all changes.

PR #6116, opened 2025-12-30
Task/eliminate ngx-stripe

dtslvr:


KenTandrian:
Thanks for working on this, @dtslvr!

By removing `ngx-stripe`, I believe we might not need `@stripe/stripe-js` as well. Can we remove it?

dtslvr:
> By removing `ngx-stripe`, I believe we might not need `@stripe/stripe-js` as well. Can we remove it?

Great point, thanks @KenTandrian! I will remove that as well.

PR #6109, opened 2025-12-29
Task/upgrade to Nx 22.3.3 and Angular 21.0

KenTandrian:
Hi @dtslvr, this PR resolves #5943. Please take a look :)

### Changes

* **Framework:** Upgraded Angular from `20.3.x` to `21.0.6`.
* **Tooling:** Upgraded Nx from `22.1.3` to `22.3.3`.
* **Ecosystem:** Bumped third-party wrappers to their Angular 21 compatible versions:
  * `ngx-markdown` (v21) and `marked` (v17)
  * `ngx-device-detector` (v11)

### Context: Migration Roadmap

*  **Stage 1:** Stabilize Tooling (Nx 22 + Angular 20.3)
*  **Stage 2:** Upgrade Ecosystem:
    * Storybook 10 + i18n
    * Jest v30
    * Markdown (`ngx-markdown` and `marked` v16)
    * Remove unused Stripe dependencies
*  **Stage 3:** Final Framework Jump (Angular 21) — *This PR*

### Key Upgrade Notes

* **Library Alignment:** Third-party libraries like `ngx-device-detector` and `ngx-markdown` were upgraded due to strict peer dependency compatibility with Angular 21, preventing `ERESOLVE` warnings during future installs.
* When running migrations, Nx automatically changed the `tsconfig.json` files of the apps and libs to use `moduleResolution: bundler`, `module: preserve` and set `lib` to have `es2022`. Therefore, I set some of them as the default in `tsconfig.base.json`.
* `CommonModule` removals were done by Nx migration.

Issue #5943, opened 2025-11-14
Upgrade to Nx 22


dtslvr:
The goal of this issue is to upgrade this workspace to _Nx_ version `22` according to the [upgrade guide](https://github.com/ghostfolio/ghostfolio/blob/main/DEVELOPMENT.md#nx).

When I run the upgrade, I get the following error:

```
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: ghostfolio@2.222.0
npm error Found: @angular-devkit/build-angular@20.2.2
npm error node_modules/@angular-devkit/build-angular
npm error   dev @angular-devkit/build-angular@"20.3.9" from the root project
npm error   peerOptional @angular-devkit/build-angular@">= 18.0.0 < 21.0.0" from @nx/angular@21.5.1
npm error   node_modules/@nx/angular
npm error     dev @nx/angular@"22.0.4" from the root project
npm error   1 more (@storybook/angular)
npm error
npm error Could not resolve dependency:
npm error dev @angular-devkit/build-angular@"20.3.9" from the root project
npm error
npm error Conflicting peer dependency: @angular/compiler@20.3.9
npm error node_modules/@angular/compiler
npm error   peer @angular/compiler@"20.3.9" from @angular/localize@20.3.9
npm error   node_modules/@angular/localize
npm error     dev @angular/localize@"20.3.9" from the root project
npm error     peerOptional @angular/localize@"^20.0.0" from @angular-devkit/build-angular@20.3.9
npm error     node_modules/@angular-devkit/build-angular
npm error       dev @angular-devkit/build-angular@"20.3.9" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

Raj-G07:
Can i take up this issue?

dtslvr:
How are you progressing with this issue, @Raj-G07?

Raj-G07:
Can i raise a pr on Monday?

KenTandrian:
@dtslvr I have seen this problem when migrating Nx in another project. This happens especially because running `npx nx migrate latest` today attempts to perform **three major upgrades simultaneously**:

1. **Nx:** v21.5.1  → v22.3.3
2. **Angular:** v20.2 → v21
3. **Storybook:** v9 → v10

[...]

PR #6117, opened 2025-12-31
Task/upgrade to marked 16.4.2 and ngx-markdown 20.1.0

KenTandrian:
Hi @dtslvr, this PR is related to #5943. Please take a look :)

### Changes

* **Library:** Upgraded `ngx-markdown` to `v20.1.0`.
* **Core Dependency:** Upgraded `marked` to `v16.4.2`.
* **Cleanup:** Removed the explicit `marked` reference from `project.json` scripts.
  * *Reason:* Since `ngx-markdown` v17+, the library automatically handles the `marked` import. The global script entry is now dead code and has been removed.

### Context

I am isolating this upgrade because it involves both a dependency bump and a build configuration cleanup.

The `marked` library recently introduced breaking changes to its distribution files (removing `marked.min.js`). However, per the `ngx-markdown` release notes ([v17.0.0](https://github.com/jfcere/ngx-markdown/releases/tag/v17.0.0)), we no longer need to include this script manually in our build config at all. This PR cleans that up, resolving the breaking change by simply removing the unnecessary dependency reference.

By merging this separately, we ensure our documentation and content rendering works correctly on the new engine before we introduce the complexity of Angular 21, which will also reduce the scope of PR #6109.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Lifted the asset profile identifier editing restriction for `MANUAL` data sources in the asset profile details dialog of the admin control panel
 - Improved the language localization for German (`de`)
 - Upgraded `angular` from version `20.2.4` to `20.3.9`
+- Upgraded `marked` from version `15.0.4` to `16.4.2`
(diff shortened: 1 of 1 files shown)


dtslvr (approved):

PR #6125, opened 2025-12-31
chore: migrate deprecated animation providers

adityadewan22-hub:
Migrates deprecated provideAnimations() and provideNoopAnimations() to their
module-based equivalents across the application.
Let me know if you need anything else from my side.

### DIFF
--- a/apps/client/src/main.ts
+++ b/apps/client/src/main.ts
@@ -21,7 +21,6 @@ import {
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { bootstrapApplication } from '@angular/platform-browser';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { RouterModule, TitleStrategy } from '@angular/router';
 import { ServiceWorkerModule } from '@angular/service-worker';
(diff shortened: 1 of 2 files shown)


adityadewan22-hub:
Ready for review. 
Open to changes or feedback
Thanks

dtslvr (changes requested):
Thanks for your suggestion, @adityadewan22-hub!

It seems that `BrowserAnimationsModule` and `NoopAnimationsModule` are also deprecated. To keep the codebase modern, we should avoid adding these.

adityadewan22-hub:
Got it. @dtslvr 
I noticed that TopHoldingsComponent uses an Angular animation trigger.  Since both the provider-based and module-based animation APIs are deprecated, I wanted to confirm the preferred direction here.

Would you prefer that I:-
1 remove the deprecated animation providers for now and leave the existing trigger as-is,
 or
2 this animation to a CSS-based solution and remove Angular animations entirely?
happy to work in either direction and open to feedback
Thanks

adityadewan22-hub:
I have removed the deprecated browserAnimationsmodule and NoopAnimationsModule without a replacement.
The expand/collapse animation continues to work. And I have kept the changes minimal
Let me know if you need anything else from my side. 
Thanks @dtslvr

PR #6151, opened 2026-01-03
Bugfix/fix filtering by asset class in get holdings endpoint

dtslvr:


dtslvr:
The _Import Dividends_ dialog using `api/v1/portfolio/holdings?assetClasses=EQUITY,FIXED_INCOME` includes the cash holdings:

Screenshot 2026-01-03 at 21 17 12

This PR attempts to fix the filtering by asset class.

cc @KenTandrian

apps/api/src/app/order/order.service.ts
KenTandrian, line 331: Update the JSDoc documentation to include `filters`?

KenTandrian:
I am not sure that filtering the activities after the iterations is a good practice. Since all the activities here have an asset class of `LIQUIDITY`, why don't we filter them upfront?

apps/api/src/app/order/order.service.ts
dtslvr: Thanks! 👍🏻

dtslvr:
Thanks for the feedback, @KenTandrian!

Could you point out where you would suggest applying the filter?

KenTandrian:
Perhaps before this line:

https://github.com/ghostfolio/ghostfolio/blob/5101c406b463c395e1bca72cfadfee37c5b764fe/apps/api/src/app/order/order.service.ts#L345

We can add something similar to this:

```ts
// Filter out asset class filters
const filtersByAssetClass = filters?.filter(({ type }) => {
  return type === 'ASSET_CLASS';
});

// If asset class filters are present and none of them is liquidity, return an empty response
if (
  filtersByAssetClass?.length > 0 &&
  !filtersByAssetClass.find(({ id }) => id === AssetClass.LIQUIDITY)
) {
  return {
    activities: [],
    count: 0
  };
}
```

What do you think?

dtslvr:
Thanks for the hint, @KenTandrian! 👍🏻

PR #6159, opened 2026-01-04
Task/move admin service to UI library

### LINKED ISSUE #5884: Allow non-relative project imports from within the same project
Currently, our ESLint configuration flags an error for imports that use a path alias to import a file from within the same project.

For example, if you are in `apps/client/src/app/components/access-table/access-table.component.ts` and you write:

```ts
import { NotificationService } from '@ghostfolio/client/core/notification/notification.service';
```

The linter produces the following error:

```
Projects should use relative imports to import from other files within the same project. 
Use "./path/to/file" instead of import from "@ghostfolio/client/core/notification/notification.service" eslint (@nx/enforce-module-boundaries)
```

Image

## Why this is an issue

While technically correct, this rule forces developers to use relative paths (e.g., `../../core/notification/notification.service`). These can be brittle, hard to read, and difficult to maintain when files are moved. Using the full path alias (`@ghostfolio/client/...`) is often cleaner and more consistent.

## The Solution

The `@nx/enforce-module-boundaries` rule has an option, `allowCircularSelfDependency`, which, when set to `true`, disables this specific check and allows developers to use path aliases for imports within the same project.

## Action Item
We need to update our root `eslint.config.cjs` to add this property.

Current Configuration:
```js
// ...
  {
    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx'],
    rules: {
      '@nx/enforce-module-boundaries': [
        'warn',
        {
          enforceBuildableLibDependency: true,
          allow: [],
          depConstraints: [
            {
              sourceTag: '*',
              onlyDependOnLibsWithTags: ['*']
            }
          ]
        }
      ],
// ...
```

Proposed Configuration:

```JavaScript
// ...
  {
    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx'],
    rules: {
      '@nx/enforce-module-boundaries': [
        'warn',
        {
          allowCircularSelfDependency: true, // add this line
          enforceBuildableLibDependency: true,
          allow: [],
          depConstraints: [
            {
              sourceTag: '*',
              onlyDependOnLibsWithTags: ['*']
            }
          ],
        }
      ],
// ...
```

## Acceptance Criteria

* The `eslint.config.cjs` file is updated with `"allowCircularSelfDependency": true`.
* Developers can import files from within the same project using the full path alias (e.g., `@ghostfolio/client/services/admin.service`) without triggering an `@nx/enforce-module-boundaries` ESLint error.

## Next Steps

Ideally, we should also start enforcing module boundaries to prevent circular dependencies. Example: `api` can import from `common`, but not vice versa.

@dtslvr Let me know what you think.

KenTandrian:
Hi @dtslvr, this PR closes #5884. Please take a look :)

### Changes

We have now **completely** resolved the circular dependency problems between `client` and `ui` modules.
* Moved `AdminService` to `ui` library.
* Created `GF_ENVIRONMENT` InjectionToken to decouple `AdminService` from client environment.
* Updated `AdminService` to use `GF_ENVIRONMENT`.
* Updated main application to provide `GF_ENVIRONMENT`.
* Updated all `AdminService` imports to `@ghostfolio/ui/services`.

### DIFF
--- a/libs/ui/src/lib/services/admin.service.ts
+++ b/libs/ui/src/lib/services/admin.service.ts
@@ -21,8 +21,8 @@ import {
   Filter
 } from '@ghostfolio/common/interfaces';
 import { DateRange } from '@ghostfolio/common/types';
+import { GF_ENVIRONMENT, GfEnvironment } from '@ghostfolio/ui/environment';
 import { DataService } from '@ghostfolio/ui/services';
-import { GfEnvironment, GF_ENVIRONMENT } from '@ghostfolio/ui/tokens';
 
 import { HttpClient, HttpHeaders, HttpParams } from '@angular/common/http';
 import { Inject, Injectable } from '@angular/core';
--- a/apps/client/src/environments/environment.prod.ts
+++ b/apps/client/src/environments/environment.prod.ts
@@ -1,4 +1,4 @@
-import type { GfEnvironment } from '@ghostfolio/ui/tokens';
+import type { GfEnvironment } from '@ghostfolio/ui/environment';
 
 export const environment: GfEnvironment = {
   lastPublish: '{BUILD_TIMESTAMP}',
--- a/apps/client/src/environments/environment.ts
+++ b/apps/client/src/environments/environment.ts
@@ -1,7 +1,8 @@
+import type { GfEnvironment } from '@ghostfolio/ui/environment';
+
 // This file can be replaced during build by using the `fileReplacements` array.
 // `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
 // The list of file replacements can be found in `angular.json`.
-import type { GfEnvironment } from '@ghostfolio/ui/tokens';
 
 export const environment: GfEnvironment = {
   lastPublish: null,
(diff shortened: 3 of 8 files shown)


libs/ui/src/lib/tokens/env.token.ts
```diff
@@ -0,0 +1,11 @@
+import { InjectionToken } from '@angular/core';
+
+export interface GfEnvironment {
```
dtslvr: Could you reuse this in `apps/client/src/environments/environment.ts` and `apps/client/src/environments/environment.prod.ts` please?

libs/ui/src/lib/services/admin.service.ts
```diff
@@ -22,22 +22,22 @@ import {
...
-
 @Injectable({
   providedIn: 'root'
 })
 export class AdminService {
   public constructor(
     private dataService: DataService,
-    private http: HttpClient
+    private http: HttpClient,
+    @Inject(GF_ENVIRONMENT) private environment: GfEnvironment
```
dtslvr: Move up ⬆️  (`environment` after `dataService`)

Issue #6155, opened 2026-01-04
Cash balances included in overview


dtslvr:
**Bug Description**

Cash balances (_Liquidity_) are currently included in the overview.

**Expected Behavior**

The overview should include only (invested) assets. Cash balances should be excluded.

**Environment**

- Ghostfolio Version `2.228.0`

**Additional Context**

This appears to be a regression introduced in https://github.com/ghostfolio/ghostfolio/pull/5650.

omkarg01:
Can I work on this?

dtslvr:
Thanks for offering to help, @omkarg01!

This one is a bit tricky, so it might be a good idea to sync with @KenTandrian to align on the approach.

omkarg01:
Sounds good, I’ll sync with @KenTandrian and align on the approach before moving forward.

KenTandrian:
This seems to be caused by changes on `OrderService.getOrdersForPortfolioCalculator` method, which now includes cash calculations. This method is being used here:
https://github.com/ghostfolio/ghostfolio/blob/5101c406b463c395e1bca72cfadfee37c5b764fe/apps/api/src/app/portfolio/portfolio.service.ts#L988-L992

What if we add a filter by asset classes in the params inside `DataService.fetchPortfolioPerformance`?
https://github.com/ghostfolio/ghostfolio/blob/5101c406b463c395e1bca72cfadfee37c5b764fe/libs/ui/src/lib/services/data.service.ts#L653-L656

fuzzysb:
This version is currently double counting my cash values in the Overview Summary page, it shows my total Assets as the value that includes the value of assets including the cash amounts. then adds  the cash value again which shows my net worth as significantly over what it should be.

I understand you are introducing cash held to cater for currency performance, but this should only be cash accounts not held in the home currency, and should only be showing the net gain or loss from currency performance as an asset, and not the actual balance recorded as an asset.

this needs to be regressed until these changes are made

Don't get me wrong, it was both nice to see and the disappointment of realising why, if my net worth had increased by as much as ghostfolio is reporting i would retire today

dtslvr:
Resolved via https://github.com/ghostfolio/ghostfolio/pull/6171.

PR #6171, opened 2026-01-06
Bugfix/prevent double counting of cash in net worth

### LINKED ISSUE #6169: [BUG] Cash balances are showing double in account's graph
**Bug Description**

Cash balances for accounts are being doubled in account balance graph.

**To Reproduce**

1. Go to accounts
2. Create a new account and set the cash balance to say 30,000
3. Click on newly created account and check graph (it should say 60,000)

**Expected Behavior**

Graph should replicate account balance.

**Screenshots**

Image

**Environment**

- Ghostfolio Version 2.228.0
- Self-hosted
- Safari
- macOS

KenTandrian:
Hi @dtslvr, this PR resolves #6169. Please take a look :)

### Changes

* Fixed a bug in `PortfolioCalculator` where Net Worth for cash assets was being calculated incorrectly (double-counted).
* Updated `portfolio-calculator-cash.spec.ts` unit tests to reflect the correct calculation logic.

#### Issue in Account Details Dialog

The issue was caused by `totalAccountBalance` being added to `totalCurrentValue` in the Net Worth calculation. Since `OrderService` converts cash balances into synthetic activities (which are already included in `totalCurrentValue`), adding `totalAccountBalance` again resulted in cash being counted twice.

https://github.com/ghostfolio/ghostfolio/blob/ac67a6e190e505c964d4fbfb891d537ee679c905/apps/api/src/app/portfolio/calculator/portfolio-calculator.ts#L608-L610

#### Issue in Summary Page

This actually also happens in `PortfolioService.getSummary()`, as reported in https://github.com/ghostfolio/ghostfolio/issues/6155#issuecomment-3709877815.

https://github.com/ghostfolio/ghostfolio/blob/ac67a6e190e505c964d4fbfb891d537ee679c905/apps/api/src/app/portfolio/portfolio.service.ts#L1937-L1941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed the net worth calculation to prevent the double counting of cash positions
 - Fixed the filtering by asset class in the endpoint `GET api/v1/portfolio/holdings`
 - Fixed the case-insensitive sorting in the accounts table component
 


apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash.spec.ts
```diff
@@ -249,6 +257,7 @@ describe('PortfolioCalculator', () => {
 
       /**
        * Net performance with currency effect: (1000 * 0.86) - (1000 * 0.85) = 10 CHF
+       * Total account balance: 1000 USD * 0.85 = 850 CHF (using the exchange rate on 2023-12-31)
```
dtslvr: Why not on `2024-01-01`?

KenTandrian:
@dtslvr I have made some changes to the logic, because it seems like modifying `getOrdersForPortfolioCalculator` in the previous PR has impacted a lot of other functionalities. Therefore, I have introduced a new parameter called `withCash` so that creating synthetic orders for cash is now optional.

## The Summary Page Issue - `currentValueInBaseCurrency` Calculation

The `currentValueInBaseCurrency` case is a bit tricky. This variable is calculated only in `RoaiPortfolioCalculator.calculateOverallPerformance` and it depends on `positions` variable passed from `computeSnapshot`.

https://github.com/ghostfolio/ghostfolio/blob/a493afb52656199356b4e20ef66ec2b90b58a245/apps/api/src/app/portfolio/calculator/roai/portfolio-calculator.ts#L45-L47

We need to handle 2 requirements at the same time: showing them in holdings page (hence cash orders need to be included in `positions`) AND exclude cash orders from `currentValueInBaseCurrency`. The iteration of `positions` happens here:

https://github.com/ghostfolio/ghostfolio/blob/a493afb52656199356b4e20ef66ec2b90b58a245/apps/api/src/app/portfolio/portfolio.service.ts#L569-L588

Removing all cash orders from `positions` is a no go because the `holdings` endpoint depends on it. Therefore, I have added a new variable called `includeInTotalAssetValue` (feel free to suggest better naming if any) and filter it out in `RoaiPortfolioCalculator.calculateOverallPerformance` method.

## The Overview Page Issue

https://github.com/ghostfolio/ghostfolio/blob/a493afb52656199356b4e20ef66ec2b90b58a245/apps/api/src/app/portfolio/calculator/portfolio-calculator.ts#L721

The portfolio value in the overview page is using a *different* variable from the summary page. It is coming from `chart` variable which depends on `PortfolioSnapshot.historicalData`. Hence, I have excluded cash orders from historical data by excluding them from `accumulatedValuesByDate` and `valuesBySymbol` using the same `includeInTotalAssetValue` boolean.

As cash orders is not available in `historicalData` anymore, I have changed the test case to match `position` instead.

## The Portfolio Allocation Page Issue

It turns out that cash values are also double calculated in the portfolio allocation page. Excluding them by not passing `withCash` parameter (defaults to `false`) solved the issue.

fuzzysb:
Thanks for your work on this @KenTandrian

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator.ts
```diff
@@ -34,7 +34,9 @@ export class RoaiPortfolioCalculator extends PortfolioCalculator {
     let totalTimeWeightedInvestment = new Big(0);
     let totalTimeWeightedInvestmentWithCurrencyEffect = new Big(0);
 
-    for (const currentPosition of positions) {
+    for (const currentPosition of positions.filter((position) => {
```
dtslvr: `positions.filter(({ includeInTotalAssetValue }) => {`

KenTandrian:
@dtslvr Unfortunately I cannot reproduce that issue. The values are the same in my case.

I see that the percentages in Portfolio Analysis page actually comes from `netPerformancePercentage` and `netPerformancePercentageWithCurrencyEffect` values returned by `PortfolioService.getPerformance`. These values are derived from the historical data chart, similar to the Home Overview page.

Cash synthetic orders are already excluded from `historicalData`, so I believe those 2 percentages should not be impacted anymore, but I might be wrong :)

In my case, it sometimes changes only due to live price changes from `getQuotes`.

apps/api/src/app/portfolio/calculator/roai/portfolio-calculator-cash.spec.ts
dtslvr, line 160: When I add this

```
{
  // Ignored future account balance
  accountId,
  id: randomUUID(),
  date: parseDate('2050-12-31'),
  value: 0,
  valueInBaseCurrency: 0
}
```
it seems to affect the results, even though this item is not in the date range. Is a date range filter currently missing?

dtslvr: @KenTandrian I have created a dedicated issue for this: https://github.com/ghostfolio/ghostfolio/issues/6185

fuzzysb:
Thanks for your work on this guy's. I have tested and is as I would expect to see

PR #6176, opened 2026-01-07
Task/improve sorting in account api #6174

### LINKED ISSUE #6174: Improve sorting in account endpoint
The goal of this issue is to improve the query which orders the accounts by name in the `GET /api/v1/account` endpoint. Please ensure the sorting is case-insensitive to maintain a consistent order regardless of capitalization.

omkarg01:
Hi @dtslvr,  this PR resolves #6174

### DIFF
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -277,6 +277,15 @@ export class PortfolioService {
       });
     }
 
+    accounts = [...accounts].sort((a, b) => {
+      const nameA = (a.name || '').trim().toLowerCase();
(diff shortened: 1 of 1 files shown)


omkarg01:
> Am I correct that case-insensitive sorting isn't possible in the query? (Reference: [prisma/prisma#5068](https://github.com/prisma/prisma/issues/5068))

Yes, its works with filters but not sorting.

dtslvr:
> Yes, its works with filters but not sorting.

Thanks, @omkarg01. I don't think it makes sense to handle the sorting on the backend then. Let's do it on the client side (see https://github.com/ghostfolio/ghostfolio/pull/6178).

PR #6198, opened 2026-01-16
Bugfix/numeric parsing error in X-ray page

dtslvr:


Lyndros:
Thomas I will check it asap with a fresh import. @dtslvr

Lyndros (commented):
Looks good, now renders the page but having issues to import back a full export of my data (around 500 activities), no scroll appears when import and could not accept, even if individually can be selected.

It would be great to have a full backup import in settings, but could not find it.

Lyndros (approved):

Issue #6201, opened 2026-01-17
Ghostfolio in Korean


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Korean-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.ko.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Add the following changelog entry:
```
## Unreleased

### Changed

- Improved the language localization for Korean (`ko`)
```
6. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

hassarch:
heyy i have went through the issue , can you assign it to me or can i start working on it??

dtslvr:
Sure @hassarch, please create a pull request with your translations. Since this is an ongoing task, I won't formally assign the issue, but I'm looking forward to your contribution!

hassarch:
i have created a PR can you please check it and lemme know so that i can continue on the other translations

PR #6213, opened 2026-01-19
Task/improve Korean (ko) localization

hassarch:
- Translated 7 missing entries from state='new' to state='translated'
- Updated CHANGELOG.md with Korean localization improvements

Related issue: https://github.com/ghostfolio/ghostfolio/issues/6201


dtslvr:
Could you please take a look at these suggested changes when you have a moment, @yootaebong?

Issue #6202, opened 2026-01-17
Ghostfolio in Chinese


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Chinese-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.zh.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Add the following changelog entry:
```
## Unreleased

### Changed

- Improved the language localization for Chinese (`zh`)
```
6. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

Eshaan-byte:
Hi, @dtslvr can I work on this issue?

dtslvr:
> Hi, [@dtslvr](https://github.com/dtslvr) can I work on this issue?

Sure @Eshaan-byte, please create a pull request with your translations. Since this is an ongoing task, I won't formally assign the issue, but I'm looking forward to your contribution!

poterkao:
Actually, there 2 different Chinese in Asia. 
1. Simplified Chinese -  It's used in China normally. 
2. Traditional Chinese - It's used in Taiwan / Hong Kong.

Most software programs typically use zh-TW (Traditional Chinese) and zh-CN (Simplified Chinese) to distinguish between these two Chinese, as they differ not only in writing style but also in translation terminology and currency.

Currently, Ghostfolio uses "Chinese-zh" in its language settings, which differs from zh-TW used in Taiwan/Hong Kong. Perhaps separating them into two different file systems could be considered:

message.zh-tw.xlf=>Traditional Chinese
message.zh-cn.xlf=>Simplified Chinese

PR #6245, opened 2026-01-25
Feature/improve zh localization

KenTandrian:
Hi @dtslvr, this PR is related to #6202. Please take a look :)

### Changes
* Fixed translations that had incorrect context. For example, "free" was translated to mean freedom instead of free of charge.
* Refactored fragmented i18n strings into a single unit in the pricing page to prevent awkward sentence structures in languages with different word orders (i.e., in Chinese).
* Updated other translations in Chinese.

### DIFF
--- a/apps/client/src/app/pages/pricing/pricing-page.html
+++ b/apps/client/src/app/pages/pricing/pricing-page.html
@@ -195,9 +195,9 @@
                 </p>
                 <ul class="list-unstyled mb-3">
                   <li class="mb-2">
-                    <span i18n>
-                      Everything in <strong>Basic</strong>, plus:
-                    </span>
+                    <ng-container i18n
+                      >Everything in <strong>Basic</strong>, plus</ng-container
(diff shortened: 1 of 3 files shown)


apps/client/src/app/pages/pricing/pricing-page.html
```diff
@@ -195,12 +195,9 @@ <h4 class="align-items-center d-flex flex-grow-1 m-0">
                 </p>
                 <ul class="list-unstyled mb-3">
                   <li class="mb-2">
-                    <span
-                      ><ng-container i18n>Everything in</ng-container>
-                      <ng-container>&nbsp;</ng-container>
-                      <strong>Basic</strong>,
-                      <ng-container i18n>plus</ng-container>:</span
-                    >
+                    <ng-container i18n
```
dtslvr, line 198: The idea was to exclude `Basic` from localization, however I see your point.

PR #6254, opened 2026-01-26
Feature/extract top holdings from Yahoo Finance for ETF and mutual funds

KenTandrian:
Hi @dtslvr, I see that ETF and mutual funds asset profile fetched from Yahoo Finance does not include holdings, while the response data from Yahoo Finance actually have that information. Therefore, I added some lines of codes to map them into `holdings` field.

### Sample result

Screenshot 2026-01-27 at 00 15 57

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added the ability to fetch top holdings for ETF and mutual fund assets from _Yahoo Finance_
 
+### Changed
+
+- Removed the deprecated `firstBuyDate` in the portfolio calculator
+
 ## 2.234.0 - 2026-01-30
 
 ### Changed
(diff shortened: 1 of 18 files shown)


apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -217,6 +217,13 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
             });
           }
         }
+
+        response.holdings = assetProfile.topHoldings.holdings.map((holding) => {
```
dtslvr: `... .map(({ holdingName, holdingPercent }) => {`

apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -217,6 +217,13 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
...
+          return {
+            name: holding.holdingName,
```
dtslvr: Can you use
```
name: this.formatName({ longName: holdingName }),
```
and change all the params of `formatName()` to optional?

Issue #3588, opened 2024-07-20
Ghostfolio en Español


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Spanish-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.es.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

DavidReque:
Hi, I would like to participate

GerardPolloRebozado:
Hi I was translating some more lines from the spanish transaltions and found that one sentence is repeated, the only difference is the space before and after. they are on line [3986](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.es.xlf#L3986)and [3914](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.es.xlf#L3914)
Is this expected to be like that?

dtslvr:
> Is this expected to be like that?

I have noticed that too. I still haven't found out why these duplicates exist.

jackcooper20:
hey @dtslvr can i contribute to this ?

dtslvr:
I have extended the instructions by a changelog entry.

slegarraga:
Hi! I'm a native Spanish speaker from Chile. I'd love to help with the Spanish translations. Will submit a PR with the remaining untranslated strings.

PR #6261, opened 2026-01-30
Task/improve Spanish translation

Lyndros:
Improved spanish translation, some strings need to be translated and another ones were corrected due to minor errors.
i.e: Cerca por Cerrar (Close)

Related issue: #3588


Lyndros:
Done!

PR #6279, opened 2026-02-05
Bugfix/fix exception when fetching top holdings in Yahoo Finance service

dtslvr:


apps/api/src/services/data-provider/data-enhancer/yahoo-finance/yahoo-finance.service.ts
```diff
@@ -206,26 +206,26 @@ export class YahooFinanceDataEnhancerService implements DataEnhancerInterface {
...
+        ).flatMap((sectorWeighting) => {
+          return Object.entries(sectorWeighting).map(([sector, weight]) => {
+            return {
               name: this.parseSector(sector),
               weight: weight as number
-            });
-          }
-        }
-
-        response.holdings = assetProfile.topHoldings.holdings.map(
```
dtslvr, line 221: `holdings` could be undefined

Issue #6246, opened 2026-01-25
[LSC] Enable Strict Null Checks & Nullish Coalescing Workspace-wide


KenTandrian:
## Large Scale Change (LSC): Improve Type Safety with Strict Null Checks

We aim to enable `strictNullChecks` across the entire workspace to improve type safety and robust code quality. This is also a prerequisite for enabling the `@typescript-eslint/prefer-nullish-coalescing` ESLint rule, which helps avoid common bugs involving `0` or `""` being treated as falsy.

This will also resolve this kind of linting warnings and errors:
<img width="681" height="72" alt="Image" src="https://github.com/user-attachments/assets/7e42250b-8663-40d0-90e6-ca1f3434383b" />

To make this manageable and reviewable, we will roll this out incrementally detailed by project. For each project, we will:
1. Enable `strictNullChecks` in each project's `tsconfig`.
2. Set `@typescript-eslint/prefer-nullish-coalescing` to `error` in each project's `eslint.config.cjs` (overriding the global `warn`).
3. Enable `strictNullChecks` in `tsconfig` globally.
4. Set `@typescript-eslint/prefer-nullish-coalescing` to `error` globally.

## Implementation Plan

We will proceed in the following order, ensuring each project builds and lints cleanly before moving to the next.

### 1. Common Library (`libs/common`)
- [x] Verify/Ensure `strictNullChecks` is enabled in `tsconfig.json`.
- [x] Resolve any existing type errors.
- [x] Enforce `@typescript-eslint/prefer-nullish-coalescing` as error for this project.

### 2. UI Library (`libs/ui`)
- [x] Enable `strictNullChecks` in `tsconfig.json` (currently explicit `strict: false`).
- [x] Refactor components to handle null/undefined explicitly.
- [x] Enforce `@typescript-eslint/prefer-nullish-coalescing` as error for this project.

### 3. Client App (`apps/client`)
- [ ] Enable `strictNullChecks` in `tsconfig.json`.
- [ ] Fix template and logic issues exposed by strict checks.
- [ ] Enforce `@typescript-eslint/prefer-nullish-coalescing` as error for this project.

### 4. API App (`apps/api`)
- [ ] Enable `strictNullChecks` in `tsconfig.json`.
- [ ] Ensure backend logic correctly handles optional fields and database results.
- [ ] Enforce `@typescript-eslint/prefer-nullish-coalescing` as error for this project.

## Final Goal
[...]

PR #6290, opened 2026-02-08
Task/improve type safety for GfValueComponent

KenTandrian:

### Changes

* Resolve `precision` input type error.
* Added `hasPrecision` getter and `getFormatOptions` method.
* Changed `precision` input to use the modern signal input instead of input decorator.

### Error resolved

```console
libs/ui/src/lib/value/value.component.ts:142:5 - error TS2322: Type 'number | undefined' is not assignable to type 'number'.
  Type 'undefined' is not assignable to type 'number'.

142     this.precision = this.precision >= 0 ? this.precision : undefined;
        ~~~~~~~~~~~~~~
```

### DIFF
--- a/libs/ui/src/lib/value/value.component.ts
+++ b/libs/ui/src/lib/value/value.component.ts
@@ -54,7 +54,7 @@ export class GfValueComponent implements OnChanges {
     };
   });
 
-  private get hasPrecision(): boolean {
+  private get hasPrecision() {
     const precision = this.precision();
     return precision !== undefined && precision >= 0;
   }


libs/ui/src/lib/value/value.component.ts
```diff
@@ -42,6 +42,14 @@ export class GfValueComponent implements OnChanges {
   public isString = false;
   public useAbsoluteValue = false;
 
+  get hasPrecision(): boolean {
```
dtslvr: `private get hasPrecision() {` ❓

(and move down ⬇️)

KenTandrian: Thanks for pointing that out. I have changed `hasPrecision` to a `private` getter.

However, seems like I cannot move it below because of this TypeScript ESLint warning:
Screenshot 2026-02-08 at 15 36 25

I believe the general order is:
* Fields
* Constructors
* **Getters** (defined with the `get` keyword)
* Setters (defined with the `set` keyword)
* Methods

dtslvr (approved):

PR #6307, opened 2026-02-11
Task/improve type safety in accounts table component

KenTandrian:

### Changes

* Reverted the implementation of `computed` signal for `dataSource`, because initializing `MatTableDataSource` for every signal update will negatively impact performance.
* Removed unused `unsubscribeSubject` variable and `ngOnDestroy`.
* Changed methods to `protected`.
* Changed all inputs to signal inputs.
* Changed all outputs to signal outputs.
* Removed `deviceType` input as it is not used in the component and template.

### DIFF
--- a/libs/ui/src/lib/accounts-table/accounts-table.component.ts
+++ b/libs/ui/src/lib/accounts-table/accounts-table.component.ts
@@ -8,12 +8,11 @@ import { CommonModule } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  EventEmitter,
-  Output,
   computed,
   effect,
   inject,
   input,
+  output,
   viewChild
 } from '@angular/core';
 import { MatButtonModule } from '@angular/material/button';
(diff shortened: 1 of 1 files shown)


libs/ui/src/lib/accounts-table/accounts-table.component.ts
```diff
@@ -135,9 +125,21 @@ export class GfAccountsTableComponent implements OnDestroy {
...
+    });
+
+    // Reactive view connection
+    effect(() => {
+      this.dataSource.sort = this.sort();
+    });
   }
 
-  public onDeleteAccount(aId: string) {
+  protected onDeleteAccount(aId: string) {
```
dtslvr, line 142: Why `protected`?

dtslvr (approved):

libs/ui/src/lib/accounts-table/accounts-table.component.ts
KenTandrian: Because it is not being used outside of the component and template. In my opinion, `public` is usually used for inputs/outputs, while for methods `protected` is enough.

PR #6309, opened 2026-02-12
Task/improve asset profile dialog routing

### LINKED ISSUE #6112: Improve asset profile dialog routing in Admin Control panel
The goal of this issue is to refactor the _Angular_ routing for the [asset profile dialog](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog) in the _Admin Control_ panel. We would like a cleaner URL structure and move from using query parameters to using path segments for the data source / symbol.

Change from `/admin/market-data?assetProfileDialog=true&dataSource={dataSource}&symbol={symbol}` to `/admin/market-data/{dataSource}/{symbol}`.

Use https://github.com/ghostfolio/ghostfolio/pull/6104 as a reference for the implementation.

carlos-ds:
Closes #6112

### DIFF
--- a/apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
+++ b/apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
@@ -206,23 +206,18 @@ export class GfAdminMarketDataComponent
     this.displayedColumns.push('comment');
     this.displayedColumns.push('actions');
 
-    if (this.route.snapshot.routeConfig.path === 'market-data/create') {
-      this.openCreateAssetProfileDialog();
-    } else {
-      this.route.paramMap
-        .pipe(takeUntil(this.unsubscribeSubject))
-        .subscribe((paramMap) => {
-          const dataSource = paramMap.get('dataSource') as DataSource;
-          const symbol = paramMap.get('symbol');
+    this.route.paramMap
+      .pipe(takeUntil(this.unsubscribeSubject))
+      .subscribe((paramMap) => {
+        const dataSource = paramMap.get('dataSource') as DataSource;
+        const symbol = paramMap.get('symbol');
 
-          if (dataSource && symbol) {
-            this.openAssetProfileDialog({
-              dataSource,
-              symbol
-            });
-          }
-        });
-    }
+        if (dataSource && symbol) {
+          this.openAssetProfileDialog({ dataSource, symbol });
+        } else if (this.route.snapshot.routeConfig.path.endsWith('/create')) {
+          this.openCreateAssetProfileDialog();
+        }
+      });
 
     this.userService.stateChanged
       .pipe(takeUntil(this.unsubscribeSubject))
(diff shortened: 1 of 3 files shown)


carlos-ds:
@dtslvr I've made this PR for #6112 , can you have a look?

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
```diff
@@ -203,22 +206,23 @@ export class GfAdminMarketDataComponent
...
-        ) {
-          this.openAssetProfileDialog({
-            dataSource: params['dataSource'],
-            symbol: params['symbol']
-          });
-        } else if (params['createAssetProfileDialog']) {
-          this.openCreateAssetProfileDialog();
-        }
-      });
+    if (this.route.snapshot.routeConfig.path === 'market-data/create') {
```
dtslvr: I would not nest the subscription into the `if` statement.
What do you think about:
```
this.route.paramMap
  .pipe(takeUntil(this.unsubscribeSubject))
  .subscribe((paramMap) => {
    const dataSource = paramMap.get('dataSource') as DataSource;
    const symbol = paramMap.get('symbol');

    if (dataSource && symbol) {
      this.openAssetProfileDialog({ dataSource, symbol });
    } else if (this.route.snapshot.routeConfig.path.endsWith('/create')) {
      this.openCreateAssetProfileDialog();
    }
  });
```

apps/client/src/app/components/admin-market-data/admin-market-data.component.ts
```diff
@@ -344,15 +348,15 @@ export class GfAdminMarketDataComponent
...
-        dataSource,
-        symbol,
-        assetProfileDialog: true
-      }
+    this.router.navigate([dataSource, symbol], {
+      relativeTo: this.route
     });
   }
 
+  public getRouterLinkToAdminControlMarketData(dataSource, symbol) {
```
dtslvr: `public getRouterLinkToAdminControlMarketData({ dataSource, symbol }: AssetProfileIdentifier) {`

Move up ⬆️

apps/client/src/app/components/holding-detail-dialog/holding-detail-dialog.component.ts
```diff
@@ -618,6 +618,14 @@ export class GfHoldingDetailDialogComponent implements OnDestroy, OnInit {
     }
   }
 
+  public getRouterLinkToAdminControlMarketDataForDataSourceAndSymbol() {
```
dtslvr: Use the same interface `public getRouterLinkToAdminControlMarketDataForDataSourceAndSymbol({ dataSource, symbol }: AssetProfileIdentifier) {` ❓

Move up ⬆️

dtslvr (changes requested):
Thanks for your contribution, @carlos-ds! 😎

carlos-ds:
@dtslvr I updated the PR according to your comments. But there's something wrong which I can't pinpoint yet.

For example, let's say I add a new profile, the call works fine (status 201):

```
POST https://localhost:4200/api/v1/admin/profile-data/YAHOO/R5L.SG
```

Immediately after that, the call to refresh the market data is done. But that always returns a 401 error because the Authorization header is not set.

```
https://localhost:4200/api/v1/admin/market-data?take=50
```

image

I had a look at the http interceptors but those work fine. Do you have any other clues on this?

dtslvr:
Thanks for the update, @carlos-ds!

I’m having trouble reproducing that exact behavior. When I try to add an asset profile, the `POST` request (including _Bearer_) to `api/v1/admin/profile-data/:dataSource/:symbol` is cancelled immediately and I am redirected to the home route 🤔

PR #6310, opened 2026-02-12
Bugfix/fix issue in annualized performance calculation

dtslvr:


libs/common/src/lib/calculation-helper.ts
dtslvr, line 26: @KenTandrian `growthFactor` can be `Infinity` 🤓

PR #6316, opened 2026-02-13
fix: scraper test returns meaningful errors (fixes #6313)

### LINKED ISSUE #6313: Scraper not giving meaningful errors
**Bug Description**

Tried adding a scraper config for my property from an Aussie valuation site.

E.g. https://www.property.com.au/nsw/ryde-2112/semple-st/1-pid-1507821/.
In browser console, selector `[data-testid="valuation-sub-brick-price-text"]` works to select the price.

This should yield the text "$2,970,000"

**To Reproduce**

1. Go to Admin Control -> Market Data
2. Create or edit a RE asset.
3. In scraping config enter the above URL and selector. Click "Test"

**Expected Behavior**

Test working, or at least telling me what went wrong.

**Screenshots**

Image

Image

**Environment**

- Ghostfolio Version 2.237.0 
- Self-hosted through Docker on Debian
- Browser: Chrome

irfanfaraaz:
## Description
Fixes #6313

When testing scraper configuration (Admin Control → Market Data → edit MANUAL asset → Scraper Configuration → Test), failures only showed a generic message: "Could not parse the current market price for …". There was no indication of the real cause (e.g. HTTP 429, selector not matching, or parse failure).

### Backend – `manual.service.ts` (`scrape()`)
- **Network:** Wrap `fetch` in try/catch and throw a clear error on failure (timeout, unreachable, etc.).
- **HTTP status:** If `!response.ok`, throw with status and a short explanation (e.g. rate limiting / blocking).
- **JSON path:** If the selector matches nothing, throw: "Selector did not match any value in the response. Check the path expression."
- **HTML selector:** If the selector matches no element (empty text), throw: "Selector matched no element on the page. Check the selector and that the page structure matches."
- **Object/array match:** If the matched value is an object/array, throw instead of stringifying to `[object Object]`.
- **Parse failure:** If `extractNumberFromString` returns `undefined`/NaN, throw with a short preview of the selected text.

### Backend – `admin.controller.ts` (`testMarketData`)
- In the catch block, use `error instanceof Error ? error.message : 'Scraper test failed. …'` so the client always receives a string message.

### Frontend – `asset-profile-dialog.component.ts`
- In `catchError`, read the message from `err.error?.message` (Nest response body) with fallbacks, and type the callback as `HttpErrorResponse` and `body` as `unknown` to satisfy lint.

## Testing
- [x] Test with URL that returns 429 → message includes "HTTP 429" and mentions blocking/rate-limiting.
- [x] Test with invalid selector → message says selector matched no element / no value.
- [x] Test with valid URL + selector that returns a number → test succeeds and shows price.

image

### DIFF
--- a/apps/api/src/app/admin/admin.controller.ts
+++ b/apps/api/src/app/admin/admin.controller.ts
@@ -259,7 +259,12 @@ export class AdminController {
     } catch (error) {
       Logger.error(error, 'AdminController');
 
-      throw new HttpException(error.message, StatusCodes.BAD_REQUEST);
+      const message =
+        error instanceof Error
+          ? error.message
+          : 'Scraper test failed. Check the URL, selector, and that the site returns the expected content.';
+
+      throw new HttpException(message, StatusCodes.BAD_REQUEST);
     }
   }
 
--- a/apps/api/src/services/data-provider/manual/manual.service.ts
+++ b/apps/api/src/services/data-provider/manual/manual.service.ts
@@ -276,24 +276,52 @@ export class ManualService implements DataProviderInterface {
   ): Promise<number> {
     let locale = scraperConfiguration.locale;
 
-    const response = await fetch(scraperConfiguration.url, {
-      headers: scraperConfiguration.headers as HeadersInit,
-      signal: AbortSignal.timeout(
-        this.configurationService.get('REQUEST_TIMEOUT')
-      )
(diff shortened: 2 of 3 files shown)


dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/pull/6322.

PR #6346, opened 2026-02-17
Feature/add new market data preset

### LINKED ISSUE #6339: Add new market data preset
The goal of this issue is to create a new [market data preset](https://github.com/ghostfolio/ghostfolio/blob/main/libs/common/src/lib/types/market-data-preset.type.ts) `WITHOUT_ACTIVITIES` for the _Markets_ section of the _Admin Control_ panel. This preset should filter the asset profile list by items that do not have no activities (`order`).

For inspiration, please refer to the implementation of `ETF_WITHOUT_COUNTRIES` and `ETF_WITHOUT_SECTORS`.

DavidReque:
Hi @dtslvr, this PR resolves #6339 

### Changes
- Added the new `WITHOUT_ACTIVITIES` preset to the shared `MarketDataPreset` type.
- Added the `WITHOUT_ACTIVITIES` filter in Admin Control → Markets (allFilters), labeled as Without Activities.
- Extended backend logic in `getMarketData()` so this preset returns only assets with `activitiesCount === 0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a _No Activities_ preset to the historical market data table of the admin control panel
 - Added support for custom cryptocurrencies defined in the database
 - Added support for the cryptocurrency _Sky_
-- Added a new market data preset (`WITHOUT_ACTIVITIES`) in the markets section of the admin control panel to filter asset profiles without activities
 
 ### Changed
 


DavidReque:
> Thanks for your contribution, @DavidReque! 😎
> 
> I have tested the changes and got the following error:
> 
> ```
> Invalid `prisma.marketData.findMany()` invocation:
> Assertion violation on the database: `too many bind variables in prepared statement, expected maximum of 32767, received 45711`
> ```
> 
> Any idea how to optimize the query?

Thanks for catching this

I’ve updated the implementation to avoid hitting PostgreSQL’s bind-parameter limit:

1) `WITHOUT_ACTIVITIES` is now filtered at DB level (`where.activities = { none: {} }`) instead of filtering in memory after fetching all rows.
2) The `marketData.findMany()` lookup for last prices was split into symbol chunks, so each query stays below the bind variable threshold.

This removes the `too many bind variables` error in large datasets while keeping the existing preset behavior intact.

Issue #5924, opened 2025-11-08
Trackinsight data gathering failing


pleich:
**Bug Description**

Currently, the `TrackinsightDataEnhancerService` is failing to fetch any country information for all ETFs that are part of the portfolio. In the logs I can see that apparently the search is returning invalid JSON (see logs).

**To Reproduce**

1. Import activity from [ghostfolio-export.json](https://github.com/user-attachments/files/23436773/ghostfolio-export-202511090928.json)
1. _Gather Profile Data_ for `UETW.DE` via _Admin Control_ panel → Market Data

**Expected Behavior**

Ghostfolio gathers country information from _Trackinsight_.

**Logs**

``` 
[Nest] 1  - 11/08/2025, 11:36:18 AM   ERROR [TrackinsightDataEnhancerService] Failed to search Trackinsight symbol for UETW (Unexpected end of JSON input)
[Nest] 1  - 11/08/2025, 11:36:18 AM   ERROR [TrackinsightDataEnhancerService] Failed to search Trackinsight symbol for UETW (Unexpected end of JSON input)
```

**Environment**

- Ghostfolio Version 2.215.0
- Self-hosted
- Experimental Features disabled
- Browser: Firefox, Chrome
- OS: hosted on Debian, running on Kubernetes

dtslvr:
Thanks for the report. Could you attach the exported activity to the issue description to simplify reproduction?

pleich:
Of course, [here](https://github.com/user-attachments/files/23436773/ghostfolio-export-202511090928.json) is a minimal example

inevs:
@dtslvr I can do this fix

inevs:
Trackinsight protected their API with AWS WAF. The requests are blocked because WAF sends a JavaScript Challenge (simple headers are insufficient)

omkarg01:
is this issue still open?

rejuce:
i hope for fix soon as well...seeing the geo distribution of my allocations was one of the most important features to me aside from sector distribution

MichalskiM:
I took a look into this because for me it's one of the core features of using ghostfolio (I'm just starting).
As long as trackinsight protects their API with a JS "turnstile" there won't be a way around it except some nasty automated browser stuff like selenium. And I didn't find any free API that would provide the necessary data.
[...]

PR #6357, opened 2026-02-20
fix: trackinsight data gathering

Pablito2020:
This PR serves as a proof of concept for fixing the TrackInsight request issue. It incorporates two alternative approaches for making the requests: Puppeteer and curl-impersonate. Each option comes with its own trade-offs.

### Option 1: Using Puppeteer

The pros of this solution is that it works reliably in all cases, the problem is that it requires running chromium, which:
 - Significantly increases the Docker image size.
 - Makes requests slower compared to a lightweight HTTP client.

### Option 2: Using curl-impersonate

The pros of this solution is that is lightweight and keeps the Docker image smaller. But the cons are that some requests can still be blocked by AWS WAF if too many requests originate from the same IP (when triggered, AWS WAF presents a javascript challenge, which of course we can't do, so the recovery requires waiting for a timeout or rotating your ip address...). 
Also, requires downloading curl-impersonate from external sources, which makes packaging more complex for restrictive package managers like for example nix.

If we were to choose one approach, I would prefer the puppeteer one, since it is reliable the 100% of the time and makes live easier for my package manager of choice.


Pablito2020:
Importing the geographical data just from funds isn't supported. You need to track the data from insight via symbol mapping on the fund (mapping your fund to a similar ETF), for example see:
image

You can read more here: https://github.com/ghostfolio/ghostfolio/discussions/4192

slenderpeppr:
Thank you so much!

Pablito2020:
Hi @dtslvr . Gentle ping 🙂

This PR addresses #5924, which is pretty annoying on the self-hosted version.
Whenever you get a chance, do you think it makes sense to merge this upstream? Any drawbacks you see?

dtslvr:
Thanks for the proof of concept, @Pablito2020.

While I appreciate the effort to fix the _TrackInsight_ integration, I am hesitant to integrate _Puppeteer_ into the core repository. It adds significant weight and increases the complexity for a single data source.

Instead of integrating heavy dependencies, I would prefer an architecture where:

- a dedicated API endpoint allows external updates to asset profiles
- specific scrapers are extracted as community projects which interacts with that API

This would keep the core lightweight while giving the community the flexibility to use their tools of choice. What are your thoughts on this approach?

Pablito2020:
Thanks @dtslvr that approach looks good to me 👍. I just don't yet know if it will be a lot of work to implement or how it fits with future plans, so I'd love to get a sense of the expected effort or whether this is something envisioned for future releases.

dtslvr:
Superseded by #6981.

Issue #6393, opened 2026-02-24
Reuse value component in tag component of Admin Control panel


dtslvr:
The goal of this issue is to reuse the [value component](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/value) in the [tag component](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-tag) of the _Admin Control_ panel (see `activityCount`).

### Acceptance Criteria

* Ensure the visual alignment remains consistent with the existing layout
* Verify that the table sorting functionality is preserved

DarkwinngDuck:
Should value component be used only for `activityCount`?

dtslvr:
> Should value component be used only for `activityCount`?

Yes, exactly.

PR #6467, opened 2026-03-02
Task/reuse value component in tag component of Admin Control panel

DarkwinngDuck:
- Alignment remains consistent with the existing layout
- Table sorting functionality is preserved

 Closes https://github.com/ghostfolio/ghostfolio/issues/6393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-- Reused the value component in the admin tag component of the admin control panel
-
 ### Changed
 
+- Reused the value component in the tag management of the admin control panel
 - Upgraded `jsonpath` from version `1.1.1` to `1.2.1`
 
 ## 2.247.0 - 2026-03-04


apps/client/src/app/components/admin-tag/admin-tag.component.html
```diff
@@ -45,7 +45,10 @@
       <ng-container i18n>Activities</ng-container>
     </th>
     <td *matCellDef="let element" class="px-1 text-right" mat-cell>
-      {{ element.activityCount }}
+      <gf-value
+        class="d-inline-block justify-content-end"
+        [value]="element.activityCount"
```
dtslvr, line 51: Could you provide `[locale]="locale"` and add

```
@Input() locale = getLocale();
```

to `GfAdminTagComponent`?

DarkwinngDuck: Thanks! Added.

DarkwinngDuck:
Thanks for your review, @dtslvr. Could you tell, please, what would you recommend to do next time: 

- make new commit with changes after the code review?
- amend initial commit including changes after the code review?

Two times I have used the second option, but I am not sure if this is a good practice. Thanks!

dtslvr (approved):
Thanks for the update, @DarkwinngDuck! 👍🏻

I would recommend new commits for review changes. Force pushing (amend) makes it harder to track what specifically changed since the last look.

Ready to merge now.

PR #6471, opened 2026-03-03
Bugfix/disabled buttons in assistant component

dtslvr:


dtslvr:
Does this fix make sense to you, @KenTandrian?

PR #6497, opened 2026-03-08
Task/improve type safety in tags selector component

KenTandrian:

### Changes

* Resolved TypeScript compilation errors in tags selector component.
* Migrated `tagInput` to a view child signal.
* Removed `unsubscribeSubject` and implemented `takeUntilDestroyed()`.


KenTandrian:
Thanks for pointing that out, @dtslvr! 

I have reverted the change to use `id: undefined` for new tags. By using a union of `Tag | NewTag`, the component is now type-safe while correctly emitting `undefined` for new tags. This ensures the application property identifies and creates new tags on the backend before attempting to link them.

PR #6551, opened 2026-03-13
Task/eliminate OnDestroy lifecycle hook from home watchlist component

Erwin-N:


apps/client/src/app/components/home-market/home-market.component.ts
```diff
@@ -47,19 +46,18 @@ export class GfHomeMarketComponent implements OnDestroy, OnInit {
...
-
   public constructor(
     private changeDetectorRef: ChangeDetectorRef,
     private dataService: DataService,
+    private destroyRef: DestroyRef,
     private deviceService: DeviceDetectorService,
     private userService: UserService
   ) {
-    this.deviceType = this.deviceService.getDeviceInfo().deviceType;
+    this.deviceType = this.deviceService.deviceInfo().deviceType;
```
dtslvr: Please revert this change

dtslvr (changes requested):

apps/client/src/app/components/home-market/home-market.component.ts
Erwin-N: I'll revert the change, but the getDeviceInfo seems to be deprecated.

PR #6553, opened 2026-03-13
Task/eliminate OnDestroy lifecycle hook from admin platform component

Erwin-N:


apps/client/src/app/components/admin-platform/admin-platform.component.ts
```diff
@@ -12,10 +12,11 @@ import {
   ChangeDetectorRef,
   Component,
   Input,
-  OnDestroy,
+  DestroyRef,
```
dtslvr, line 14: Move up ⬆️

dtslvr (changes requested):

PR #6557, opened 2026-03-14
Task/improve type safety of admin and data services

KenTandrian:

### Changes

* Resolved type errors in admin service and data service.
* Implemented inject function to replace constructor-based injection.
* Reduced usage of `any` type in data service.

### DIFF
--- a/.github/workflows/build-code.yml
+++ b/.github/workflows/build-code.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         node_version:
-          - 22
+          - 22.22.1
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Consolidated the sign-out logic within the user service to unify cookie, state and token clearance
 - Improved the language localization for Polish (`pl`)
+- Upgraded `@ionic/angular` from version `8.7.3` to `8.8.1`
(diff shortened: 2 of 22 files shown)


libs/ui/src/lib/services/admin.service.ts
```diff
@@ -153,7 +151,7 @@ export class AdminService {
     skip,
     take = DEFAULT_PAGE_SIZE
   }: {
-    skip?: number;
+    skip: number;
```
dtslvr: `skip` is still optional, no? (https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/admin/admin.controller.ts#L322)

KenTandrian: yeah, my bad. Should be optional.

libs/ui/src/lib/services/admin.service.ts
```diff
@@ -158,7 +156,10 @@ export class AdminService {
   }) {
     let params = new HttpParams();
 
-    params = params.append('skip', skip);
+    if (skip) {
```
dtslvr: Should we better check for `isNumber()`? 🤔

libs/ui/src/lib/services/admin.service.ts
```diff
@@ -158,7 +156,10 @@ export class AdminService {
...
+      params = params.append('skip', skip);
+    }
+
     params = params.append('take', take);
```
dtslvr: What about here? I would expect the same.

libs/ui/src/lib/services/admin.service.ts
KenTandrian: That makes sense. `isNumber` would be better in handling 0.

PR #6559, opened 2026-03-14
Task/add destroyRef explicitly to lib components

dtslvr:


KenTandrian:
hi @dtslvr, just a few thoughts on this.

As far as I know, `destroyRef` can be omitted when calling `takeUntilDestroyed` in an [injection context](https://angular.dev/guide/di/dependency-injection-context), typically the `constructor` of a component or directive. A `DestroyRef` should be provided if the code calls `takeUntilDestroyed` outside of an injection context.

Docs: https://angular.dev/ecosystem/rxjs-interop/take-until-destroyed

dtslvr:
Thank you for the feedback, @KenTandrian.

You are correct about the injection context. My preference is to consistently provide `destroyRef` to avoid potential issues if the code is ever refactored outside the constructor. From what I have seen, there is no technical downside to being explicit and it is a safer default to prevent leaks.

What do you think about that approach?

KenTandrian:
I see, that makes sense. There is no technical downside indeed. Thanks for clarifying, @dtslvr!

PR #6581, opened 2026-03-17
fix(manual-scraper): respect HTTP_PROXY env vars

### LINKED ISSUE #6206: Respect HTTP_PROXY env by the manual scraper
I'm setting up some HTTP_PROXY at container level. It is respected to query regular assets (from yahoo for instnace), but the "manual" scraper seems to not respect it (other websites)

1. Set HTTP_PROXY, HTTPS_PROXY, NO_PROXY
2. Query an update
3. Notice the traffic avoiding the proxy 

Selfhosted

mvanhorn:
## Summary
- The manual web scraper's `fetch` calls now respect `HTTP_PROXY`, `HTTPS_PROXY` (and their lowercase variants) by using undici's `ProxyAgent` as the fetch dispatcher
- No new dependencies required - `undici` ships with Node.js >= 22 (the project's minimum version)
- Other data providers (e.g. Yahoo) already respect proxy env vars; this brings the manual scraper in line

## Changes
- `apps/api/src/services/data-provider/manual/manual.service.ts`: import `ProxyAgent` from `undici`, read proxy URL from environment, and pass `dispatcher` option to `fetch` when a proxy is configured

Closes #6206

### DIFF
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -18,9 +18,15 @@ import type { NestExpressApplication } from '@nestjs/platform-express';
 import cookieParser from 'cookie-parser';
 import { NextFunction, Request, Response } from 'express';
 import helmet from 'helmet';
+import { EnvHttpProxyAgent, setGlobalDispatcher } from 'undici';
 
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
+
+// Route Node's global `fetch` (used by all data-provider services) through
+// HTTP_PROXY / HTTPS_PROXY when those env vars are set. Native fetch
+// otherwise ignores them. EnvHttpProxyAgent reads NO_PROXY natively.
+setGlobalDispatcher(new EnvHttpProxyAgent());
 
 async function bootstrap() {
   const configApp = await NestFactory.create(AppModule);
--- a/apps/api/src/services/data-provider/manual/manual.service.ts
+++ b/apps/api/src/services/data-provider/manual/manual.service.ts
@@ -27,6 +27,7 @@ import { Injectable, Logger } from '@nestjs/common';
 import { DataSource, SymbolProfile } from '@prisma/client';
 import * as cheerio from 'cheerio';
 import { addDays, format, isBefore } from 'date-fns';
+import { ProxyAgent } from 'undici';
 
 @Injectable()
 export class ManualService implements DataProviderInterface {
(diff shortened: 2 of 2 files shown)


Booyaka101:
Thanks for tackling this, @mvanhorn — and apologies for chiming in unprompted, just wanted to surface something from the original issue thread in case it's useful for getting the review unstuck.

In [#6206](https://github.com/ghostfolio/ghostfolio/issues/6206), @dtslvr specifically suggested the `setGlobalDispatcher` + `EnvHttpProxyAgent` approach:

> Integrating `undici` to handle the environment variables might be the cleanest approach. Something like this:
> ```ts
> import { setGlobalDispatcher, EnvHttpProxyAgent } from 'undici';
> setGlobalDispatcher(new EnvHttpProxyAgent());
> ```

Three things that switch would pick up that the current diff doesn't:

1. **Coverage of other data providers.** The same native-`fetch`-ignores-proxy issue exists in `coingecko.service.ts`, `openfigi.service.ts`, `trackinsight.service.ts`, `eod-historical-data.service.ts`, `financial-modeling-prep.service.ts`, `ghostfolio.service.ts`, and `rapid-api.service.ts` — all use native `fetch`. A global dispatcher set once in `apps/api/src/main.ts` proxies all of them. @Rqirus's note in the issue about Yahoo being unreachable from mainland China would fall under this umbrella too.

2. **`NO_PROXY` handling.** `EnvHttpProxyAgent` reads `NO_PROXY` natively. The current `new ProxyAgent(proxyUrl)` path proxies all URLs whenever `HTTPS_PROXY`/`HTTP_PROXY` is set, including ones the user explicitly excluded.

3. **Allocation cost.** `new ProxyAgent(proxyUrl)` is built on every scrape call rather than once at bootstrap. Minor, but free to fix when moving to global.

Practically, this would shrink the diff in `manual.service.ts` back to zero and add a couple of lines near the top of `main.ts`:

```ts
import { setGlobalDispatcher, EnvHttpProxyAgent } from 'undici';
// ...early in bootstrap(), before any data-provider call
setGlobalDispatcher(new EnvHttpProxyAgent());
```

Completely your call — if the narrower scope was intentional, please disregard. Just figured the maintainer's specific suggestion was worth flagging here on the PR alongside the diff so the review conversation has the full context.

mvanhorn:
Thanks @Booyaka101 - good catch. Pushed f961ab350 switching to setGlobalDispatcher(new EnvHttpProxyAgent()) in apps/api/src/main.ts and reverting the inline ProxyAgent in manual.service.ts. Covers the seven other services that hit native fetch and picks up NO_PROXY natively. @dtslvr - this matches what you suggested in #6206; happy to update the PR title/description if you'd like.

dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/pull/6952.

Issue #6617, opened 2026-03-21
Eliminate OnDestroy lifecycle hook from import activities dialog component


dtslvr:
The goal of this issue is to remove the `OnDestroy` lifecycle hook from the [import activities dialog component](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/pages/portfolio/activities/import-activities-dialog).

Follow the implementation used in PR #6419 as a reference.

Erwin-N:
Hi @dtslvr
You can assign this to me

PR #6621, opened 2026-03-21
Task/eliminate OnDestroy lifecycle hook from import activities dialog component

Erwin-N:
#6617


dtslvr (approved):
Thanks for your contribution, @Erwin-N! 😎

PR #6630, opened 2026-03-24
Task/improve language translation for ES

jhernaezayuso:
**Changed:**
- Complete all missing spanish translations
- Correct inaccurate translations
- Ensured terminology consistency across the entire application

Related issue: #3588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


jhernaezayuso:
@dtslvr 
Thanks for the review!
I don't know what I touched before uploading the changes that broke the format. It's already solved. I have added the changelog. Anything else that needs to be changed, let me know

There are anyways several lines affected since I have also re-translated some things that were inaccurate

Have you thought about changing the translation system to something like Crowdin? 
I think it would allow many more users to contribute more easily without so much fuss about doing a PR, I think it would make many more people contribute and add new languages

dtslvr:
> Have you thought about changing the translation system to something like Crowdin?

We considered once to integrate _Crowdin_, but we prefer to keep the translation process as simple as possible for now.

PR #6705, opened 2026-04-09
Bugfix/implement text-nowrap for activity type badge

KenTandrian:
Hi @dtslvr, this PR fixes an issue on activity type badge when the text has more than 1 word and is wrapped into 2 lines.

I added `text-nowrap` class so that the text is not wrapped.

### Before
Screenshot 2026-04-09 at 11 36 28

### After
Screenshot 2026-04-09 at 11 36 58


dtslvr (approved):
Thanks for the improvement, @KenTandrian!

PR #6710, opened 2026-04-09
Fix import activities dialog overflow for long notice lists

### LINKED ISSUE #6600: [BUG] To many notices in "import activities" modal make modal action buttons inaccessible
**Bug Description**

When "import activities" produces too many notices on screen the modal action buttons (cancel / import) are inaccessible. It is impossible to scroll to the buttons.

**To Reproduce**

Import a JSON or CSV file that produces many notices and the modal dialog will expand into a non-reachable area off screen.

**Expected Behavior**

The modal dialog should be max 100% screen height with a title on top, action buttons on the bottom and the content area with an inline vertical scrollbar.

Alternatively the whole page could be scrollable to the very end to reach the modal buttons at the bottom.

**Screenshots**

Image

**Environment**

- Ghostfolio Version 2.250.0 - 2026-03-17
- Self-hosted (Ubuntu Server 24.04 / docker compose)
- Browser: Vivaldi (latest release)
- OS: Ubuntu 25.10

**Additional info**

I am aware that an input file should not produce that many notices. I'll fix my import script to produce more valid data. Still think the modal import modal should be more robust and be able to handle this edge case.

officialasishkumar:
## Summary

Fixes #6600

Keep the import activities dialog content inside the viewport when an import produces many notices. The dialog content area now scrolls instead of allowing the modal to grow past the screen height.

## Changes

- cap the import dialog content height relative to the viewport
- enable vertical scrolling inside the dialog content area
- add a changelog entry

## Test plan

- [x] `npx nx lint client --quiet`
- [x] `npx prettier --check CHANGELOG.md apps/client/src/app/pages/portfolio/activities/import-activities-dialog/import-activities-dialog.scss`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed the import activities dialog so long notice lists remain scrollable within the viewport
 - Improved the style of the activity type component
 
 ## 2.253.0 - 2026-03-06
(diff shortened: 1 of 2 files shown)


dtslvr:
Superseded by #7047.

PR #6722, opened 2026-04-11
Task/improve type safety in asset profile dialog

KenTandrian:

## Changes

- Resolved type errors in the component and template files.
- Scoped properties with `protected`, `private`, and `readonly` access modifiers.
- Replaced untyped `.get('propertyName')` string lookups with strongly-typed `.controls.propertyName` dot-notation to explicitly adhere to Angular 14+ Typed Forms standards.
- Explicitly typed `JSON.parse` blocks as `Prisma.InputJsonArray` and `Record<string, string>`, and introduced strict `?? undefined` fallbacks to satisfy `UpdateAssetProfileDto` nullability constraints.
- Applied explicit `HttpErrorResponse` typings to `catchError` and API `.subscribe()` error handlers. 

### Why we migrated to `.controls` syntax:

Historically, reading Angular form controls required the string-based `.get('propertyName')` method. Because this string could dynamically map to anything, the compiler could not protect you from simple typos, causing silent runtime failures:

```typescript
// The issue here is:
this.assetProfileForm.get('assetSubClass')    // correct, but throws `Object is possibly 'null'.`!
this.assetProfileForm.get('assetSubClassx')   // does not throw an error at compile time!

// By upgrading to dot-notation on Angular 14+ Typed Forms:
this.assetProfileForm.controls.assetSubClass    // correct!
this.assetProfileForm.controls.assetSubClassx   // throws a strict compilation error!
```

By leveraging the strongly-typed `.controls` property, the compiler now statically guarantees that the field exists and strictly enforces the exact shape of the DTO payload. This completely eliminates silent assignment bugs and prevents malformed payloads from ever crossing the network boundary.

## Resolved type errors

49 errors (before: 226, after: 177).


dtslvr (approved):

PR #6731, opened 2026-04-12
Task/improve type safety of account detail dialog

KenTandrian:

## Changes

- Migrated constructor-based dependency injection to inject function.
- Tighten up encapsulation and immutability.
- Removed unused `activities` variable.
- Resolved type errors.

## Resolved type errors

5 errors (before: 231, after: 226).

### DIFF
--- a/apps/client/src/app/components/account-detail-dialog/account-detail-dialog.component.ts
+++ b/apps/client/src/app/components/account-detail-dialog/account-detail-dialog.component.ts
@@ -27,7 +27,7 @@ import {
   ChangeDetectorRef,
   Component,
   DestroyRef,
-  Inject,
+  inject,
   OnInit
 } from '@angular/core';
 import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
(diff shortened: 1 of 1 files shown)


apps/client/src/app/components/account-detail-dialog/account-detail-dialog.component.ts
```diff
@@ -330,7 +330,10 @@ export class GfAccountDetailDialogComponent implements OnInit {
       next: ({ accountBalances, portfolioPerformance }) => {
         this.accountBalances = accountBalances.balances;
 
-        if (portfolioPerformance.chart.length > 0) {
+        if (
+          portfolioPerformance.chart &&
+          portfolioPerformance.chart.length > 0
```
dtslvr, line 335: Is `portfolioPerformance.chart?.length > 0` not the same?

apps/client/src/app/components/account-detail-dialog/account-detail-dialog.component.ts
```diff
@@ -78,41 +77,42 @@ import { AccountDetailDialogParams } from './interfaces/interfaces';
   templateUrl: 'account-detail-dialog.html'
 })
 export class GfAccountDetailDialogComponent implements OnInit {
-  public accountBalances: AccountBalancesResponse['balances'];
-  public activities: OrderWithAccount[];
```
dtslvr, line 82: 👍🏻

dtslvr (changes requested):

apps/client/src/app/components/account-detail-dialog/account-detail-dialog.component.ts
KenTandrian: No, because if `portfolioPerformance.chart` can be `undefined`, then `portfolioPerformance.chart.length` can also be `undefined`, and `undefined` cannot be compared with a number (e.g., `0`).

This is what happen if we do that:

Screenshot 2026-04-13 at 21 54 46

dtslvr: Makes sense, thanks for the explanation! 👍🏻

dtslvr (approved):

PR #6704, opened 2026-04-09
Task/upgrade to Nx 22.6

DavidReque:
Hi @dtslvr, this PR resolves #6700 

### Summary
Upgrades `Nx 22.5.3 → 22.6.4` and bumps the Angular 21.2.x toolchain so the workspace stays on the supported Nx/Angular stack. Includes small migration follow-ups (UI + Nx config) and fixes api:lint by removing a disallowed relative `package.json` import in production env.

### Changes
- `package.json` — Nx + Angular-related versions updated and lockfile refreshed.
- `libs/ui/.../holdings-table.component.ts` — Dropped redundant `CommonModule` from standalone imports. Why: Post–control-flow migration cleanup; avoids an unnecessary dependency.
- `apps/api/src/environments/environment.prod.ts` — Version from `process.env.npm_package_version ?? 'dev'` instead of require(`'../../../../package.json'`).
Why: Satisfies `@nx/enforce-module-boundaries` and unblocks `api:lint`.
- `nx.json — "analytics": false`. Why: Nx 22.6 migration/config normalization.
- `.gitignore` — Ignore `.claude/worktrees` and `.claude/settings.local.json`.
Why: Nx 22.6 migration; keeps local Claude files out of git.
- `CHANGELOG.md` — Unreleased “Changed” bullets for the upgrade and env fix.

PR #6742, opened 2026-04-16
Bugfix/resolve admin version fallback in production

### LINKED ISSUE #6726: Version is showing as "dev" in Admin page
**Bug Description**

The version number, which should be "2.254.0", is now shown as "dev".

**To Reproduce**

1. Open the Ghostfolio app.
2. Go to Admin page (`/en/admin`).
3. See the version number.

**Expected Behavior**

It should be shown as "2.254.0"

**Screenshots**

Image

**Environment**

- Ghostfolio Version 2.254.0
- Self-hosted
- Chrome Browser 146.0.7680.178
- macOS Tahoe 26.4.1

**Additional Context**

This might be related to the change in #6704 on this line:

https://github.com/ghostfolio/ghostfolio/blob/ea620ee76b48f14c1928d4a8f7d698dc04afdd9f/apps/api/src/environments/environment.prod.ts#L6

DavidReque:
Hi @dtslvr, this PR resolves #6726, please take a look

### Summary
Fix the version displayed on the Admin page in production.

The API no longer relies only on `npm_package_version`, which may be unavailable in some production environments. It now resolves the version from `APP_VERSION`, `npm_package_version`, or the runtime `package.json` before falling back to `dev`.

### Changes
- Updated the production API environment to resolve the app version more reliably.
- Added fallbacks for `APP_VERSION`, `npm_package_version`, and the runtime `package.json`.
- Prevented the Admin page from incorrectly showing `dev` in self-hosted production setups.

### DIFF
--- a/apps/api/src/environments/environment.prod.ts
+++ b/apps/api/src/environments/environment.prod.ts
@@ -1,7 +1,30 @@
 import { DEFAULT_HOST, DEFAULT_PORT } from '@ghostfolio/common/config';
 
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+const getVersion = () => {
+  if (process.env.APP_VERSION) {
+    return process.env.APP_VERSION;
+  }
+
+  if (process.env.npm_package_version) {
+    return process.env.npm_package_version;
+  }
+
+  try {
+    const packageJson = JSON.parse(
+      readFileSync(join(process.cwd(), 'package.json'), 'utf8')
+    );
+
+    return packageJson.version ?? 'dev';
+  } catch {
+    return 'dev';
+  }
+};
+
 export const environment = {
   production: true,
   rootUrl: `http://${DEFAULT_HOST}:${DEFAULT_PORT}`,
-  version: process.env.npm_package_version ?? 'dev'
+  version: getVersion()
 };


apps/api/src/environments/environment.prod.ts
dtslvr, line 1: What is the reason we can't keep using the original code instead of this new implementation? https://github.com/ghostfolio/ghostfolio/pull/6704/changes#diff-198f93e70575715d3493bc9c5d828322a68c201884c912d210dadc80728388e9L6

DavidReque: The reason we can't keep the original implementation is that `require('../../../../package.json')` was the part that started failing `api:lint` after the Nx upgrade because of `@nx/enforce-module-boundaries`.

So this change is not really introducing a new source of truth. It still resolves the version from `package.json`, but without importing it through Nx's module graph. Reading it at runtime avoids the lint violation while preserving the same behavior in production.

Using only `process.env.npm_package_version` turned out to be unreliable because that variable is not guaranteed to be present in all production startup paths (for example when the app is launched with `node` rather than through an npm script), which is why Admin started showing `dev`.

dtslvr (approved):
Thanks for the contribution, @DavidReque! 😎

PR #6800, opened 2026-04-29
Task/reset background color of paginators

### LINKED ISSUE #6799: Reset background color of paginators
The goal of this issue is to reset the background color of the [paginator](https://material.angular.dev/components/paginator/styling) container (transparent).

Image

Remove all outdated `.mat-mdc-paginator` CSS overrides as well.

mvanhorn:
Resolves #6799.

Removes the two `.mat-mdc-paginator` CSS overrides in `apps/client/src/styles.scss`:

- The dark-theme block setting the background to `rgba(var(--palette-foreground-base-dark), 0.02)`
- The global block setting the background to `rgba(var(--palette-foreground-base-light), 0.02)` plus the `.mat-mdc-paginator-page-size { display: none }` rule (the selector was duplicated in the source)

The default Angular Material paginator background is transparent and follows the surface theme, which matches the issue's request. Components that need to hide the page-size selector should use the paginator's `[hidePageSize]` input rather than a CSS override.

Diff: 13 lines removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Modernized the layout of the overview tab in the admin control panel
+- Improved the styling of the paginator across various table components
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


dtslvr:
Please comment in #6799 so we can assign the issue to you, @mvanhorn.

dtslvr (approved):
Thank you, @mvanhorn.

PR #6807, opened 2026-05-01
Task/improve type safety in fire page component

KenTandrian:
Hi @dtslvr, this PR is related to #6246. Please take a look :)

## Changes

- Resolved type errors.
- Replaced deprecated `getDeviceInfo()` with `deviceInfo` signal.
- Tightened up encapsulation and immutability.
- Replaced constructor-based dependency injection with `inject` function.

## Resolved type errors

6 errors (before: 150, after: 144).

PR #6831, opened 2026-05-02
Bugfix/reactivate auto calculate for retirement date and projected total amount in FIRE page

KenTandrian:
Hi @dtslvr, this PR is a bugfix for #6807. Please take a look :)

## Changes

- Changed `projectedTotalAmount` and `retirementDate` back to `null`.
- Updated `UpdateUserSettingDto` to support `null` type for both variables
- Added JSDoc documentation about the `null` behavior.

## Additional Context

Setting one of the fields as `null` is necessary because of the conditional statement to delete the field from the user settings, which is triggered when the field is `null` but not `undefined`:

https://github.com/ghostfolio/ghostfolio/blob/a04099b12d23e2cbf71274aadd2868ff6826cd32/apps/api/src/app/user/user.controller.ts#L183-L185

Therefore, we should modify the DTO to support `null` instead of changing the input to `undefined`.

### DIFF
--- a/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
+++ b/apps/client/src/app/pages/portfolio/fire/fire-page.component.ts
@@ -174,7 +174,7 @@ export class GfFirePageComponent implements OnInit {
     this.dataService
       .putUserSetting({
         projectedTotalAmount,
-        retirementDate: undefined
+        retirementDate: null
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thank you very much for the fix, @KenTandrian! 👍🏻

PR #6833, opened 2026-05-03
Bugfix/improve typography for client app and charts

KenTandrian:
Hi @dtslvr, this is a bugfix for typography related to Material 3 upgrade in #6768.

## Changes

* Bring back typography config to `theme.scss`.
* Configure default font family for charts.

## Context

This change is made after I discovered that the fonts are mostly using "Roboto" instead of using our preconfigured CSS variable, which should point to "Inter" font.

Screenshot 2026-05-03 at 11 16 05

###

However, as the method `mat.m2-define-typography-config()` is now deprecated, we should configure the typography config map manually following the guide here: https://material.angular.dev/guide/theming#typography-map.


dtslvr (approved):
Great catch, @KenTandrian! This was a subtle one. Thanks for providing a PR.

PR #6845, opened 2026-05-05
fix(activities): preserve existing symbol when cloning a manual asset activity 

### LINKED ISSUE #5324: [BUG] Cloning an activity that uses a Manual asset creates a new Manual asset
**Bug Description**

 Cloning an activity that uses a Manual asset creates a new Manual asset instead of adding the activity to the existing asset.

**To Reproduce**

1. Have a manual asset (see screenshot 1)
2. In the Portofio->activities page, press the ... button and then Clone
3. Enter some value in unit price and Save  (see screenshot 2)

**Expected Behavior**

The cloned activity should belong to the manual asset

**Incorrect Behavior**

The cloned activity is added to a new manual asset (that didn't exist before) (See screenshot 3)

**Screenshots**

Screenshot 1

Image

Screenshot 2

Image

Screenshot 3

Image

**Environment**

- Ghostfolio Version  2.188.0 
- Self-hosted
- Firefox 141

BenHolmes:
fix: preserve existing symbol when cloning a manual asset activity

Update create-or-update-activity-dialog.component.ts

When cloning an activity linked to a manual asset (VALUABLE, FEE, INTEREST, LIABILITY), the symbol field in onSubmit() was explicitly set to undefined for these types, causing the backend to generate a new UUID and create a duplicate SymbolProfile instead of reusing the existing one.

Simplify the symbol assignment to always prefer searchSymbol.value.symbol (which correctly holds the existing asset's identifier when cloning) and fall back to name.value only when no symbol exists (i.e. when creating a brand-new manual asset).

Fixes #5324

### DIFF
--- a/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.component.ts
+++ b/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.component.ts
@@ -500,11 +500,7 @@ export class GfCreateOrUpdateActivityDialogComponent {
       fee: this.activityForm.get('fee')?.value,
       quantity: this.activityForm.get('quantity')?.value,
       symbol:
-        (['FEE', 'INTEREST', 'LIABILITY', 'VALUABLE'].includes(
-          this.activityForm.get('type')?.value
-        )
-          ? undefined
-          : this.activityForm.get('searchSymbol')?.value?.symbol) ??
+        this.activityForm.get('searchSymbol')?.value?.symbol ??
         this.activityForm.get('name')?.value,
       tags: this.activityForm.get('tags')?.value?.map(({ id }) => {
         return id;


BenHolmes:
**Summary**

Fixes #5324 — when cloning an activity linked to a manual asset (types `VALUABLE`, `FEE`, `INTEREST`, `LIABILITY`), the cloned activity was incorrectly creating a brand-new duplicate asset instead of linking to the existing one.

**Root cause**

In `onSubmit()`, the `symbol` field was explicitly set to `undefined` for MANUAL-type activities, causing the backend's `connectOrCreate` to never find the existing `SymbolProfile`. Instead it generated a fresh UUID, creating a duplicate asset on every clone.

**Fix**

Simplified the `symbol` assignment to always prefer `searchSymbol.value.symbol` first (which correctly holds the existing asset's UUID when cloning), falling back to `name.value` only when no symbol exists — i.e. when genuinely creating a brand-new manual asset. This single-line change handles both cases correctly without any behaviour change for new asset creation.

**Temporary workaround for self-hosted users (until this is merged)**

For anyone affected by this bug, duplicate activities and orphaned SymbolProfiles can be cleaned up directly in the database. First find the correct `symbolProfileId` for the existing asset by querying an existing activity for that asset via `GET /api/v1/order`, then:

```sql
-- 1. Remove the incorrectly created activity
DELETE FROM "Order" WHERE id = '<duplicate-activity-id>';

-- 2. Remove the orphaned duplicate SymbolProfile
DELETE FROM "SymbolProfile" WHERE id = '<duplicate-symbol-profile-id>';

-- 3. Re-insert the activity linked to the correct SymbolProfile
INSERT INTO "Order" (id, "accountId", "userId", "symbolProfileId", type, date, quantity, "unitPrice", fee, currency, "isDraft", "createdAt", "updatedAt")
VALUES (
  gen_random_uuid(),
  '<account-id>',
  '<user-id>',
  '<correct-symbol-profile-id>',
  'BUY',
  '<date>',
  <quantity>,
  <unit-price>,
  0,
  '<currency>',
  false,
  now(),
  now()
);
```

The correct `symbolProfileId` can be found by inspecting any existing activity for that asset in the `GET /api/v1/order` response and reading its `symbolProfileId` field.

**Testing**

1. Create a manual asset activity (e.g. type `VALUABLE`)
2. Clone it via the ⋯ menu → Clone, change the unit price, and save
3. Confirm the cloned activity links to the original asset — no duplicate asset is created

**Requesting review** — happy to adjust anything based on feedback. Thanks!

joeshmoe57:
Any chance at this being merged? Still impacting me (and I assume others via #5324 ) and this seems like a small change to address it while a longer term solution is devised. (the manual db migrations listed in #5324 seem like a poor expectation)

PR #6843, opened 2026-05-05
Task/migrate Angular components from NgClass to class bindings

DavidReque:
Hi @dtslvr, this PR resolves #6815, please take a look

### Summary
- Replace `NgClass` with native class bindings in client and `ui` templates (Angular `ngclass-to-class` migration + small manual fixes).
- Update `CHANGELOG.md`

PR #6848, opened 2026-05-09
Bugfix/resolve styling issues by splitting space-separated class bindings

KenTandrian:
Hi @dtslvr, this is a bugfix for missing styles on certain UI components (such as the `large` size text in `gf-value` and styling in the `historical-market-data-editor`) following the `ngClass` to `class` migration in #6843.

## Changes

* Split space-separated string keys in `[class]` bindings into individual `[class.classname]="condition"` bindings across:
  * `libs/ui/src/lib/value/value.component.html`
  * `libs/ui/src/lib/historical-market-data-editor/historical-market-data-editor.component.html`

## Context

In modern Angular, the native `[class]` object syntax does not support multiple space-separated classes within a single key (e.g., `'font-weight-bold h2': condition`). Under the hood, Angular uses `classList.toggle()`, which strictly expects a single class name. Passing spaces causes the classes to fail silently, leaving the elements unstyled.

This resolves the issue by adopting the official Angular recommended migration pattern, which breaks the multi-class keys down into individual `[class.x]` bindings for cleaner evaluation and performance.

Reference: [Angular ngClass to class Migration Guide](https://angular.dev/reference/migrations/ngclass-to-class#--migrate-space-separated-key)

### Before

Screenshot 2026-05-09 at 09 32 43

### After

Screenshot 2026-05-09 at 09 32 26


dtslvr (approved):
Thanks for providing this fix, @KenTandrian! 👍🏻

PR #6864, opened 2026-05-15
fix: set default activity query limit to prevent Prisma transaction timeout with large portfolios

### LINKED ISSUE #6863: [BUG] getActivities() defaults to Number.MAX_SAFE_INTEGER causing Prisma transaction timeout with large portfolios
## Bug Description

The `getActivities()` method in `apps/api/src/app/activities/activities.service.ts` defaults the `take` parameter to `Number.MAX_SAFE_INTEGER`. When users have large portfolios (>10k+ activities), calling this method without an explicit `take` value causes Prisma transaction timeouts (`P2028` error).

## Impact

- The `/api/v1/activities` endpoint (and other endpoints that internally call `getActivities`) returns HTTP 500 when the user has many activities
- The portfolio UI shows "Oops! Something went wrong" and fails to load
- The error is: `Transaction API error: Unable to start a transaction in the given time`

## Root Cause

Line 473 of `apps/api/src/app/activities/activities.service.ts`:
```typescript
take = Number.MAX_SAFE_INTEGER,
```

Internal callers (`deleteActivities`, `getActivitiesForPortfolioCalculator`, portfolio service) don't pass `take`, so the query fetches ALL activities at once. With 48k+ activities including joined symbol profile data and exchange rate calculations, the Prisma transaction exceeds the 5-second timeout.

## Steps to Reproduce

1. Import 20k+ activities into Ghostfolio
2. Visit the portfolio activities page
3. The UI fails to load with "Internal server error"
4. Backend logs show: `PrismaClientKnownRequestError: Transaction API error: Unable to start a transaction in the given time`

## Suggested Fix

Change the default `take` value from `Number.MAX_SAFE_INTEGER` to a reasonable finite default, and update internal callers to pass explicit limits where they need all data.

### Changes needed:

1. **`apps/api/src/app/activities/activities.service.ts` line 473:**
   ```diff
   -    take = Number.MAX_SAFE_INTEGER,
   +    take = 100,
   ```

2. **Internal callers that need all data** should pass an explicit high limit:
   - `deleteActivities()`
   - `getActivitiesForPortfolioCalculator()`
   - `portfolio.service.ts`
   - `export.service.ts` (exporting all activities)
   - `import.service.ts` (checking for duplicates during imports against all activities)

arogan178:
## Description

The `getActivities()` method in `ActivitiesService` defaults the `take` parameter to `Number.MAX_SAFE_INTEGER`, which causes unbounded queries on the `Order` table. When users have large portfolios (>10k activities), this leads to Prisma transaction timeouts (`P2028`) and the portfolio UI fails to load with "Internal server error".

## Changes

1. **`activities.service.ts`**: Changed the default `take` from `Number.MAX_SAFE_INTEGER` to `100`. This ensures that any caller that doesn't explicitly set `take` will only fetch a limited number of records.
2. **`activities.service.ts`**: Added explicit `take: Number.MAX_SAFE_INTEGER` to `deleteActivities()` and `getActivitiesForPortfolioCalculator()` internal callers that genuinely need all matching records.
3. **`portfolio.service.ts`**: Added explicit `take: Number.MAX_SAFE_INTEGER` to the portfolio calculation call that needs all activities.
4. **`portfolio.controller.ts`**: Added explicit `take: Number.MAX_SAFE_INTEGER` to the dividends endpoint that needs all dividend records.
5. **`export.service.ts` and `import.service.ts`**: Added explicit `take: Number.MAX_SAFE_INTEGER` to `getActivities()` calls to ensure duplicate checking and export generation operates against the entire portfolio rather than being artificially limited to 100 records.

## Impact

- HTTP endpoints now default to returning `100` activities instead of fetching all records
- Internal operations (portfolio calculation, bulk delete, duplicate check during import, exporting) continue to work with all records by passing an explicit high limit
- Prevents `Transaction API error: Unable to start a transaction in the given time` for users with large portfolios
- The UI can paginate properly through the data

## Related Issues

Closes #6863

### DIFF
--- a/apps/api/src/app/activities/activities.controller.ts
+++ b/apps/api/src/app/activities/activities.controller.ts
@@ -177,6 +177,7 @@ export class ActivitiesController {
 
     const { activities } = await this.activitiesService.getActivities({
       userCurrency,
+      take: Number.MAX_SAFE_INTEGER,
       includeDrafts: true,
       userId: impersonationUserId || this.request.user.id,
       withExcludedAccountsAndActivities: true
--- a/apps/api/src/app/activities/activities.service.ts
+++ b/apps/api/src/app/activities/activities.service.ts
@@ -288,6 +288,7 @@ export class ActivitiesService {
       filters,
       userId,
       includeDrafts: true,
+      take: Number.MAX_SAFE_INTEGER,
       userCurrency: undefined,
       withExcludedAccountsAndActivities: true
     });
@@ -470,7 +471,7 @@ export class ActivitiesService {
     sortColumn,
     sortDirection = 'asc',
     startDate,
-    take = Number.MAX_SAFE_INTEGER,
+    take = 100,
     types,
     userCurrency,
     userId,
@@ -766,6 +767,7 @@ export class ActivitiesService {
   }) {
     const activities = await this.getActivities({
       filters,
+      take: Number.MAX_SAFE_INTEGER,
       userCurrency,
       userId,
       withExcludedAccountsAndActivities: false // TODO
--- a/apps/api/src/app/export/export.service.ts
+++ b/apps/api/src/app/export/export.service.ts
@@ -46,6 +46,7 @@ export class ExportService {
       includeDrafts: true,
       sortColumn: 'date',
       sortDirection: 'asc',
+      take: Number.MAX_SAFE_INTEGER,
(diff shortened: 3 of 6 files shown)


arogan178:
Superseded by #6901 which implements the Float32Array cache approach preferred by @dtslvr.

PR #6865, opened 2026-05-15
fix: internal getActivities callers missing take parameter

arogan178:
## Summary
* The recent change to add a default `take: 100` limit to `getActivities` missed some internal callers.
* Added `take: Number.MAX_SAFE_INTEGER` to `export.service.ts` to ensure exports include all activities, rather than being truncated to 100.
* Added `take: Number.MAX_SAFE_INTEGER` to `import.service.ts` so duplicate activity checks correctly validate against the entire portfolio instead of just the oldest 100 activities.


arogan178:
Closing as I've merged these changes into #6864 so they can be a single PR.

PR #6886, opened 2026-05-17
Task/improve type safety in admin overview component

KenTandrian:

## Changes

- Set `value` in `UpdatePropertyDto` to accept `undefined` as it is optional.
- Set `aLanguageCode` to accept `undefined` because `this.user?.settings?.language` can be undefined.
- Tightened up encapsulation and immutability.
- Replaced constructor-based dependency injection with `inject` function.

## Resolved type errors

4 errors (before: 133, after: 129).


dtslvr (approved):

Issue #2241, opened 2023-08-15
[BUG] JavaScript heap out of memory error


dlsnet:
I have a clean installation of Ghostfolio in docker. When I add a new item to a portfolio or try to update the existing items, I get following error:

[Nest] 132  - 08/15/2023, 1:27:00 PM     LOG [DataGatheringProcessor (GATHER_HISTORICAL_MARKET_DATA)] Historical market data gathering has been started for AFRM (YAHOO) at 2023-08-07
<--- Last few GCs --->
[132:0x5ec29b0]    93913 ms: Scavenge 978.6 (1000.1) -> 976.7 (1000.1) MB, 1.2 / 0.0 ms  (average mu = 0.315, current mu = 0.292) allocation failure; 
[132:0x5ec29b0]    93925 ms: Scavenge 978.6 (1000.1) -> 976.7 (1000.3) MB, 9.3 / 0.0 ms  (average mu = 0.315, current mu = 0.292) allocation failure; 
[132:0x5ec29b0]    94725 ms: Mark-sweep 978.7 (1000.3) -> 976.6 (1000.3) MB, 796.9 / 0.0 ms  (average mu = 0.201, current mu = 0.068) allocation failure; scavenge might not succeed
<--- JS stacktrace --->
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb83f50 node::Abort() [/usr/local/bin/node]
 2: 0xa94834  [/usr/local/bin/node]
 3: 0xd647c0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 4: 0xd64b67 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: 0xf42265  [/usr/local/bin/node]
 6: 0xf43168 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [/usr/local/bin/node]
 7: 0xf53673  [/usr/local/bin/node]
 8: 0xf544e8 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/usr/local/bin/node]
[...]

PR #6916, opened 2026-05-20
Improve resilience of the portfolio snapshot computation for large portfolios

### LINKED ISSUE #6914: Portfolio calculator stalls (job stalled more than allowable limit) → HTTP 500 on all portfolio/account endpoints with large transaction history
### Issue

`RoaiPortfolioCalculator.initialize()` throws `Error: job stalled more than allowable limit`, which bubbles up as **HTTP 500 on every portfolio/account endpoint** once an account has a large transaction history. A single heavy account takes down the entire user's dashboard, `/api/v1/account`, and `/api/v1/portfolio/*`.

### Environment

- Ghostfolio **3.3.0**, self-hosted via Docker
- `postgres:15-alpine`, `redis:7-alpine`
- ~14,400 activities across 5 accounts (four IRAs ≈ 4,700 + one taxable brokerage ≈ 9,700), ~250 symbols, 12 years of history (2014–2026)
- Container given 24 GB / `--max-old-space-size=20480` (see "Heap" note below — it OOM'd at lower limits)

### Symptom

`GET /api/v1/portfolio/holdings`, `/api/v1/portfolio/details`, and `/api/v1/account` all return `500 Internal Server Error`. Stack trace:

```
ERROR [ExceptionsHandler] Error: job stalled more than allowable limit
    at Job.finished (node_modules/bull/lib/job.js:494:13)
    at async RoaiPortfolioCalculator.initialize (main.js)
    at async RoaiPortfolioCalculator.getSnapshot (main.js)
    at async PortfolioService.getDetails (main.js)
    at async Promise.all (index 1)
    at async PortfolioService.getAccounts (main.js)
    at async PortfolioService.getAccountsWithAggregations (main.js)
```

`initialize()` appears to enqueue a historical-data-gather job and `await job.finished()`; under load that job exceeds Bull's stall limit and the rejection propagates out as a 500 for the whole request.

### Contributing factors observed

1. **Prisma interactive-transaction timeout on historical writes.** During the gather triggered by the calc:
   ```
   Transaction API error: A rollback cannot be executed on an expired transaction.
   The timeout for this transaction was 5000 ms, however 5406 ms passed since the start
   of the transaction. Consider increasing the interactive transaction timeout or doing less work.
   ```
   Large `MarketData` writes (12 years of daily bars per symbol) exceed the hardcoded 5000 ms transaction timeout, roll back, write nothing, and the gather job then stalls — so the calc never gets the data it's waiting on.

2. **Per-request full recomputation.** `getSnapshot` rebuilds the entire time-series on each request rather than serving a cached/precomputed snapshot, so the cost scales with total activity count and a `range=max` request is ~17 s cold even when it does complete.

3. **JS heap.** Before the calc started failing outright, `range=max` for this dataset OOM'd Node at the default heap; it needed `--max-old-space-size` raised to ~20 GB to stop crashing. An in-memory time-series over a few hundred thousand price points shouldn't require 20 GB.

### Reproduction

1. Import ~10k activities for one account spanning many years and ~150+ symbols (a tax-loss-harvesting / direct-indexing export does this naturally — same symbol bought/sold hundreds of times across stock splits).
2. Let the historical gather run; observe `Transaction ... expired` errors and gather jobs that "complete" without writing rows.
3. `GET /api/v1/account` → 500 with the stack trace above.

Notably: with only the ~4,700-activity accounts loaded, everything worked and reconciled correctly. Adding the ~9,700-activity account is what tipped the calculator over.

### Suggested directions

- Don't block `RoaiPortfolioCalculator.initialize()` on a synchronous gather job; treat missing market data as a soft gap (use last-known / current price) and gather asynchronously, so a data hole in one symbol can't 500 the whole user.
- Make the Prisma interactive-transaction timeout configurable (env), and/or chunk `MarketData` writes so a 12-year backfill doesn't exceed it.
- Cache/precompute the portfolio snapshot and invalidate incrementally, rather than full recompute per request.
- Expose Bull `lockDuration` / `stalledInterval` / `maxStalledCount` as config for large datasets.

Happy to provide a sanitized export that reproduces this.

ecohash-co:
## Problem

On instances with a large transaction history, **all** `/api/v1/portfolio/*` and `/api/v1/account` requests return `HTTP 500` with `job stalled more than allowable limit`, and the failure is **permanent** — restarting the container does not help; the data has to be flushed from Redis by hand. Reported repeatedly in #6914 and #3755 (the heap-related #2241 is adjacent).

## Root cause

`PortfolioCalculator.initialize()` enqueues a snapshot computation on `PORTFOLIO_SNAPSHOT_COMPUTATION_QUEUE` with `jobId = userId` and then `await job.finished()`. The queue's `lockDuration` comes from `DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT`, which defaults to `30000` (30 s).

For a large portfolio the synchronous `computeSnapshot()` runs well beyond 30 s, so Bull's lock cannot be renewed in time, the still-running job is flagged as *stalled*, retried, and finally fails with `job stalled more than allowable limit`. That rejection propagates out of `getSnapshot()` as a `500`.

It then becomes **permanent**: `PORTFOLIO_SNAPSHOT_PROCESS_JOB_OPTIONS` sets `removeOnComplete: true` but has **no `removeOnFail`**, and the job id is the user id. The failed job therefore stays in the queue, so every later request finds it via `getJob(userId)` and `job.finished()` rejects immediately with the stored failure. There is no recovery short of manually clearing Redis — which is exactly the "reboot the container and flush data" workaround users describe in #3755.

## Changes

1. **Add `removeOnFail: true` to `PORTFOLIO_SNAPSHOT_PROCESS_JOB_OPTIONS`.** A failed snapshot job is now removed instead of being pinned under the user id, so the next request enqueues a fresh computation rather than re-reading a stale failure. This turns a permanent outage into a transient, self-healing one, and mirrors the existing `removeOnComplete: true`.
2. **Raise `DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` from `30000` to `300000` (30 s → 300 s)** so large portfolios compute successfully out of the box. The value stays overridable via the existing `PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` environment variable (added in #3894).

## How it was verified

Reproduced on a self-hosted instance with roughly **14,600 activities across ~12 years and 6 accounts**:

- With the 30 s default, every `/api/v1/account` and `/api/v1/portfolio/details?range=max` request returned `500`:
  ```
  ERROR [ExceptionsHandler] Error: job stalled more than allowable limit
      at async RoaiPortfolioCalculator.getSnapshot
  ```
  and stayed failed across container restarts until the failed Bull job was deleted from Redis by hand.
- The snapshot computation itself completes in about **152 s**:
  ```
  [PortfolioSnapshotProcessor (PORTFOLIO)] Portfolio snapshot calculation ... has been completed in 152.501 seconds
  ```
  i.e. ~5× the default lock duration.
- With the higher default lock duration the computation finishes and every endpoint returns `200`; with `removeOnFail` a one-off stall (or any transient failure) no longer wedges the user permanently — the following request simply recomputes.

## Test plan

- [ ] `nx lint common` is clean
- [ ] Existing portfolio-snapshot / calculator unit tests pass
- [ ] On a large portfolio, `/api/v1/portfolio/details?range=max` returns `200` once the computation completes
- [ ] After a forced snapshot-job failure, the next request recomputes instead of returning a cached `500`

## References

- Closes #6914
- Refs #3755, #2241
- Builds on #3894 (which exposed the timeout as an environment variable)


libs/common/src/lib/config.ts
```diff
@@ -89,7 +89,7 @@ export const DEFAULT_PORT = 3333;
 export const DEFAULT_PROCESSOR_GATHER_ASSET_PROFILE_CONCURRENCY = 1;
 export const DEFAULT_PROCESSOR_GATHER_HISTORICAL_MARKET_DATA_CONCURRENCY = 1;
 export const DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_CONCURRENCY = 1;
-export const DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT = 30000;
+export const DEFAULT_PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT = 300000;
```
dtslvr, line 92: Since this can already be configured via the `PROCESSOR_PORTFOLIO_SNAPSHOT_COMPUTATION_TIMEOUT` environment variable, let's keep the default value as is.

dtslvr (changes requested):

dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/pull/7221.

PR #6934, opened 2026-05-24
Task/improve type safety in user account access component

KenTandrian:

## Changes

- Made `access` optional in `CreateOrUpdateAccessDialogParams` to correctly represent dialog creation state.
- Formulated strictly typed reactive form control for token generation, eliminating null-coalescing workarounds.
- Removed mock domain objects (such as fictional empty-string IDs) from parent-to-dialog creation calls.
- Safely normalized database-nullable properties (like `alias` and `grantee`) to `undefined` or default strings.
- Tightened up encapsulation (`protected` and `private` visibilities) and immutability (`readonly` modifiers).
- Replaced constructor-based dependency injection with the `inject` function.
- Replaced deprecated `getDeviceInfo()` method with computed signal.

## Resolved type errors

5 errors (before: 108, after: 103)


dtslvr (approved):

PR #6953, opened 2026-05-28
Fix yahoo-finance search for US-listed ETFs with options chains

ccfiel:
Yahoo Finance's /v1/finance/search endpoint returns OPTION and FUTURE quoteTypes alongside the requested EQUITY/ETF row. yahoo-finance2's SearchQuoteYahoo* schemas don't include those quoteTypes, so AJV rejects the entire payload and the search() call throws. Ghostfolio's YahooFinanceService.search() returns an empty array, breaking symbol lookup in the Add Activity dialog for any US-listed ETF with an active options chain (IXUS, VTI, IVV, VOO, QQQM, SCHF, VIG, etc.).

The downstream quoteTypes.includes(quoteType) filter already excludes OPTION rows, so disabling strict response validation here is safe and preserves all existing behavior.

## Tested

- Manual repro: typed `IXUS`, `VTI`, `IVV`, `VOO` in Add Activity → empty dropdown before fix, correct ETF results after fix.
- Library-level verification: `yahooFinance.search('IXUS')` returns 5 results (ETF + regional variants) with the fix, throws without it.
- API-level verification: `GET /api/v1/symbol/lookup?query=IXUS` returns the IXUS ETF as the top hit.
- No regression: searches for tickers without options chains (`VWRA.L`, `BTC-USD`, etc.) return identical results before and after.

## Related

This is independent of my SELL fee PR — both bugs were uncovered while running Ghostfolio v3.3.0 against a portfolio that combines PH equities, UCITS ETFs, and legacy US-domiciled ETFs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Upgraded `nestjs` from version `11.1.19` to `11.1.21`
 
+### Fixed
+
+- Fixed the symbol lookup for US-listed ETFs with active options chains (e.g. `IXUS`, `VTI`, `IVV`, `VOO`, `QQQM`, `SCHF`, `VIG`) by disabling strict AJV validation on the Yahoo Finance search response (Yahoo returns `OPTION`/`FUTURE` quoteTypes that `yahoo-finance2`'s schemas reject, causing the entire payload to be discarded; the downstream `quoteType` filter already excludes options)
+
 ## 3.5.0 - 2026-05-24
 
 ### Added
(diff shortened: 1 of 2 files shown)


suzunn:
I noticed the patch adds the new 3.5.1 entry but also removes the existing 3.5.0 and 3.4.0 sections from `CHANGELOG.md`. The Yahoo Finance fallback itself looks well scoped, but this changelog hunk would drop release history for unrelated changes when merged.

I would keep the new `Fixed` bullet under the unreleased/current section and restore the removed historical sections before landing this. A small regression test around `search()` returning `error.result.quotes` would also make this safer, since that is the important behavior boundary here.

ccfiel:
Thanks for the careful review @suzunn  After digging into the rebase, I realized this PR is no longer needed.

The root cause was actually a `typeDisp` capitalization mismatch in yahoo-finance2 v3.14.0's schemas — `const: "option"`, `"equity"`, `"etf"`, etc. (lowercase) — versus Yahoo's actual API response which uses `"Option"`, `"Equity"`, `"ETF"` (capitalized). Every quote failed validation, not just OPTION rows.

yahoo-finance2 v3.14.2 corrected the schema constants to match Yahoo's capitalization. Verified locally: `yf.search('IXUS')` in v3.14.2 returns 7 quotes (including OPTION rows) successfully, no throw. Ghostfolio's existing downstream `quoteTypes.includes(quoteType)` filter then excludes OPTION rows from the user-facing dropdown, so the behavior is correct end-to-end.

Upstream already upgraded to v3.14.2 in 5bf04c48f, so this is fixed natively in Release 3.6.0. Closing — the upgrade resolves the root cause properly. Thanks again for the review; the CHANGELOG-history note is what led me to look at the rebase and discover the upstream fix!

PR #6954, opened 2026-05-28
Feature/auto-refresh user table in admin control panel every 30s

### LINKED ISSUE #6929: Auto-refresh user table every 30s
The goal of this issue is to add an automatic 30-second refresh interval (use `ms()`) to the [user table](https://github.com/ghostfolio/ghostfolio/tree/main/apps/client/src/app/components/admin-users) in the admin control panel.

The refresh should respect the pagination and retain the currently selected page.

x52-dev:
**Description:**

_What does this PR do?_

This adds a background polling mechanism to the Admin Users control panel, automatically refreshing the user list every 30 seconds.

**Architectural Details**

To ensure this is highly performant and production-ready, it avoids using standard setInterval polling:

1. RxJS Stream: Utilizes a BehaviorSubject for pagination state combined with timer(0, ms('30s')) for silent background polling.
2. Zero DOM Tearing: Updates this.dataSource.data directly rather than re-instantiating new MatTableDataSource(), preventing the Angular Material table from tearing down and rebuilding DOM nodes every tick.
3. Pagination Persistence: The background refresh respects the currently selected page. Navigating to a new page instantly drops the previous timer, updates the UI skeleton loader, and restarts the 30-second countdown.
4. Resiliency: Wrapped in a catchError block to return a safe fallback, ensuring the polling stream does not permanently die if there is a brief network timeout.

**Testing**

-Verified manual pagination instantly updates data.
-Verified 30-second loop triggers silently in the background.
-Verified pagination index remains correct after background fetch.

### DIFF
--- a/apps/client/src/app/components/admin-users/admin-users.component.ts
+++ b/apps/client/src/app/components/admin-users/admin-users.component.ts
@@ -59,8 +59,10 @@ import {
   personOutline,
   trashOutline
 } from 'ionicons/icons';
+import ms from 'ms';
 import { DeviceDetectorService } from 'ngx-device-detector';
 import { NgxSkeletonLoaderModule } from 'ngx-skeleton-loader';
+import { interval } from 'rxjs';
 import { switchMap, tap } from 'rxjs/operators';
 
 @Component({
@@ -184,6 +186,15 @@ export class GfAdminUsersComponent implements OnInit {
 
   public ngOnInit() {
     this.fetchUsers();
+
+    interval(ms('30 seconds'))
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe(() => {
+        this.fetchUsers({
+          pageIndex: this.paginator().pageIndex,
+          showLoading: false
+        });
+      });
   }
 
   protected formatDistanceToNow(aDateString: string) {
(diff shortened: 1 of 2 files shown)


apps/client/src/app/components/admin-users/admin-users.component.ts
```diff
@@ -116,6 +118,8 @@ export class GfAdminUsersComponent implements OnInit {
   private readonly router = inject(Router);
   private readonly userService = inject(UserService);
 
+  private readonly pageIndex$ = new BehaviorSubject<number>(0);
```
dtslvr: Could you make use of `this.paginator().page` as the single source of truth? That would let you drop the `pageIndex$` subject, since `MatPaginator` already tracks the page.

dtslvr (changes requested):

apps/client/src/app/components/admin-users/admin-users.component.ts
x52-dev: Sure, I have enhanced the code to address this issue. Now using, pageIndex from <mat-paginator>. @dtslvr

dtslvr (approved):
Thanks for your contribution, @x52-dev! 🙂

PR #6957, opened 2026-05-28
Task/improve type safety in HTTP interceptor

KenTandrian:

## Changes

- **Modernized Dependency Injection**: Refactored `HttpResponseInterceptor` and `AuthInterceptor` to use Angular's modern `inject()` function, removing the constructor parameter boilerplate.
- **Enhanced Type-Safety**: Refactored the signature of the `intercept` method in both interceptors to use generic types (`HttpRequest<T>` and `Observable<HttpEvent<T>>`) instead of `any`.
- **Resolved Null and Undefined Compiler Warnings**: Handled strict compilation issues in `HttpResponseInterceptor` by adding optional chaining for `error.url` (which is nullable) and `snackBarRef` (which can be undefined).
- **Cleaned Up Interception Chain**: Removed the redundant, no-op `tap` operator block and its unused import in `HttpResponseInterceptor` to reduce execution overhead.

## Additional context (if any)

These modernizations align both HTTP interceptors with Angular 21 best practices and eliminate runtime crash risks (such as trying to call `.afterDismissed()` on an uninitialized `snackBarRef` when triggering a forbidden error on the auth page).

## Resolved type errors

8 errors (before: 106, after: 98)

### DIFF
--- a/apps/client/src/app/core/auth.guard.ts
+++ b/apps/client/src/app/core/auth.guard.ts
@@ -3,7 +3,7 @@ import { UserService } from '@ghostfolio/client/services/user/user.service';
 import { internalRoutes, publicRoutes } from '@ghostfolio/common/routes/routes';
 import { DataService } from '@ghostfolio/ui/services';
 
-import { Injectable } from '@angular/core';
+import { inject, Injectable } from '@angular/core';
 import {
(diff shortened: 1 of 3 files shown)


dtslvr (approved):

PR #6961, opened 2026-05-30
Task/improve type safety in file drop directive

KenTandrian:

## Changes

- **Resolved Strict Type Errors**: Guarded the file drop emission to guarantee `event.dataTransfer` and its associated `files` property are present, resolving the strict compilation error where a potentially undefined value was being emitted.
- **Modernized to Angular 21 Standards**:
  - Replaced the legacy `@Output()` decorator and `EventEmitter` class with the modern, functional `output()` API.
  - Refactored `@HostListener` decorators into declarative `host` metadata bindings within the `@Directive` decorator for cleaner, centralized configuration.

## Resolved type errors

2 errors (before: 98, after: 96)

### DIFF
--- a/apps/client/src/app/directives/file-drop/file-drop.directive.ts
+++ b/apps/client/src/app/directives/file-drop/file-drop.directive.ts
@@ -20,9 +20,10 @@ export class GfFileDropDirective {
     event.preventDefault();
     event.stopPropagation();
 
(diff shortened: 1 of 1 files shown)


dtslvr (approved):
Thanks for your contribution, @KenTandrian! 🙂

PR #6965, opened 2026-05-31
Task/improve type safety in analysis page component

KenTandrian:

## Changes

- **Type Safety**: Resolved all 8 strict type errors in `analysis-page.component.ts` (safe navigation, undefined checks, and list fallback mappings).
- **Signal Queries**: Replaced the legacy `@ViewChild` decorator with the modern `viewChild.required()` signal query.
- **Encapsulation**: Enforced `protected` visibility for template-bound members, and `private` for internal helpers.
- **Readability**: Enforced `readonly` on properties that are never reassigned.
- **API Updates**: Replaced the deprecated `getDeviceInfo()` call with a `computed` signal reading `deviceInfo()`.
- **Signal State**: Converted the `mode` property into a writable signal and updated associated template bindings (`mode()`).

## Resolved type errors

8 errors (before: 98, after: 90)


dtslvr (approved):
Thanks for your contribution, @KenTandrian 👍🏻

PR #6968, opened 2026-05-31
Feature/removed precious metals from the analysis data

### LINKED ISSUE #6826: Commodity assets (precious metals) schould be excluded from the “By Country” and “By Sector” charts
`MANUAL` assets in the Commodities category (e.g., Precious Metals like physical gold) are currently included in the “By Country” and “By Sector” charts on the _Portfolio/Allocations_ view.

These dimensions are irrelevant to such assets, resulting in misleading or artificial allocations.

Therefore, excluding those commodity assets (specifically Precious Metals) from the “By Country” and “By Sector” charts would resolve this issue.

Cryptocurrencies are already excluded from these charts, so applying the same logic to `MANUAL` commodity assets would make the behavior consistent.

swissbuechi:
Removed precious metals from the analysis data by continents, countries, sectors and markets.

Solves: https://github.com/ghostfolio/ghostfolio/issues/6826

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Removed
+
+- Removed precious metals (Commodities) from the analysis data by continents, countries, sectors and markets.
+
 ### Added
 
 - Added an automatic refresh every 30 seconds to the users table in the admin control panel
--- a/apps/api/src/app/portfolio/portfolio.service.ts
+++ b/apps/api/src/app/portfolio/portfolio.service.ts
@@ -1453,7 +1453,10 @@ export class PortfolioService {
     for (const [, position] of Object.entries(holdings)) {
       const value = position.valueInBaseCurrency;
 
-      if (position.assetProfile.assetClass !== AssetClass.LIQUIDITY) {
+      if (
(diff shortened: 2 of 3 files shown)


dtslvr (changes requested):
Thanks for your contribution, @swissbuechi! 😎

I would check the asset class instead of the sub-class (as the comment below suggests). What do you think?

Please also add an entry to the changelog.

swissbuechi:
Ooh yeah I actually had it on the asset first and decided to make it more granular. This is also the reason it doesn't align with my extended comment.

I'll fix it 👍🏻

swissbuechi:
@dtslvr I completed the requested changes.

PS: Have you ever considered moving to something like https://github.com/semantic-release/semantic-release that would automate the changelog and release generation based on commit messages? I use it on all our company GitLab projects and it saves so much time. You'd need to create a new dev or staging branch and then only merge to main if you want a new version to be published thought. It would also require all contributors to strictly use the angular commit syntax.

dtslvr:
I will be happy to review your changes once the build is green, @swissbuechi.

Regarding [semantic-release](https://github.com/semantic-release/semantic-release), thanks for the suggestion. We will keep it in mind.

swissbuechi:
@dtslvr Thanks :)

Build should pass now. I initially missed a new line in the `CHANGELOG.md`...

dtslvr:
Superseded by #7011.

PR #6975, opened 2026-06-01
Feature/support migrating an existing asset profile to manual data source

### LINKED ISSUE #6939: Support migrating an existing asset profile to MANUAL data source
Extend the [asset profile dialog](https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/app/components/admin-market-data/asset-profile-dialog) with a **Convert to MANUAL** button so users can take over price maintenance for an asset whose upstream data source has stopped working.

Image

## Prerequisites

1. Navigate to _Admin Control_ > _Market Data_.
2. Add an asset profile (e.g. search for `AAPL`)
3. Click the edit (pencil) icon to switch to edit mode

### Frontend

1. Generate a UUID
2. Send a PATCH to `/api/v1/admin/profile-data/:dataSource/:symbol` with:
   - `dataSource: 'MANUAL'`
   - `symbol: <uuid>`
   - the current values of `assetClass`, `assetSubClass`, `countries`, `holdings`, `name`, `sectors`, `url`

The _UUID_ avoids collisions in the `MANUAL` namespace.

### Backend

Changes in `patchAssetProfileData()`:

1. The rename branch currently returns right after swapping the identifier. Extend it so it also applies any field values sent in the same request.
2. When `dataSource` becomes `MANUAL`, delete the `SymbolProfileOverrides` row (no longer read for `MANUAL` asset profiles).

Historical market data is already re-pointed to the new identifier by the existing rename logic.

AyushMishraa:
On the Frontend side, I have extended the asset profile dialogue to support Manual dataSource.
On the backend side, I have implemented the required changes for the manual datasource support.

### DIFF
--- a/apps/api/src/app/admin/admin.service.ts
+++ b/apps/api/src/app/admin/admin.service.ts
@@ -613,13 +613,16 @@ export class AdminService {
         // Check if we need to delete SymbolProfileOverrides when converting to MANUAL
         let deleteOverrides = false;
         if (finalDataSource === 'MANUAL') {
-          const [renamedProfile] =
-            await this.symbolProfileService.getSymbolProfiles([
-              {
+          const renamedProfile =
+            await this.prismaService.symbolProfile.findFirst({
+              where: {
                 dataSource: finalDataSource,
                 symbol: newSymbol as string
+              },
+              select: {
+                SymbolProfileOverrides: true
               }
-            ]);
+            });
 
           deleteOverrides = !!renamedProfile?.SymbolProfileOverrides;
         }
--- a/apps/api/src/app/admin/admin.service.ts
+++ b/apps/api/src/app/admin/admin.service.ts
@@ -610,6 +610,20 @@ export class AdminService {
           url: url as string
         };
 
+        // Check if we need to delete SymbolProfileOverrides when converting to MANUAL
+        let deleteOverrides = false;
+        if (finalDataSource === 'MANUAL') {
+          const [renamedProfile] =
+            await this.symbolProfileService.getSymbolProfiles([
+              {
+                dataSource: finalDataSource,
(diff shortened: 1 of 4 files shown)


apps/api/src/app/admin/admin.service.ts
dtslvr, line 679: I get the following error here, when I try to convert an asset profile to `MANUAL`:
```
Error updating asset profile identifier: PrismaClientKnownRequestError: 
Invalid `prisma.symbolProfile.update()` invocation:
An operation failed because it depends on one or more records that were required but not found. No 'SymbolProfileOverrides' record was found for a nested delete on one-to-one relation 'SymbolProfileToSymbolProfileOverrides'.
```

apps/client/src/app/components/admin-market-data/asset-profile-dialog/asset-profile-dialog.html
```diff
@@ -162,6 +162,16 @@ <h1 class="flex-grow-1 m-0" mat-dialog-title>
                     <ng-container i18n>Cancel</ng-container>
                   </button>
                 </form>
+                <p class="text-muted">or</p>
```
dtslvr: Use `my-1` instead of `text-muted`

dtslvr (changes requested):

apps/api/src/app/admin/admin.service.ts
AyushMishraa: According to what I have seen when deleting the SymbolProfileOverrides, it did not check for the symbolProfileOverrides.
Included this check before deleting it.

dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/pull/7376.

PR #6976, opened 2026-06-01
Feature/add active or closed toggle to holdings overview on mobile

swissbuechi:
Feature/add active or closed toggle to holdings overview on mobile.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added active/closed toggle to holdings overview on mobile
 - Added support for routing selected requests through the _OpenRouter_ `web_fetch` tool in the `FetchService`
 
 ### Changed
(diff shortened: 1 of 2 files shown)


dtslvr:
Thanks for the PR, @swissbuechi! The behavior is actually by design as we intentionally hide this element on mobile. I will close this PR for now.

swissbuechi:
@dtslvr What's the exact reasoning behind this decision? It works perfectly fine on mobile, and hiding this feature is basically an unnecessary limitation that, in my opinion, doesn't align with the advertising at all.

> Progressive Web App (PWA) with a mobile-first design

Thanks for further clarification.

dtslvr:
Thanks for the feedback, @swissbuechi. The choice was made to reduce clutter on smaller screens, but we will keep it in mind for future versions.

swissbuechi:
Hi @dtslvr, thank you for the explanation. For me as a nearly mobile-only user, it's kind of frustrating as there's plenty of space left on my fairly big phone screen.

Could we perhaps consider this as an experimental feature? I’m currently running my own fork, so it’s already solved for me, but I imagine others would appreciate it as well.

PR #6981, opened 2026-06-02
Feature/extend public API with endpoint to update asset profile data

Sjohn21:
## Summary

Adds a dedicated admin API endpoint to update asset profile data for countries, sectors and holdings.

This enables external community scrapers or tools to update ETF profile data without adding browser automation or WAF workaround logic to Ghostfolio core.

## Changes

1. Added `UpdateAssetProfileDataDto`
2. Added `PUT /api/v1/admin/asset-profile-data/:dataSource/:symbol`
3. Reused the existing `patchAssetProfileData()` logic
4. Added `holdings` to `SymbolProfileOverrides` persistence

## Context

This is related to the Trackinsight data gathering issues discussed in #5924 and the architectural direction discussed in #6357.

## Test

Tested locally with `npx nx run api:build`.

Also tested the new endpoint locally with `WTCH.AS`, including countries, sectors and holdings. The endpoint returned HTTP 200 and the ETF holdings were visible in the Allocations page with experimental features enabled.


dtslvr (changes requested):
Thanks for your contribution, @Sjohn21!

Here is a suggestion on how this is exposed. Rather than adding it to the admin controller, I'd prefer a dedicated `PATCH /api/v1/asset-profiles/:dataSource/:symbol` endpoint in a new `AssetProfilesModule` under `/app/endpoints/asset-profiles`.

* `PATCH` instead of `PUT` (fields are optional / partial update)
* Dedicated module (`admin` is internal, this is intended to be a stable endpoint)

What do you think?

Sjohn21:
Thanks for the feedback @dtslvr. 
I refactored the change to a dedicated `AssetProfilesModule` under `apps/api/src/app/endpoints/asset-profiles` and changed the endpoint to `PATCH /api/v1/asset-profiles/:dataSource/:symbol`.

For non manual assets, the endpoint now writes to `SymbolProfileOverrides` only, so provider data remains untouched.

dtslvr (approved):
Thanks again for your contribution, @Sjohn21! 👍🏻

PR #7011, opened 2026-06-09
Task/improve unknown bucket grouping in allocations

### LINKED ISSUE #6812: Assets Class "Cryptocurrency" should not be listed as USD (or any other currency) in the "By Currency" allocation chart
All assets with the sub-category `Cryptocurrency` should be excluded from the "By Currency" chart under _Portfolio/Allocations_ as the tracking currency is not relevant for crypto.

This currently creates a false illusion of having for example USD exposure just because you track BTC in USD.

But in crypto is not like a stock or cash when it comes to it's underlying currency. It's not relevant.

Alternatively it could also be considered to just list the cryptocurrency directly and not the underlying currency in the "By Currency" chart.

For example:
- 50% USD
- 25% BTC
- 25% SOL

I'm using a MANUAL asset profile and track my crypto in my home currency which I use to buy through our local exchange.

Same thing should probably be considered in the X-Ray reports accordingly.

### LINKED ISSUE #6826: Commodity assets (precious metals) schould be excluded from the “By Country” and “By Sector” charts

dtslvr:


swissbuechi:
Hi @dtslvr. I really appreciate your efforts to fix both of my reported issues. Works great!

To make things more transparent, I'd recommend to forther split the "no data available" slices by introduction something like a "no exposure" category.

This way, the graphs would be easier for users to comprehend, and we could more easily identify a potential misconfiguration of an asset.

PR #7014, opened 2026-06-12
fix (ui): make dividend import load request trigger correctly

sebmatton:
- Problem: Imports Dividends from the activities page did not trigger any network request.
- Root cause: selected holding value shape in template did not match handler expectation.
- Fix: pass holding object directly from select option and update select trigger path.
- Verification: clicking Load Dividends now sends the correct request and import works end-to-end.


dtslvr:
Superseded by https://github.com/ghostfolio/ghostfolio/pull/7020.

PR #7047, opened 2026-06-15
Bugfix/restrict maximum height of import activities dialog

### LINKED ISSUE #6600: [BUG] To many notices in "import activities" modal make modal action buttons inaccessible

jhernaezayuso:
resolves #6600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Improved the loading state when customizing the rule thresholds on the _X-ray_ page
+- Fixed an issue where the import activities dialog could overflow the screen
 
 ## 3.10.0 - 2026-06-13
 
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks for your contribution, @jhernaezayuso! 😎

PR #7050, opened 2026-06-16
Task/improve type safety in holding detail dialog component

KenTandrian:

## Changes

- **Type Safety**: Resolved all 8 strict type errors in `holding-detail-dialog.component.ts` (safe navigation, undefined checks, form control type safety, and list fallback mappings).
- **Encapsulation**: Enforced `protected` visibility for template-bound members, and `private` for internal helpers and state.
- **Immutability**: Enforced `readonly` on properties that are never reassigned.
- **Form Safety**: Strongly-typed the `holdingForm` controls (`tags` as `FormControl<Tag[]>`) and migrated to type-safe control access (`holdingForm.controls.tags.valueChanges`) instead of using `.get('tags')`.
- **Dead Code Cleanup**: Removed unused instance fields (`netPerformance`, `netPerformancePrecision`, and `netPerformancePercent`) and their associated assignments.

## Resolved type errors

8 errors (before: 92, after: 84)

### DIFF
--- a/apps/client/src/app/components/holding-detail-dialog/holding-detail-dialog.component.ts
+++ b/apps/client/src/app/components/holding-detail-dialog/holding-detail-dialog.component.ts
@@ -115,7 +115,7 @@ export class GfHoldingDetailDialogComponent implements OnInit {
   protected averagePrice: number;
   protected averagePricePrecision = 2;
   protected benchmarkDataItems: NullableLineChartItem[];
-  protected benchmarkLabel = $localize`Average Unit Price`;
+  protected readonly benchmarkLabel = $localize`Average Unit Price`;
   protected countries: {
     [code: string]: { name: string; value: number };
   };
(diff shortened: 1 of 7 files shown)


apps/client/src/app/components/holding-detail-dialog/holding-detail-dialog.component.ts
```diff
@@ -106,75 +113,78 @@ import { HoldingDetailDialogParams } from './interfaces/interfaces';
...
-  public accounts: Account[];
-  public assetClass: string;
-  public assetSubClass: string;
-  public averagePrice: number;
-  public averagePricePrecision = 2;
-  public benchmarkDataItems: LineChartItem[];
-  public benchmarkLabel = $localize`Average Unit Price`;
-  public countries: {
+  protected accounts: Account[];
+  protected activitiesCount: number;
```
dtslvr, line 117: 👍🏻

apps/client/src/app/components/holding-detail-dialog/holding-detail-dialog.component.ts
```diff
@@ -106,75 +113,78 @@ import { HoldingDetailDialogParams } from './interfaces/interfaces';
...
-  public netPerformanceWithCurrencyEffectPrecision = 2;
-  public pageIndex = 0;
-  public pageSize = DEFAULT_PAGE_SIZE;
-  public quantity: number;
-  public quantityPrecision = 2;
-  public reportDataGlitchMail: string;
-  public routerLinkAdminControlMarketData =
+  protected dataProviderInfo: DataProviderInfo;
+  protected dataSource: MatTableDataSource<Activity>;
+  protected dateOfFirstActivity: string;
```
dtslvr: Could you change to `Date` and convert in `fetchHoldingDetail()` of `DataService` as well?

libs/common/src/lib/interfaces/line-chart-item.interface.ts
dtslvr, line 1: What about

```
export interface ChartItem<T = number> {
  date: string;
  value: T;
}

export type LineChartItem = ChartItem<number>;
export type NullableLineChartItem = ChartItem<number | null>;
```

dtslvr (changes requested):

libs/common/src/lib/interfaces/line-chart-item.interface.ts
KenTandrian: Actually since `T` defaults to `number`, we don't need another `ChartItem`?

```typescript
export interface LineChartItem<T = number> {
  date: string;
  value: T;
}

export type NullableLineChartItem = LineChartItem<number | null>;
```

What do you think?

dtslvr: Nice! 🙂

dtslvr (approved):

PR #7053, opened 2026-06-16
Bugfix/dark mode styling of safe withdrawal rate selector in FIRE section

jhernaezayuso:
This fixes the SWR dropdown on the FIRE page so the opened options remain legible in dark mode

Before:
Screenshot From 2026-06-16 20-10-16

After:
Screenshot From 2026-06-16 19-56-05

### DIFF
--- a/apps/client/src/app/pages/portfolio/fire/fire-page.scss
+++ b/apps/client/src/app/pages/portfolio/fire/fire-page.scss
@@ -32,7 +32,6 @@
 
     option {
       color: rgb(var(--dark-primary-text));
-      background-color: rgb(var(--light-background));
     }
   }
 }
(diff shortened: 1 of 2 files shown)


apps/client/src/app/pages/portfolio/fire/fire-page.scss
```diff
@@ -29,5 +29,10 @@
     @include select-arrow(variables.$light-primary-text);
 
     color: rgb(var(--light-primary-text));
+
+    option {
+      color: rgb(var(--dark-primary-text));
+      background-color: rgb(var(--light-background));
```
dtslvr: Based on the screenshot, the background color doesn't appear to be needed 🤔

jhernaezayuso: Right! It's not necessary. I've removed it

dtslvr (approved):
Thanks for this improvement, @jhernaezayuso! 👍🏻

PR #7057, opened 2026-06-16
Fix page tab label truncation

### LINKED ISSUE #7055: Truncate long labels in page tabs
Long labels in the [page tabs component](https://github.com/ghostfolio/ghostfolio/tree/main/libs/ui/src/lib/page-tabs) should be truncated with an ellipsis. Leverage _Bootstrap_ utility classes instead of adding custom CSS rules.

Image

WilliamK112:
## Description

- Truncates visible page tab labels with Bootstrap utility classes
- Keeps the existing icon layout and avoids adding custom CSS

Closes #7055

## Testing

- npx prettier --check libs/ui/src/lib/page-tabs/page-tabs.component.html
- npx nx lint ui
- npx prisma generate
- npx nx test ui --runInBand

### DIFF
--- a/libs/ui/src/lib/page-tabs/page-tabs.component.html
+++ b/libs/ui/src/lib/page-tabs/page-tabs.component.html
@@ -42,8 +42,13 @@
 
 <ng-template #tabContent let-tab>
   <ion-icon
+    class="flex-shrink-0"
     [name]="tab.iconName"
     [size]="deviceType === 'mobile' ? 'large' : 'small'"
   />
-  <div class="d-none d-sm-block ml-1" [innerHTML]="tab.label"></div>
+  <div
+    class="d-none d-sm-block ml-1 text-truncate"
+    style="max-width: 10.5rem"
+    [innerHTML]="tab.label"
+  ></div>
 </ng-template>
--- a/libs/ui/src/lib/page-tabs/page-tabs.component.html
+++ b/libs/ui/src/lib/page-tabs/page-tabs.component.html
@@ -47,8 +47,7 @@
     [size]="deviceType === 'mobile' ? 'large' : 'small'"
   />
   <div
-    class="d-none d-sm-block ml-1 text-truncate"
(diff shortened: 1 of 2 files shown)


dtslvr (changes requested):
This change doesn't seem to resolve the issue.

WilliamK112:
Thanks for the quick review. You were right: the previous change applied `text-truncate`, but Angular Material's generated tab content still auto-sized to the label text, so the label itself was not actually overflowing.

I updated the tab content so the icon does not shrink and the visible label has a bounded width while still using Bootstrap's `text-truncate` for the ellipsis. I also verified locally with a temporary long tab label: the tab link stayed within the sidebar width and the label had `scrollWidth > clientWidth`, so the ellipsis path is active.

Validation:
- `npx prettier --check libs/ui/src/lib/page-tabs/page-tabs.component.html`
- `npx nx lint ui` (passes; existing warnings only)
- `npx nx test ui --runInBand`

dtslvr (changes requested):
Thanks for the update, @WilliamK112!

The approach works, but `10.5rem` is a magic number we can't merge. Could you find a way to derive the width instead of hardcoding it?

WilliamK112:
Thanks, agreed on avoiding the hardcoded width. I updated the tab layout so the label width is derived from the available tab link space instead:

- removed the inline `max-width: 10.5rem`
- set the Material tab content/text-label wrappers to `min-width: 0` and full width
- made the visible label flex into the remaining space while the icon stays fixed-size

Validation:
- `npx prettier --check libs/ui/src/lib/page-tabs/page-tabs.component.html libs/ui/src/lib/page-tabs/page-tabs.component.scss`
- `npx nx lint ui` (passes with existing warnings)
- `npx nx test ui --runInBand`

dtslvr:
This looks a bit different from the expected layout.

overflow-in-page-tabs-component-pr

PR #7077, opened 2026-06-20
Task/improve type safety in import activities service

KenTandrian:

## Changes

- **Type Safety**: Resolved all 14 strict type errors in `import-activities.service.ts` by replacing `null` assignments to optional DTO fields with `undefined`, mapping nullable fields with nullish coalescing / empty string fallbacks, adding optional chaining check for `account.name`, and validating helper date parse return values.
- **Type Safety / Refactoring**: Removed all `any` types (and `any[]` array signatures) from the service methods in favor of type-safe `Record<string, unknown>` mapping, and introduced Lodash `isString` and `isNumber` checks to guard the parser methods.
- **RxJS Modernization**: Replaced the manual subscription and Promise wrapper in `importJson` with RxJS `firstValueFrom` to modernize Observable-to-Promise conversion.
- **Modern JS/TS**: Simplified object key conversion in `lowercaseKeys` using modern `Object.fromEntries` / `Object.entries`, and refactored loop-push logic to functional `.map()` calls in `importSelectedActivities`.

## Resolved type errors

14 errors (before: 84, after: 70)

### DIFF
--- a/apps/client/src/app/services/import-activities.service.ts
+++ b/apps/client/src/app/services/import-activities.service.ts
@@ -83,22 +83,22 @@ export class ImportActivitiesService {
         assetProfiles.push({
           currency,
           symbol,
-          assetClass: null,
-          assetSubClass: null,
-          comment: null,
+          assetClass: undefined,
+          assetSubClass: undefined,
+          comment: undefined,
           countries: [],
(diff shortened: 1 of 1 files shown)


dtslvr (approved):

PR #7118, opened 2026-06-23
Add data decimation support to line chart

### LINKED ISSUE #2553: Setup data decimation option in line chart component

chandranilbakshi:
Closes #2553 

## Summary

  Adds optional Chart.js data decimation support to `LineChartComponent`.

  The component now accepts:

  - `dataDecimation`
  - `dataDecimationThreshold`

  The home overview chart opts into decimation with a lower threshold so dense
  portfolio history data can be reduced on smaller screens. Storybook also exposes
  both controls for local verification.

  ## Verification

  - `npx tsc -p libs/ui/tsconfig.lib.json --noEmit`
  - `npx ngc -p apps/client/tsconfig.app.json --noEmit`

  Manual Storybook check:

  - Opened the Line Chart story
  - Tested with 1000 historical data points
  - Set `dataDecimation=true`
  - Set `dataDecimationThreshold=100`
  - Confirmed Chart.js reduced the rendered dataset from 1000 points to 901 points
  during local debugging

### DIFF
--- a/apps/client/src/app/components/home-overview/home-overview.html
+++ b/apps/client/src/app/components/home-overview/home-overview.html
@@ -69,6 +69,8 @@
             unit="%"
             [class.pr-3]="deviceType() === 'mobile'"
             [colorScheme]="user()?.settings?.colorScheme"
+            [dataDecimation]="true"
+            [dataDecimationThreshold]="100"
             [hidden]="historicalDataItems()?.length === 0"
             [historicalDataItems]="historicalDataItems()"
             [isAnimated]="user()?.settings?.dateRange === '1d' ? false : true"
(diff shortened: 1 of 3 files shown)


chandranilbakshi:
The graph may not be visually different with data decimation on/off but it is working and I have verified by console logging the initial and the final number of data points in the file `libs/ui/src/lib/line-chart/line-
  chart.component.ts` (I haven't pushed the logging code). Here is the log:
  
image

As you can see in the image, the rendered datapoints is less than the original ones.

Right now I have only implemented this in the home-overview component. I can also implement it in the other components as well.

PR #7148, opened 2026-06-27
Feature/include cash in portfolio performance

dtslvr:


dtslvr:
With this PR, cash is part of the portfolio value and performance calculation instead of being added on top, so it shows up on the home screen. Would appreciate your view on this, @KenTandrian.

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
```diff
@@ -448,11 +446,16 @@ export abstract class PortfolioCalculator {
         quantity: item.quantity,
         symbol: item.symbol,
         tags: item.tags,
-        valueInBaseCurrency: new Big(marketPriceInBaseCurrency).mul(
-          item.quantity
-        )
+        valueInBaseCurrency
       });
 
+      if (item.assetSubClass === 'CASH') {
```
KenTandrian: using `AssetSubClass.CASH` would be better compared to the raw string literal `'CASH'`, I guess?

KenTandrian (approved):
LGTM! 🚀

Including cash positions directly into the portfolio value and performance calculation makes total sense and cleans up the handling of cash assets nicely.

apps/api/src/app/portfolio/calculator/portfolio-calculator.ts
dtslvr: 👍🏻

dtslvr:
Thanks for the feedback, @KenTandrian! 😎

PR #7145, opened 2026-06-26
Fix manual asset symbol mapping

SeineEloquenz:
This PR should hopefully fix #4376.
I've tested the changes locally against an asset with mapping to QQQ and found the correct sector data etc. in the database (after I changed ghostfolio's useragent for the Trackinsight call, as that just returned an empty response with the original one)

PR #7150, opened 2026-06-27
Make trackinsight URL configurable

SeineEloquenz:
This PR makes the trackinsight URL configurable. This allows to run a puppeteering proxy to pass trackinsights AWS WAF protection (e.g. via [this project](https://github.com/SeineEloquenz/trackinsight-proxy). Together with #7145 this fixes the Trackinsight gathering flow for Manual (and likely Yahoo as well?) assets

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -11,6 +11,9 @@ on:
 jobs:
   build_and_push:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
     steps:
       - name: Checkout code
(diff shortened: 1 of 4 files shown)


dtslvr (commented):
I would rather not add a specific environment variable for this.

We already centralize requests in [FetchService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/fetch/fetch.service.ts). Could your service be wired in there as a generic per-domain override instead?

SeineEloquenz:
You mean like the OpenRouter web fetch service? Sure thing, that would work fine

PR #7173, opened 2026-06-30
Bugfix/data source validation of create asset profile with market data DTO

AkashNegi1:
## Description

Fix incorrect validation in `CreateAssetProfileWithMarketDataDto` where `@IsEnum()` was called with an array literal as the enum entity.

### Problem

`@IsEnum()` expects an enum type (like `DataSource`) but receives an array literal `[DataSource.MANUAL]`. Class-validator internally calls `Object.keys(['MANUAL'])` which returns `['0']` — the array index. This means validation always checks if the input equals `'0'` instead of checking if it's `'MANUAL'`, causing every valid request with `dataSource: 'MANUAL'` to be rejected with a 400 error.

### Fix

Replaced `@IsEnum` with `@IsIn`, which is designed for validating against a literal set of values.

### Changes

| File | Change |
|------|--------|
| `libs/common/src/lib/dtos/create-asset-profile-with-market-data.dto.ts` | Replaced `IsEnum` import with `IsIn` and updated the decorator |
| `CHANGELOG.md` | Added changelog entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed the validation of the `dataSource` field in the `CreateAssetProfileWithMarketDataDto` to use `IsIn`
 - Fixed a recurring issue where single-value fields were incorrectly validated as arrays in various endpoints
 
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks for your contribution, @AkashNegi1! 🙂

PR #7174, opened 2026-06-30
Task/add explicit @prisma/config dependency

slagiewka:
It's used in `.config/prisma.ts` and is coincidentally available for the `postinstall` to run correctly (already present in the lockfile).

However, some tools (that are not npm) might complain that it should either be explicit or hoisted with an `.npmrc` flag.


dtslvr (approved):
Thanks for your contribution, @slagiewka! 😎

PR #7175, opened 2026-07-01
Task/improve type safety in client services

KenTandrian:

## Changes

- **Modernized Dependency Injection**: Refactored dependency injection in `UserService` and `WebAuthnService` to use Angular's modern `inject()` function, simplifying constructors and class setup.
- **Signal-Based Device Detection**: Modernized `deviceType` in `UserService` to be a reactive `computed` signal wrapping `DeviceDetectorService`.
- **Impersonation Storage Type Safety**: Updated `ImpersonationStorageService.getId()` and the `hasImpersonationChangeSubject` stream type to support `string | null`, aligning with local storage APIs. Adjusted `header.component.ts` to expect `string | null` while keeping the classic `BehaviorSubject` reactivity.
- **Settings & Token Storage Type Safety**: Updated `SettingsStorageService.getSetting()` and `TokenStorageService.getToken()` return signatures to return `string | null`. Safeguarded callers (`admin-page.component.ts` and `demo-page.component.ts`) against null values.
- **User Service Type Safety**: Resolved type errors in `UserService` by using `undefined` instead of `null` on state reset. Implemented a clean type-guard filter inside `clearAllCookies()` to safely narrow types for cookie deletions.
- **WebAuthn Option Checks**: Enhanced `WebAuthnService.register()` to throw a runtime error if attestation options are null to prevent application crashes and propagate the error cleanly.

## Resolved type errors

7 errors (before: 36, after: 29)

### DIFF
--- a/apps/client/src/app/components/header/header.component.ts
+++ b/apps/client/src/app/components/header/header.component.ts
@@ -101,7 +101,7 @@ export class GfHeaderComponent implements OnChanges {
   protected hasPermissionToAccessAssistant: boolean;
   protected hasPermissionToAccessFearAndGreedIndex: boolean;
   protected hasPermissionToCreateUser: boolean;
-  protected impersonationId: string;
+  protected impersonationId: string | null;
   protected readonly internalRoutes = internalRoutes;
   protected isMenuOpen: boolean;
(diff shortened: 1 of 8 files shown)


dtslvr (approved):

PR #7183, opened 2026-07-02
Task/improve parsing of integer query parameters in GET api/v1/activities endpoint

### LINKED ISSUE #7180: Improve integer query parameter parsing in activities controller
The `skip` and `take` query parameters in the activities controller are currently parsed with manual `isNaN` [checks](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/app/activities/activities.controller.ts#L153).

The goal of this issue is to replace them with the built-in `ParseIntPipe` from _NestJS_ using the `optional` option.

AkashNegi1:
## Description

Replaced manual `isNaN` validation of the `skip` and `take` query parameters in the activities controller with NestJS's built-in `ParseIntPipe({ optional: true })`.

### Changes

- `apps/api/src/app/activities/activities.controller.ts`
  - Added `ParseIntPipe` to NestJS imports
  - Applied `@Query('skip', new ParseIntPipe({ optional: true }))` and `@Query('take', new ParseIntPipe({ optional: true }))` to delegate validation to the framework
  - Removed the now-unnecessary `isNaN(skip) ? undefined : skip` / `isNaN(take) ? undefined : take` guards

### Behavior

| Scenario | Before | After |
|---|---|---|
| `?skip=10` | `10` | `10` |
| `?skip=abc` | `undefined` (silently ignored) | `400 Bad Request` |
| No param | `undefined` | `undefined` |

`ParseIntPipe` with `optional: true` means the param is validated only when present — missing params pass through as `undefined`, while invalid values throw a proper HTTP error instead of being silently swallowed.

Closes
#7180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Updated the _Privacy Policy_
 - Updated the _Terms of Service_
+- Improved the parsing of integer query parameters (`skip` and `take`) in the `GET api/v1/activities` endpoint
 - Improved the language localization for German (`de`)
 - Improved the language localization for Japanese (`ja`)
 - Upgraded `nestjs` from version `11.1.21` to `11.1.27`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks for your contribution, @AkashNegi1! 🙌🏻

PR #7192, opened 2026-07-03
Bugfix/missing navigation on landing page for unauthenticated users

### LINKED ISSUE #7191: Unable to login with security token
**Bug Description**

The _Login_ button is missing in the header on the landing page.

**Expected Behavior**

The _Login_ button is available.

**Environment**

- Ghostfolio Version `3.19.0`

**Additional Context**

Might be a regression of https://github.com/ghostfolio/ghostfolio/pull/7175.

AkashNegi1:
## Description

Fixes a regression from #7175 where the "Sign in" button disappeared from the landing page for unauthenticated users. Fixes #7191.

### Root Cause

PR #7175 changed `UserService.reset()` to set `state.user = undefined` instead of `null`. The header template used `@if (user() === null)` to render the public header, but `undefined === null` evaluates to `false`, causing the entire public header block (including the "Sign in" and "Get Started" buttons) to be hidden.

### Changes

- `apps/client/src/app/components/header/header.component.html`: Changed `@if (user() === null)` to `@if (!user())` to correctly handle both `null` and `undefined` as unauthenticated states, while `@if (user())` continues to handle the logged-in state.

### Related Issues

- Fixes #7191
- Regression introduced in #7175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed the missing "Sign in" button on the landing page for unauthenticated users
 - Fixed the validation of the data source field of an asset profile with market data
 - Fixed a recurring issue where single-value fields were incorrectly validated as arrays in various endpoints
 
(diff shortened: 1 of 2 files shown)


dtslvr:
@KenTandrian, could you double-check if this change looks correct to you?

KenTandrian:
Hi @dtslvr, I can reproduce the bug and this change successfully fixes it.

Thanks @AkashNegi1 for the fix!

dtslvr (approved):
Thank you for providing this fix, @AkashNegi1! 🤓

PR #7214, opened 2026-07-04
Task/improve type safety in api page and public page components

KenTandrian:

## Changes

- Resolved type errors in API page and public page.
- Enforced encapsulation.
- Replaced constructor-based dependency injection with inject function.
- Added fallback for name to prettified symbol in public page since `name` cannot be undefined.

## Resolved type errors

2 errors (before: 24, after: 22)


dtslvr (approved):

PR #7219, opened 2026-07-04
Task/improve type safety of key in PropertyService

### LINKED ISSUE #7216: Improve type safety of key in PropertyService
The methods `delete()`, `getByKey()` and `put()` of the [PropertyService](https://github.com/ghostfolio/ghostfolio/blob/main/apps/api/src/services/property/property.service.ts) accept any `string` as key. Typos or unknown keys are not caught at compile time.

### Suggested Improvement

Introduce a union type derived from the `PROPERTY_*` constants in `libs/common/src/lib/config.ts` and use it as the type of the key parameter.

AkashNegi1:
Closes #7216 

### Problem

The methods `delete()`, `getByKey()` and `put()` of `PropertyService` accept `string` as the key parameter. Typos or unknown keys are not caught at compile time.

### Solution

Introduced a `PropertyKey` union type derived from the 27 `PROPERTY_*` constants and used it to type the `key` parameter in `PropertyService`.

### Changes

**`libs/common/src/lib/config.ts`**
- Added `PropertyKey` union type (lines 270–297) using `typeof` on each `PROPERTY_*` constant. No new runtime values, no import changes needed — the type can be imported from the same `@ghostfolio/common/config` path.

**`apps/api/src/services/property/property.service.ts`**
- `delete({ key }: { key: PropertyKey })`
- `getByKey<TValue extends PropertyValue>(aKey: PropertyKey)`
- `put({ key, value }: { key: PropertyKey; value: string })`

**`apps/api/src/app/admin/admin.service.ts`**
- Added `key as PropertyKey` cast in `putSetting()` at the boundary between the untyped HTTP layer and the typed service layer.

### Why admin.service.ts needed a change

`admin.service.ts` is the only call site where the key originates from an external source (URL parameter) rather than being a direct `PROPERTY_*` constant. The `putSetting(key: string, value: string)` method receives `key` as `string` from the HTTP request and passes it to `propertyService.put()` / `propertyService.delete()`. Since those methods now require `PropertyKey`, a cast is needed at this boundary.

All other call sites across the codebase (~35 locations in 15+ files) pass `PROPERTY_*` constants directly, which are already valid `PropertyKey` values — no changes were needed.

### What this catches

A typo like `propertyService.put({ key: 'BENCHMARKZ', value: '...' })` now fails at compile time:


libs/common/src/lib/config.ts
```diff
@@ -267,6 +267,35 @@ export const PROPERTY_SYSTEM_MESSAGE = 'SYSTEM_MESSAGE';
 export const PROPERTY_UPTIME = 'UPTIME';
 export const PROPERTY_WEB_FETCH_ROUTES = 'WEB_FETCH_ROUTES';
 
+export type PropertyKey =
```
dtslvr: What about
```
import type * as config from '../config';

export type PropertyKey = {
  [Key in keyof typeof config]: Key extends `PROPERTY_${string}`
    ? (typeof config)[Key]
    : never;
}[keyof typeof config];
```
in `libs/common/src/lib/types/property-key.type.ts` instead of this?

dtslvr (changes requested):
Thanks for your contribution, @AkashNegi1! 😎

libs/common/src/lib/config.ts
AkashNegi1: @dtslvr  Great suggestion, thank you! That's much cleaner — the mapped type with template literal inference avoids the manual duplication entirely, and adding a new PROPERTY_* constant will automatically be picked up.

I'll update the PR:

Revert the explicit PropertyKey union from config.ts
Create libs/common/src/lib/types/property-key.type.ts with the mapped type approach
Export it from the types barrel
Update the import paths in property.service.ts and admin.service.ts to import from @ghostfolio/common/types instead of @ghostfolio/common/config
Thanks for the review!

apps/api/src/app/admin/admin.service.ts
```diff
@@ -11,6 +11,7 @@ import {
   PROPERTY_IS_READ_ONLY_MODE,
   PROPERTY_IS_USER_SIGNUP_ENABLED
 } from '@ghostfolio/common/config';
+import { PropertyKey } from '@ghostfolio/common/types';
```
dtslvr: Please move down ⬇️

dtslvr (approved):
Thanks for your contribution, @AkashNegi1! 👍🏻

Issue #3590, opened 2024-07-20
Ghostfolio in het Nederlands


dtslvr:
As Ghostfolio continues to evolve, some translations are still missing. We need your help to make Ghostfolio fully accessible to Dutch-speaking users!

Here’s how you can participate in the translation work:

1. Open the translation file at https://github.com/ghostfolio/ghostfolio/blob/main/apps/client/src/locales/messages.nl.xlf
2. Search for entries with `state="new"`
3. Translate the content of the `<target>` element
4. Change the state from `"new"` to `"translated"`
5. Create a pull request with your translations

You don’t need to translate everything at once. Even small contributions are incredibly valuable and easier for us to review.

Your efforts will make a big difference. Many thanks in advance for your help!

turkaturki:
Hi I'd like to be assigned to this issue.

dtslvr:
I have extended the instructions by a changelog entry.

PR #7260, opened 2026-07-06
Task/improve language localization for NL

munzzyy:
Fills in the remaining untranslated strings in `messages.nl.xlf` (64 entries, all previously `state="new"`).

Covers GICS sector names (Energy, Utilities, Consumer Defensive/Cyclical, etc.), admin/market-data labels (Paid, Trial, Duration, Data Gathering Frequency), the feature list on the about page (Stock Tracking, Dividend Tracking, Tax Reporting, ...), and some marketing copy on the landing/pricing pages.

Also caught two entries (`currency effect` block-if strings) that were already translated but had drifted out of sync with the source's variable name (`assetProfile` vs a stale `SymbolProfile`), which is presumably why the extractor kept flagging them as `new` even though the Dutch text itself was fine — fixed those too so they match source exactly.

Checked existing translated entries for terminology already in use (e.g. Real Estate -> Vastgoed, Monthly -> Maandelijks) to keep new strings consistent with the rest of the file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Set the change detection strategy to `OnPush` in the _FIRE_ page
 - Improved the language localization for German (`de`)
+- Improved the language localization for Dutch (`nl`)
 
 ## 3.21.0 - 2026-07-05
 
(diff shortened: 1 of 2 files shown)


dtslvr (approved):

dtslvr:
Related to #3590.

PR #7261, opened 2026-07-06
Task/improve language localization for FR

munzzyy:
Fills in the remaining untranslated strings in `messages.fr.xlf` (186 entries, all previously `state="new"`).

Covers GICS sector names, admin/market-data labels, the FIRE calculator page, the about-page marketing copy, and the risk-rule engine's category names and threshold message templates (account/asset-class/currency/regional-market cluster risk, buying power, emergency fund). For the rule templates I matched the phrasing already used in the handful that were translated (e.g. "La majeure partie de votre investissement actuel... et n'excède pas...") so the new ones read consistently with the existing ones instead of introducing a second style.

Also fixed three entries that were already translated but had gone stale relative to the current source: two "currency effect" strings whose `<x>` interpolation still referenced `SymbolProfile` after the component was renamed to `assetProfile`, and a "Set up" emergency-fund label that had picked up an extra "Fonds d'urgence :" prefix somewhere along the way (checked against the German file, which has this one clean, to confirm the expected scope).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Set the change detection strategy to `OnPush` in the _FIRE_ page
-- Improved the language localization for German (`de`)
 - Improved the language localization for French (`fr`)
+- Improved the language localization for German (`de`)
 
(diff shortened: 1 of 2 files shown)


dtslvr (approved):

dtslvr:
Related to #3589.

PR #7283, opened 2026-07-09
fix(portfolio): correct currency-effect return for reopened positions

### LINKED ISSUE #6857: [BUG] `netPerformancePercentWithCurrencyEffect` produces misleading percentage for multi-trade positions with currency conversion
### Bug Description

**Setup:**
- Base currency: MYR
- Asset: MU (Micron Technology, USD-denominated stock)
- Activity: Multiple round-trip trades (buy→sell→buy→sell→buy)

**Observed Behavior:**
Ghostfolio reports these three metrics for the same position simultaneously:

| Metric | Value |
|--------|-------|
| `grossPerformancePercent` | 59.2% |
| `netPerformancePercent` | 53.9% |
| `netPerformancePercentWithCurrencyEffect` | 121.9% |

The `netPerformancePercentWithCurrencyEffect` (121.9%) is more than double the non-FX-adjusted version (53.9%). Over this period, MYR **strengthened** against USD (from ~4.40 to ~3.93), so the currency-adjusted return should be *lower* than the USD return, not higher. This indicates the calculation is producing economically nonsensical results.

### Root Cause

`netPerformancePercentWithCurrencyEffect` = `netPerformanceWithCurrencyEffect / investmentValueWithCurrencyEffect`

The problem is that `investmentValueWithCurrencyEffect` is the **sum of all buy-side transactions** converted to base currency at historical rates. It does **not** net out sell transactions. For a position with multiple round-trip trades, the denominator keeps growing with each buy even though the user's net cash outlay is much smaller. This inflates the denominator and produces a percentage that does not reflect actual economic return.

### Concrete Example

Suppose base currency is MYR, and a user trades a USD stock:

| Date | Action | USD Amount | USD/MYR Rate | MYR Amount |
|------|--------|-----------|-------------|-----------|
| Jan | BUY 10 shares @ $100 | -$1,000 | 4.00 | +RM 4,000 |
| Feb | SELL 10 shares @ $120 | +$1,200 | 4.10 | -RM 4,920 |
| Mar | BUY 3 shares @ $500 | -$1,500 | 3.90 | +RM 5,850 |

**User's true P&L:**
- Net cash outlay = $1,000 - $1,200 + $1,500 = **$1,300 USD** ≈ RM 5,070
- Current value = 3 × $500 = **$1,500 USD** ≈ RM 5,850
- **Real return ≈ +15.4%**

**Ghostfolio calculates:**
- `investmentValueWithCurrencyEffect` = RM 4,000 + RM 5,850 = **RM 9,850** (sells excluded)
- `netPerformanceWithCurrencyEffect` = RM 5,850 - RM 9,850 = **-RM 4,000**
- **Reported return = -40.6%** (shows a loss when the user actually gained)

### Expected Behavior

`netPerformancePercentWithCurrencyEffect` (and related performance percentage metrics) should be calculated against the user's **net cash invested** (cumulative buys minus cumulative sells), not the gross cumulative buy-side total. Alternatively:
1. The `investment` field in the API should represent net investment, or
2. A separate `netInvestment` field should be added alongside the existing `investment` field, with clear documentation on which basis each percentage is calculated

### Screenshots

N/A

### Logs

N/A

### Environment

- Ghostfolio Version: 3.2.0
- Cloud or Self-hosted: Self-hosted
- Experimental Features: NA
- Browser: Chrome
- OS: MacOS

### Additional Context

This was discovered via the API (`/api/v1/portfolio/holdings` endpoint). The `investment` field for the position was **$8,924.83** (gross buys), while the actual net cash invested was approximately **$2,122.51** after accounting for sells. The stock's price moved from ~$72 to ~$804 during the period. See related issues: #3330, #5793.

Ripwords:
## Summary

`netPerformancePercentWithCurrencyEffect` was divided by a day-weighted average of *accumulated* investment (`investmentValuesAccumulatedWithCurrencyEffect`). For a position that is fully closed and reopened, that average spans lots no longer held, so the denominator no longer reflects the net invested capital — producing a misleading percentage (e.g. `22.4%` instead of `15.7%` in the added test).

This divides the `max` currency-effect percentage by the current net investment (`totalInvestmentWithCurrencyEffect`) once a position has been closed and reopened. Numerator, `investment`, and net-performance values are unchanged.

Scope is deliberately narrow (only the `max` range, only the currency-effect metric) to keep the change small.

Closes #6857

## Testing

- Added `portfolio-calculator-multi-trade-currency-effect.spec.ts` (USD/MYR buy→sell→buy round trip; fails without the fix)
- `nx test api` — calculator suite passes, no regressions
- `nx affected:lint`, `nx format:check`


Ripwords:
Closing this. I applied the change to a scratch branch, ported the test to the current API (it no longer compiles against `main` — `symbolProfileDummyData` was renamed and the `Activity` shape changed) and measured it. It makes the metric worse, not better.

The change divides the currency-effect return by the current net investment instead of the average investment. Measured across three scenarios:

| Scenario | Return, no FX | `main` | With this PR |
|---|---|---|---|
| This PR's test: one round trip, MYR 4.00 → 3.90 | 19.69% | 22.37% | 15.73% |
| #6857 shape: five trades, MYR 4.40 → 3.93 | 50.20% | **43.12%** | **62.06%** |
| Same, larger swing: MYR 4.40 → 3.00 | 50.20% | **28.24%** | **52.96%** |

When MYR strengthens against USD, a MYR-reporting investor holding a USD asset must earn *less* than the USD return. `main` gets that right in both round-trip scenarios (43% and 28% against a USD return of 50%). This PR pushes the figure *above* the USD return, which is not attainable. In the scenario #6857 actually describes, the patch produces the exact pathology the issue complains about.

The cause is the denominator. Dividing by the final position size attributes the entire gain to capital that was not deployed when the gain was earned — in the second scenario `main`'s average of about RM 5,090 tracks the capital genuinely at work, while this PR uses RM 3,537, which is just the last buy.

I have also closed #6857: the behaviour there turned out not to be a defect. See https://github.com/ghostfolio/ghostfolio/issues/6857#issuecomment-5384658089 for the detail.

PR #7296, opened 2026-07-09
Task/improve language localization for NL

munzzyy:
Fills in the 14 entries in `messages.nl.xlf` that went to `state="new"` since #7260 merged (2026-07-07). These are new strings added by other PRs since then: `Copy`, `Close Account`, `Net Worth Tracking`, `Tax Reporting`, the security hint text on the close-account flow, and the `Post to Ghostfolio on X` share link.

3 of the 14 already had correct Dutch text but were still flagged `new`, the same drift #7260 ran into. The source's `<x>` interpolation picked up an `i18n-title` attribute that the target's copy never got, so the extractor kept treating an already-translated string as untranslated. Fixed those by matching the `equiv-text` back to source and flipping the state.

Kept to the terminology already in the file. `Dividend Tracking` translates to `Dividend volgen` and `Stock Tracking` to `Aandelen volgen`, so `Net Worth Tracking` follows the same pattern (`Netto waarde volgen`, with `Netto waarde` itself already used elsewhere). Same for `Close Account`, which the file already translates as `Account Sluiten` at a different call site for the identical button label.

### DIFF
--- a/apps/client/src/locales/messages.nl.xlf
+++ b/apps/client/src/locales/messages.nl.xlf
@@ -7218,7 +7218,7 @@
       </trans-unit>
       <trans-unit id="2137905359212133111" datatype="html">
         <source><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is Open Source Software</source>
-        <target state="translated"><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is Open Source Software</target>
+        <target state="translated"><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is open source software</target>
         <context-group purpose="location">
           <context context-type="sourcefile">apps/client/src/app/pages/resources/personal-finance-tools/product-page.html</context>
           <context context-type="linenumber">183</context>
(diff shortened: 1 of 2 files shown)


apps/client/src/locales/messages.nl.xlf
```diff
@@ -7218,7 +7218,7 @@
       </trans-unit>
       <trans-unit id="2137905359212133111" datatype="html">
         <source><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is Open Source Software</source>
-        <target state="new"><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is Open Source Software</target>
+        <target state="translated"><x id="INTERPOLATION" equiv-text="{{ product1().name }}"/> is Open Source Software</target>
```
dtslvr: ❓

dtslvr (changes requested):

apps/client/src/locales/messages.nl.xlf
munzzyy: Good catch, that one was just a state flip with no actual translation behind it. Fixed it to real Dutch now (`is open source software`, lowercase, matching how the rest of the file handles that term). Checked the rest of the diff too and this was the only entry that slipped through untranslated.

dtslvr (approved):
Thanks again, @munzzyy! 👍🏻

PR #7383, opened 2026-07-19
Setup VS Code Dev Container configuration

### LINKED ISSUE #7380: Setup VS Code Dev Container configuration
Introduce a _Dev Container_ configuration to simplify the local development process for _VS Code_ users. This issue is best suited for regular contributors who are already familiar with the _Ghostfolio_ codebase and setup.

### Implementation details

* Add the `.devcontainer` configuration files to the repository
* Update [DEVELOPMENT.md](https://github.com/ghostfolio/ghostfolio/blob/main/DEVELOPMENT.md) with a "How to" guide for developing inside the _Dev Container_

kojinkai:
## Summary
- Add `.devcontainer/devcontainer.json`, `.devcontainer/docker-compose.yml` and `.devcontainer/Dockerfile` so contributors can develop Ghostfolio fully inside VS Code's Dev Containers, with Node.js 22, PostgreSQL and Redis wired together automatically
- The `app` service extends `docker/docker-compose.dev.yml` (rather than duplicating the Postgres/Redis definitions) and waits on their healthchecks before starting
- Document the new workflow in `DEVELOPMENT.md`, including the two non-obvious gotchas discovered while testing this locally:
  - Inside the container, `.env` must be based on `.env.example` (service names `postgres`/`redis`), not `.env.dev` (`localhost`), since the app runs as its own container on the same Docker network
  - The Angular client dev server binds to `localhost` by default, so it must be started with `--host 0.0.0.0` for VS Code's forwarded ports to reach it from the host browser; the API server already binds to `0.0.0.0` by default and needs no change

Closes #7380

## Test plan
- [x] Built the `.devcontainer/Dockerfile` image and confirmed it builds successfully
- [x] Brought up the merged compose stack (`docker/docker-compose.dev.yml` + `.devcontainer/docker-compose.yml`) and confirmed the `app` service can reach `postgres:5432` and `redis:6379` by service name
- [x] Confirmed the existing Postgres data volume is unaffected by adding the `app` service
- [x] Opened the repo in VS Code via "Dev Containers: Reopen in Container", ran `npm run database:setup`, `npm run start:server`, and `npm run start:client -- --host 0.0.0.0`, and confirmed the app loads at `https://localhost:4200/en` in the host browser

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,9 @@
 FROM node:22-slim
 
-# Match the build tooling used by the production Dockerfile so native
+# Match the build tooling used by the production Dockerfile (keep this
+# list and the --no-install-suggests flag in sync with it) so native
 # dependencies (e.g. Prisma) compile correctly inside the Dev Container
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN apt-get update && apt-get install -y --no-install-suggests \
   g++ \
   git \
   make \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:22-slim
+
+# Match the build tooling used by the production Dockerfile so native
+# dependencies (e.g. Prisma) compile correctly inside the Dev Container
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  g++ \
+  git \
+  make \
+  openssl \
+  python3 \
+  && rm -rf /var/lib/apt/lists/*
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -1,0 +1,13 @@
+services:
+  app:
+    build:
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
+    volumes:
+      - ..:/workspaces/ghostfolio:cached
+    command: sleep infinity
+    depends_on:
+      postgres:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -5,9 +5,13 @@ services:
       dockerfile: .devcontainer/Dockerfile
     volumes:
       - ..:/workspaces/ghostfolio:cached
+      - node_modules:/workspaces/ghostfolio/node_modules
     command: sleep infinity
     depends_on:
       postgres:
         condition: service_healthy
       redis:
         condition: service_healthy
+
+volumes:
+  node_modules:
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -26,9 +26,9 @@ As an alternative to the manual _Setup_ above, [Visual Studio Code](https://code
 
 - [Docker](https://www.docker.com/products/docker-desktop)
 - [Visual Studio Code](https://code.visualstudio.com) with the [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension
-- Copy the file `.env.example` to `.env` and populate it with your data (`cp .env.example .env`)
+- Copy the file `.env.dev` to `.env` and populate it with your data (`cp .env.dev .env`)
 
-**Info:** Use `.env.example`, not `.env.dev`, as the base for `.env`. Inside the Dev Container, the application and the databases run as separate containers on the same Docker network, so `DATABASE_URL` and `REDIS_HOST` must point to the service names `postgres` and `redis` (as `.env.example` already does), not `localhost`.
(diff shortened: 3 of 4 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  app:
+    build:
+      context: ..
+      dockerfile: .devcontainer/Dockerfile
+    volumes:
+      - ..:/workspaces/ghostfolio:cached
```
dtslvr, line 7: Should we add a separate volume for `node_modules` here? Right now the whole repo is bind-mounted, so if someone already ran `npm install` on their host (Mac vs. Windows) the container reuses those native binaries and things like _Prisma_ break with "wrong platform" errors.

DEVELOPMENT.md
```diff
@@ -18,6 +18,30 @@
...
+
+As an alternative to the manual _Setup_ above, [Visual Studio Code](https://code.visualstudio.com) users can develop inside a [Dev Container](https://containers.dev). It runs the application in a Docker container alongside [PostgreSQL](https://www.postgresql.org) and [Redis](https://redis.io), pre-installed with the required build tooling (Node.js 22, `g++`, `make`, `python3`).
+
+#### Prerequisites
+
+- [Docker](https://www.docker.com/products/docker-desktop)
+- [Visual Studio Code](https://code.visualstudio.com) with the [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension
+- Copy the file `.env.example` to `.env` and populate it with your data (`cp .env.example .env`)
+
+**Info:** Use `.env.example`, not `.env.dev`, as the base for `.env`. Inside the Dev Container, the application and the databases run as separate containers on the same Docker network, so `DATABASE_URL` and `REDIS_HOST` must point to the service names `postgres` and `redis` (as `.env.example` already does), not `localhost`.
```
dtslvr: `.env.dev` is the only env file that sets `NX_ADD_PLUGINS=false`, and the rest of the dev setup relies on it. Pointing people to `.env.example` here quietly drops that. Could we instead keep `.env.dev` as the base and override `DATABASE_URL` / `REDIS_HOST` for the app service, so the _Nx_ config stays the same as everyone else's?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:22-slim
+
+# Match the build tooling used by the production Dockerfile so native
+# dependencies (e.g. Prisma) compile correctly inside the Dev Container
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
dtslvr: This package list is a copy of the one in the root `Dockerfile`. Worth a note that they need to stay in sync. By the way, this uses `--no-install-recommends` while the production `Dockerfile` uses `--no-install-suggests`. Might be nice to match them so the "same as production" comment holds.

DEVELOPMENT.md
```diff
@@ -18,6 +18,30 @@
...
+
+#### Setup
+
+1. Open the repository folder in Visual Studio Code
+1. Run the command _Dev Containers: Reopen in Container_ (this builds the container and runs `npm install` automatically)
+1. In the container's integrated terminal, run `npm run database:setup` to initialize the database schema
+1. Start the [server](#start-server)
+1. Start the client with `npm run start:client -- --host 0.0.0.0` instead of the usual command (see note below)
```
dtslvr, line 39: `start:client` includes `-o`, which tries to open a browser. Inside the container there isn't one.

DEVELOPMENT.md
```diff
@@ -18,6 +18,30 @@
 1. Open https://localhost:4200/en in your browser
 1. Create a new user via _Get Started_ (this first user will get the role `ADMIN`)
 
+### Dev Container
+
+As an alternative to the manual _Setup_ above, [Visual Studio Code](https://code.visualstudio.com) users can develop inside a [Dev Container](https://containers.dev). It runs the application in a Docker container alongside [PostgreSQL](https://www.postgresql.org) and [Redis](https://redis.io), pre-installed with the required build tooling (Node.js 22, `g++`, `make`, `python3`).
```
dtslvr: The list here says `g++`, `make`, `python3` but the `Dockerfile` also installs `openssl` and `git`.

dtslvr (changes requested):
Thanks for your contribution, @kojinkai 😎

DEVELOPMENT.md
kojinkai: thanks for the review and the patience @dtslvr 

WRT the `-o` flag - it appears to be a harmless no-op when running the `npm run start:client` command in the dev container so maybe there are 2 options here:
1. Update `DEVELOPMENT.md` to explain that the `-o` command is benign but has been considered.
2. Add a new `npm run start:client:container` command e.g. `"start:client:container": "nx run client:copy-assets && nx run client:serve --configuration=development-en --hmr --host 0.0.0.0",`

I've gone with option 1 here as it has the smallest footprint but let me know if you have any strong feelings about another option here.

kojinkai:
@dtslvr All five review comments have been addressed and the threads resolved:

- Added a named `node_modules` volume so it doesn't get shadowed by a host bind-mount
- Switched the `.env` base to `.env.dev` (preserving `NX_ADD_PLUGINS=false`) instead of `.env.example`
- Aligned `--no-install-suggests` with the production `Dockerfile`
- Noted that `start:client`'s `-o` flag is a harmless no-op in the container
- Synced the prerequisites list in `DEVELOPMENT.md` with the actual packages installed

Branch is up to date with `main` and mergeable. Could you take another look when you have a chance? Thanks!

PR #7557, opened 2026-08-06
fix: display cryptocurrency quantities with asset precision in holdings table

### LINKED ISSUE #7545: [BUG] Account Holdings table rounds cryptocurrency quantities to 2 decimals
## Bug description

Cryptocurrency quantities are still formatted with two decimal places in **Accounts → account details → Holdings**.

For Bitcoin this hides materially important precision. A holding of `0.01525217 BTC` is displayed as `0.02`, so the displayed quantity can differ materially from the actual position.

The underlying activity quantities remain precise; this appears to be a presentation-only issue in the account holdings table.

## Steps to reproduce

1. Add Bitcoin activities whose resulting quantity is `0.01525217 BTC`.
2. Go to **Accounts**.
3. Open the account containing the BTC activities.
4. Select the **Holdings** tab.
5. Observe the **Quantity** column.

## Actual behavior

The Quantity column displays:

```
0.02
```

## Expected behavior

Cryptocurrency quantities should use asset-appropriate precision, ideally up to 8 decimal places for BTC:

```
0.01525217
```

Trailing zeroes could be omitted, but significant fractional digits must not be discarded.

## Environment

- Ghostfolio version: 3.42.0
- Self-hosted
- Browser: Safari
- Route: `/en/accounts?accountId=...&accountDetailDialog=true`
- View: account detail dialog → Holdings

## Related work

- #4253 was fixed for a different view.
- #4085 described the same user impact but was closed as not planned.
- #6038 increased cryptocurrency precision in the holding detail dialog, but the account Holdings table is still limited to two decimals.
- #5160 was closed without merge.

This may be a missed consumer of the cryptocurrency-aware quantity precision logic rather than a calculation/storage problem.

slegarraga:
Fixes #7545.

## Summary

The account detail dialog's **Holdings** tab reuses the shared `gf-holdings-table`. Its Quantity column formats every value with `gf-value` and `isCurrency="true"`, which defaults to 2 decimal places. As a result, a Bitcoin holding of `0.01525217 BTC` is displayed as `0.02`, hiding materially important precision.

This PR makes quantity precision asset-aware in the shared holdings table:

- Integer quantities keep `0` decimal places.
- Fractional cryptocurrency quantities use `8` decimal places, preserving satoshi-level precision regardless of position size.
- Other asset classes keep the existing `2` decimal places.

## Changes

- Add a small pure helper for holdings-table quantity precision.
- Pass the computed precision to `gf-value` in the Quantity column.
- Add focused tests covering small and larger fractional BTC quantities, integer crypto quantities, and stocks.

## Testing

- UI lint passes with no errors.
- Relevant UI tests pass: 3 suites, 9 tests.
- Client production build passes.

### DIFF
--- a/libs/ui/src/lib/holdings-table/holdings-table.component.ts
+++ b/libs/ui/src/lib/holdings-table/holdings-table.component.ts
@@ -1,3 +1,4 @@
+import { NUMERICAL_PRECISION_THRESHOLD_6_FIGURES } from '@ghostfolio/common/config';
 import {
   canOpenHoldingDetail,
   getLocale,
@@ -7,6 +8,7 @@ import {
   AssetProfileIdentifier,
   PortfolioPosition
 } from '@ghostfolio/common/interfaces';
+import { AssetSubClass } from '@prisma/client';
 
 import {
   CUSTOM_ELEMENTS_SCHEMA,
@@ -101,6 +103,28 @@ export class GfHoldingsTableComponent {
 
   protected canShowDetails(holding: PortfolioPosition): boolean {
     return this.hasPermissionToOpenDetails() && canOpenHoldingDetail(holding);
+  }
+
+  protected getQuantityPrecision(holding: PortfolioPosition): number {
+    if (Number.isInteger(holding.quantity)) {
+      return 0;
+    }
+
+    if (holding.assetProfile.assetSubClass === AssetSubClass.CRYPTOCURRENCY) {
+      if (holding.quantity < 10) {
+        return 8;
+      }
+
+      if (holding.quantity < 1000) {
+        return 6;
+      }
+
+      if (holding.quantity >= NUMERICAL_PRECISION_THRESHOLD_6_FIGURES) {
+        return 0;
+      }
+    }
+
+    return 2;
   }
 
   protected onOpenHoldingDialog({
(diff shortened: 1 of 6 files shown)


libs/ui/src/lib/holdings-table/holdings-table.component.ts
```diff
@@ -103,6 +105,28 @@ export class GfHoldingsTableComponent {
     return this.hasPermissionToOpenDetails() && canOpenHoldingDetail(holding);
   }
 
+  protected getQuantityPrecision(holding: PortfolioPosition): number {
+    if (Number.isInteger(holding.quantity)) {
+      return 0;
+    }
+
+    if (holding.assetProfile.assetSubClass === AssetSubClass.CRYPTOCURRENCY) {
```
DireMunchkin: I would  personally prefer if you just `return 8` here without looking at `holdings.quantity`

DireMunchkin: I should elaborate why - It's because Bitcoin tracks amounts as a integral unit called a Satoshi or sat internally, which is exactly 1/100 000 000 of a full Bitcoin. So 8 decimal places becomes the magic number at which you have full precision.

slegarraga: You're absolutely right — thank you for catching that and for the satoshi example. The size-based logic could still truncate valid BTC precision once the quantity reached 10.

I've updated the branch so every fractional cryptocurrency quantity uses 8 decimal places (integers still use 0, and non-crypto assets remain at 2). I also added focused coverage for both `0.01525217 BTC` and `10.12345678 BTC`, rebased the change onto current `main`, and removed an unrelated commit that had accidentally landed on the branch.

The relevant tests, UI lint, and client production build all pass on the updated commit (`0c0b3e602`). Really appreciate the careful review!

PR #7616, opened 2026-08-13
Bugfix/cash balance update related to activities in custom currency

vicbentu:
The cash movement for activities with "Update Cash Balance" enabled used the asset profile's currency instead of the activity's currency.

The amount is derived from the unit price and is therefore denominated in the activity's currency. Using the asset profile's currency caused activities with a currency override to be converted from the wrong unit, silently writing an incorrect account balance.

A EUR 4,000 purchase of 7974.T, quoted in JPY, on a EUR account with a EUR 1 fee debited EUR 21.78 instead of EUR 4,001.00. The EUR 4,001 amount was incorrectly interpreted as JPY and converted using the JPY/EUR exchange rate.

Portfolio valuation is unaffected because it already uses the activity's currency correctly. Only the cash balance drifts.

The fix uses the activity's currency for the balance update and falls back to the asset profile's currency when no activity currency is set. This preserves the existing behavior for activities without a currency override.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dtslvr (approved):
Thanks and congrats on your first contribution, @vicbentu! 😎

PR #7631, opened 2026-08-15
Bugfix/missing benchmark in performance chart for calendar year date ranges

### LINKED ISSUE #7627: Missing benchmark in performance chart for calendar year date ranges
## Bug Description

The benchmark is missing in the performance chart when a calendar year date range (e.g. `2025`) is selected. The chart shows the portfolio line only.

The API logs an error for every affected benchmark:

```
ERROR [BenchmarksService] No historical market data has been found for BTCUSD (YAHOO) at 2024-12-31
```

Historical market data _is_ present for that date in the `MarketData` table, so the message is misleading.

## Expected Behavior

The benchmark line is shown next to the portfolio line, as it is for the other date ranges (`1d`, `wtd`, `mtd`, `ytd`, `1y`, `5y`, `max`).

## Steps to Reproduce

1. Include an asset profile in _Benchmark / Markets_.
1. Select the benchmark in _Compare with..._ on the _Analysis_ page.
1. Select a calendar year date range, e.g. `2025`.
1. The benchmark line is missing. The API log contains the error above.

## Environment

- Ghostfolio version: `3.50.0`

AaryamanVyas:
#7627
## Problem

For calendar year date ranges (e.g. `2025`), the benchmark line is missing from the performance chart in the Analysis page.

## Root Cause

Commit `031b3b68c` changed the calendar year `startDate` from `new Date(dateRange)` (midnight, inclusive) to `subMilliseconds(yearStartDate, 1)` (23:59:59.999, exclusive). The `BenchmarksService` was never updated to account for this exclusive boundary.

For calendar year '2025':
- `startDate = 2024-12-31T23:59:59.999Z`
- `endDate = 2025-12-31T23:59:59.999Z`

The portfolio chart filters activities with `!isBefore(date, start)`, so `2024-12-31` (midnight) is excluded from the chart. The market data query is built from the chart dates, so `2024-12-31` is never fetched. Then `isSameDay(date, startDate)` fails because no market data item exists for that date.

Named ranges (ytd, 1d, 1y, etc.) work because their startDate is computed at midnight where `isBefore(X, X)` = false.

## Fix

Add `resetHours(startDate)` to the `date.in` array in the market data query in `BenchmarksService` so the start date's market data is always fetched, regardless of whether it appears in the portfolio chart. For named ranges where the start date is already in the chart, this is a harmless duplicate.

## Testing

- All existing tests pass (81 common, 2 benchmark service)
- Added 5 new unit tests for calendar year and named date ranges
- Lint passes (0 errors, 0 warnings)
- Build succeeds


apps/api/src/app/endpoints/benchmarks/benchmarks.service.ts
dtslvr, line 1: With a calendar year selected, the two lines do not start at 0.

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## 3.51.0 - 2026-08-14
```
dtslvr: Please add a proper changelog entry.

dtslvr (changes requested):
Thanks for looking into this, @AaryamanVyas!

apps/api/src/app/endpoints/benchmarks/benchmarks.service.ts
AaryamanVyas: Thanks for the feedback! I updated the implementation to use the first date in the chart as the baseline date, so the benchmark and portfolio use the same starting point for calendar year ranges.

CHANGELOG.md
AaryamanVyas: Done — I added the changelog entry under Unreleased > Fixed.

dtslvr (approved):
Thanks again and congrats on your first contribution, @AaryamanVyas! 🙌🏻

PR #7647, opened 2026-08-17
Task/upgrade typescript-eslint to version 8.67.0

KenTandrian:
Hi @dtslvr, this PR upgrades `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` from `8.43.0` to `8.67.0`.

> This PR is part of a series of targeted dependency updates acting as a prerequisite for the upcoming TypeScript 6.0 and Angular 22 upgrade, ensuring parser compatibility and rule correctness ahead of the TypeScript bump.

## Changes Made
- Updated `@typescript-eslint/eslint-plugin` to `8.67.0`
- Updated `@typescript-eslint/parser` to `8.67.0`
- Fixed `@typescript-eslint/prefer-optional-chain` lint errors surfaced by the parser update in:
  - `apps/api/src/app/activities/activities.controller.ts`
  - `apps/api/src/app/import/import.service.ts`


apps/api/src/app/activities/activities.controller.ts
```diff
@@ -301,7 +301,7 @@ export class ActivitiesController {
       id
     });
 
-    if (!originalActivity || originalActivity.userId !== this.request.user.id) {
+    if (originalActivity?.userId !== this.request.user.id) {
```
dtslvr: Doesn't `?.` change the logic here? If `originalActivity` is `null` / `undefined`, `originalActivity?.userId !== ...` evaluates to `false` instead of `true`, so it might bypass the check if the activity is missing.

apps/api/src/app/import/import.service.ts
```diff
@@ -573,7 +573,7 @@ export class ImportService {
         // If there is no asset profile or if the asset profile belongs to a
         // different user, then reuse the custom asset profile of the user or
         // create a new asset profile
-        if (!existingAssetProfile || existingAssetProfile.userId !== user.id) {
+        if (existingAssetProfile?.userId !== user.id) {
```
dtslvr, line 576: See above

apps/api/src/app/activities/activities.controller.ts
KenTandrian: Ah, good catch! When `originalActivity` is missing, `originalActivity?.userId` evaluates to `undefined`, so `undefined !== user.id` still evaluates to `true` (entering the block).

However, using `?.` can be risky if `user.id` ever happens to be `undefined` as well (`undefined !== undefined` evaluates to `false`). I've updated it to explicit guards to be safer and keep the intent clear.

apps/api/src/app/import/import.service.ts
KenTandrian: I believe `existingAssetProfile?.userId !== user.id` actually does the exact same thing. 

When `existingAssetProfile` is missing, `existingAssetProfile?.userId` is `undefined`, and `undefined !== user.id` evaluates to `true`. So both expressions enter the `if` block as intended.

dtslvr (approved):
Thanks again for your contribution, @KenTandrian! 👍🏻

PR #7660, opened 2026-08-18
Bugfix/handling of cash positions

### LINKED ISSUE #7556: [BUG] Cash must be allocated to a known sector, to have a proper distribution graph
**Bug Description**
Cash is appearing as an unknow sector in the distribution graph.

**To Reproduce**
1. Porfolio
2. Distribution
3. See the graph cash appears with no sector.

**Expected Behavior**
A new sector could be created and recognized by Ghostfolio such as:
[{"name":"Cash","weight":1}]

**Screenshots**

Image

**Logs**
No logs are needed.

**Environment**
- Ghostfolio Version 3.43
- Self-hosted
- Brave
- MacOs

**Additional Context**
All the others assets must be 100% percent complete, with sector and weights.

dtslvr:


apps/api/src/app/endpoints/public/public.controller.ts
```diff
@@ -178,23 +174,19 @@ export class PublicController {
         assetProfile: {
           ...portfolioPosition.assetProfile,
           assetClass:
-            hasDetails ||
-            portfolioPosition.assetProfile.assetClass === AssetClass.LIQUIDITY
+            hasDetails || isCashPosition(portfolioPosition.assetProfile)
               ? portfolioPosition.assetProfile.assetClass
               : undefined,
```
Lyndros: Hi @dtslvr, Are you sure that we are not mixing two different things ?

Should we create an additional helper for liquidity just in case...

apps/api/src/app/portfolio/portfolio.controller.ts
```diff
@@ -131,40 +133,7 @@ export class PortfolioController {
...
-      }
-
-      for (const [name, { valueInBaseCurrency }] of Object.entries(accounts)) {
-        accounts[name].valueInPercentage = valueInBaseCurrency / totalValue;
-      }
-
-      for (const [name, { valueInBaseCurrency }] of Object.entries(platforms)) {
-        platforms[name].valueInPercentage = valueInBaseCurrency / totalValue;
-      }
+      convertValuesToPercentages({ accounts, holdings, platforms });
```
Lyndros, line 133: Great refactor happy to see a function here :-)

Lyndros (commented):
LGTM, happy to see more functions in the refactors. 

l only have a question regarding the meaning of liquidity and cash for the assets profiles.

Lyndros (commented):
LGTM only a question regarding the meaning of liquidity and cash for the assets profiles.

Lyndros:
@dtslvr why don't we merge it ?

dtslvr:
Hi @Lyndros, thanks for the review!

`LIQUIDITY` and `CASH` are indeed two different things and that is why the check has changed: the asset class `LIQUIDITY` also contains cryptocurrencies, so it was never a reliable criterion for “this position is cash”. In that block `assetClass` was kept for `LIQUIDITY` while `assetSubClass` right below was kept for `CASH` only. Now both check the same thing. No `isLiquidityPosition()` helper for now, since no other place derives “is cash” from the asset class. It stays a single comparison if we ever need it.

I have rebased the branch and fixed one more issue: in the restricted view and on the public page the excluded cash came back as a “No data available” slice, because the chart allocates the difference to 100% to the unknown bucket.

Lyndros (approved):
Thanks @dtslvr now is clear.

LGTM! Great :-)

PR #7665, opened 2026-08-19
Task/migrate away from deprecated node10 module resolution

KenTandrian:
Hi @dtslvr, this PR migrates API app away from the deprecated `node10` module resolution. It now inherits the module resolution from the base TypeScript configuration, which is `bundler`.

## Changes

* Removed `moduleResolution=node10` configuration from both API app and spec.
* Changed `module` to `preserve`.
* Updated `yahoo-finance2` imports.
* Remove ignore deprecations flag.


apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
dtslvr, line 1: Could you use `import type { ...` as well here?

dtslvr (changes requested):
Thanks for your contribution, @KenTandrian! 🙂

apps/api/src/services/data-provider/yahoo-finance/yahoo-finance.service.ts
KenTandrian: Done!

dtslvr (approved):

PR #7671, opened 2026-08-20
Task/migrate services to Angular 22 @Service()

### LINKED ISSUE #7669: Migrate services to Angular 22 @Service()
Use Angular 22's `@Service()` as a leaner alternative to `@Injectable({ providedIn: 'root' })`.

```ts
// Before
@Injectable({ providedIn: 'root' })

// After
@Service()
```

RaghavShubham:
Resolves #7669

## What

Replaces `@Injectable({ providedIn: 'root' })` with `@Service()` across the 13 services that used it — 10 in `apps/client`, 3 in `libs/ui`.

`@Service()` marks a class as automatically available in the DI system, so it is equivalent to `providedIn: 'root'` (the opt-out is `@Service({ autoProvided: false })`).

## Two things worth a reviewer's attention

**1. `@Service()` rejects constructor injection.** `LayoutService` and `CacheService` both used it, and the build fails with:

```
NG2028: @Service class cannot use constructor dependency injection.
        Use the `inject` function instead.
```

Both now inject as fields, matching the pattern already used elsewhere in the codebase (`private readonly x = inject(X)`). `LayoutService` keeps its constructor body — it wires up `shouldReloadContent$` and sizes the notification dialogs — and only the parameters moved out. Field initialisers run before the constructor body, so the ordering is unchanged.

**2. Bare `@Injectable()` services are deliberately untouched.** `ModulePreloadService`, `HttpResponseInterceptor`, `LanguageService` and friends are not auto-provided; they're supplied through a providers list. Converting them would mean `@Service({ autoProvided: false })`, which is a different change from the one the issue describes. Happy to do it in a follow-up if you'd like.

One comment was preserved rather than dropped — `entity-logo-image-source.service.ts` carried `// Required to allow mocking in Storybook` inside the options object; it now sits above the decorator, since the rationale still holds.

## How this was verified

| Check | Result |
| --- | --- |
| `nx run client:build` | passes — covers the 10 `apps/client` services plus `data.service.ts` (110 refs) and `admin.service.ts` (20 refs) from `libs/ui` |
| `nx run ui:build-storybook` | passes — this is what actually compiles `entity-logo-image-source.service.ts`, which has no `apps/client` references |
| `nx run client:lint`, `nx run ui:lint` | pass, warnings unchanged from `main` |
| `prettier --check` | clean |
| pre-commit `affected:lint` + `format:check` | pass |

I checked that `Service` is genuinely exported by the installed `@angular/core@22.1.2` (`declare const Service: ServiceDecorator` in `types/core.d.ts`) rather than taking it from the issue text.

**Not verified:** I have not run the app and exercised these services at runtime. The DI change is compile-time checked and `inject()` is already used throughout, but a smoke test of the layout/notification behaviour would be worth a reviewer's eye, since `LayoutService` is the one file where I moved code rather than just swapping a decorator.

> [!NOTE]
> AI model used — `claude-opus-5`, reasoning effort `high`. Every line was reviewed before submitting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Refactored the services to use the `@Service()` decorator of Angular
+- Refactored the services to use the `@Service()` decorator of _Angular_
 
 ## 3.55.0 - 2026-08-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Refactored the services to use the `@Service()` decorator of Angular
+
 ## 3.55.0 - 2026-08-19
 
 ### Added
--- a/apps/client/src/app/adapter/custom-date-adapter.ts
+++ b/apps/client/src/app/adapter/custom-date-adapter.ts
@@ -1,9 +1,10 @@
 import { getDateFormatString } from '@ghostfolio/common/helper';
 
-import { inject } from '@angular/core';
+import { inject, Service } from '@angular/core';
 import { MAT_DATE_LOCALE, NativeDateAdapter } from '@angular/material/core';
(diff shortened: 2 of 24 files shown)


dtslvr:
Hi @RaghavShubham,

thanks for the contribution!

Could you please comment on #7669 so I can assign the issue to you? Also, please add a changelog entry. Thanks!

dtslvr (approved):
Thanks again and congrats on your first contribution, @RaghavShubham! 🙌🏻

PR #7745, opened 2026-08-28
Task/improve language localization for JA (20260828)

nyxst4ck:
## Description

Improves Japanese localization for the holding detail and home holdings views.

- translates Close Holding, Benchmarks, Chart, and Table

## Validation

- Parsed messages.ja.xlf as XML
- Ran git diff --check


dtslvr (changes requested):
Please add an entry to the changelog.

PR #7774, opened 2026-08-31
Task/upgrade zod to version 4.5.4

slagiewka:
And use zod/compile in MCP. Tested MCP before & after and works fine. Going over the UI itself works well still.

The only holdout in `package-lock.json` is `@angular/cli` that's pinned strictly to the old version.

Just upgrading to 4.5 has memory benefits. Using `compile` gives extra performance benefits. It's possible to hook the compile to the entire process (which would benefit imported code), but unlikely to do much.

https://zod.dev/blog/zod-4-5

### DIFF
--- a/apps/api/src/app/endpoints/mcp/mcp.controller.ts
+++ b/apps/api/src/app/endpoints/mcp/mcp.controller.ts
@@ -10,6 +10,7 @@ import { UseFilters } from '@nestjs/common';
 import { Payload } from '@nestjs/microservices';
 import { McpController, Tool } from '@rekog/mcp-nest';
 import { z } from 'zod';
+import 'zod/compile';
 
 import {
   GET_ACCOUNTS_PARAMETERS,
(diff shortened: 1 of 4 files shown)


dtslvr:
Thanks for creating this PR, @slagiewka!

I read about the Zod 4.5 release and it looks promising. We just need to wait for our `min-release-age=7` npm policy window to pass before we can merge this.

apps/api/src/app/endpoints/mcp/mcp.controller.ts
```diff
@@ -10,6 +10,7 @@ import { UseFilters } from '@nestjs/common';
 import { Payload } from '@nestjs/microservices';
 import { McpController, Tool } from '@rekog/mcp-nest';
 import { z } from 'zod';
+import 'zod/compile';
```
dtslvr, line 13: Why is this needed?

slagiewka: Because compile mode is opt-in. There are three ways you can use the mode:
1. import at an entry point (as shown)
2. use node's `--import` preload that ensures everything is compiled
3. explicitly build with `.compile()` and use the compiled output. In case of this MCP controller, it would be `z.infer<typeof GET_ACCOUNTS_PARAMETERS>.compile()` repeated in each zod instance.

I assume that benefits for this controller's performance are negligible since the instances are quite small.

PR #7838, opened 2026-09-06
Fixed the tag filter of the holdings so that it no longer affects the…

### LINKED ISSUE #7802: [BUG] Active and close positions filtering wrongly by tag
**Bug Description**

In the positions window filtering is not working as expected. 

**To Reproduce**

I have 2 positions, 1 active and 1 close with a gold asset.

Both positions are tagged as DEFENSIVE, the appears both correctly in active and close.

Image

Image

1. Remove the tag of one of the last sell activities.
2. Open positions filter by active.
3. Now appears both active independently that Invesco one is fully close.

Image

**Expected Behavior**

Independently of the tag an position status should not change, tags has nothing to do with the active or close status.

**Screenshots**

**Logs**

N/A

**Environment**

- Ghostfolio Version 3.65.0
- Self-hosted
- Brave
- Mac Os

Lyndros:
Fix filtering by tag

Now it keeps only the activities that belong to a holding matching at least one tag filter.

A holding matches if any of its activities or its account has the tag; once it matches, every one of its activities is kept (not just the tagged ones). 

This way, a holding's quantity and status (active or closed) always reflect its complete activity history instead of being fragmented by the tag of an individual activity.


Lyndros:
@dtslvr, Tested locally works as intended. 

Take a look to the new approach. I think is they way to go. 

Take into account that with the current implementation perfo/values only in the positions columns were completely wrong when filtering by tag, the situation is that currently is fragmenting the activities for the computation even if positions are fully closed... which makes no sense.

PR #7848, opened 2026-09-08
fix: allow updating cash balance for historical activities

### LINKED ISSUE #7834: [BUG] updating total cash fails if activity older then today
**Bug Description**
Feature "update cash balance" does its job if interest type activity set to today. If it is older, the activity is created but cash balance does not get updated.

**To Reproduce**

1. Create new activity 
2. Check update cash balance
3. Set date to yesterday or older

**Expected Behavior**

Cash is updated irrespective of what date is set would be preferred. If not doable for a reason, feature should be disabled when date not set to today (hidden, shaded...).

**Environment**

- Ghostfolio Version 3.64.0
- Self-hosted
- Edge last version
- Win last verison

**Additional Context**

Related find: https://github.com/ghostfolio/ghostfolio/issues/3162

jellyfishing2346:
## Description

The “Update Cash Balance” option was disabled whenever an activity date was
older than today, even though the backend supports updating account balances
using historical dates.

This change allows the option for historical activities while preserving the
existing restrictions for liabilities, valuables, and manual buy activities.

## Testing

- `npx nx build client --configuration=development-en`

## Related issue

Closes #7834

### DIFF
--- a/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.component.ts
+++ b/apps/client/src/app/pages/portfolio/activities/create-or-update-activity-dialog/create-or-update-activity-dialog.component.ts
@@ -560,12 +560,10 @@ export class GfCreateOrUpdateActivityDialogComponent {
     const accountBalanceControl = this.activityForm.get('updateAccountBalance');
     const accountId = this.activityForm.get('accountId')?.value;
     const dataSource = this.activityForm.get('dataSource')?.value;
-    const date = this.activityForm.get('date')?.value;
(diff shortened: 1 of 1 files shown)


jellyfishing2346:
@dtslvr Could you please review this PR when you have time? It fixes issue https://github.com/ghostfolio/ghostfolio/issues/7834 by allowing cash-balance updates for historical activities. The client build passes.

dtslvr:
Thanks for the contribution, @jellyfishing2346!

Updating balances requires backend support to recalculate subsequent dates. A frontend-only change leaves newer entries out of sync and breaks the history.

I'll close this PR, but I appreciate you taking the time to look into this.

PR #7850, opened 2026-09-08
Task/move lodash to lodash-es

slagiewka:
`lodash-es` is a ESM-only version of `lodash` (released from the same source). Potential benefits include 
- better tree-shaking of imports, especially for client side. However, I did a before and after build that doesn't seem to be much smaller.
- better cooperation with bundler+preserve `tsconfig` setup in API. The setup means that technically there's `import` on the CJS version that has no proper support for exports. This makes is easier to run API directly over source code rather than from webpack-ed `dist/`. 

Ghostfolio is already on node v22 which supports `require(esm)` - the case for server code that is still CJS.

The dep is already vendored in as a transitive dependency (in dev). Types are pretty much re-export of base types.


slagiewka:
Crikey, Jest. The usual ESM-troubles suspect.

Converting to draft. Unless you don't want to pursue this path - by all means, go ahead and close it.

PR #7871, opened 2026-09-13
fix(portfolio): ignore dividends when calculating TWI between trades

### LINKED ISSUE #5451: [BUG] Incorrect Performance % when dividends are received before later buys
Image

**Bug Description**

Performance % is calculated incorrectly when dividends are received before later buys. The absolute values (Investment, Current Value, Change) are correct, but the Performance % is distorted (e.g. –60.47% instead of –4.65%).

**To Reproduce**

- Buy 1 share of stock on 20/12/2024
Price: $10.80
Fee: $2.00
Total: $12.80

- Receive dividends on 17/04/2025
$0.17 + $0.10 = $0.27

- Buy 14 additional shares on 29/08/2025
Price: $12.00
Fee: $2.00
Total: $170.00

**Expected Behaviour**

Investment = $182.80

Current Value = $174.30

Change = –$8.50

Performance = –4.65%

**Screenshots**

**Logs**

**Environment**

- Ghostfolio Version 2.195.0
- Self-hosted
- Chrome
- Mac OS

**Additional Context**

CodeSageAbhijit:
## Description

Fixes the incorrect Time-Weighted Investment (TWI) calculation when a
dividend activity occurs between two BUY/SELL activities.

### Problem

The TWI calculation used the immediately preceding activity as the reference
date for the next BUY/SELL activity.

When a dividend was received between two trades, the dividend date was
therefore used as the previous investment date. This incorrectly changed the
TWI calculation and could result in an incorrect performance percentage.

For example:

BUY → DIVIDEND → BUY

The dividend activity should not reset the TWI period between the two trades.

### Changes

- Track the previous BUY/SELL activity date separately for the TWI calculation.
- Ignore non-BUY/SELL activities, such as dividends, when determining the
  previous investment date.
- Add a regression test covering a BUY → DIVIDEND → BUY sequence.

### Tests

- BALN BUY/BUY regression test
- BALN BUY/SELL test
- MSFT BUY with dividend test
- Full ROAI calculator test suite: 217 passed, 1 skipped
- `nx format:check`
- `git diff --check`

Fixes #5451


CodeSageAbhijit:
@dtslvr Could you please take another look at this PR when you get a chance?

I’ve resolved the merge conflict with the latest main and adapted the original fix to the current averageInvestment implementation.

During the conflict resolution, I kept the previous BUY/SELL activity as the reference date so that intermediate activities such as DIVIDEND don't reset the investment period.

I also updated the regression test to use the current averageInvestment fields. The BUY → DIVIDEND → BUY regression test passes, along with the existing BUY → BUY test.

In particular, it would be great if you could review the conflict resolution and confirm that the adapted implementation still matches the intended fix for #5451. Thanks!

